PR #194, opened 2017-08-10
Pic

reginafcompton:


reginafcompton:
@fgregg - I did not merge this branch into master, since I wanted to see how any edits to Councilmatic core might affect this. 

We still need to make these edits here (not Councilmatic core). Why? The edits for LA Metro related to a custom filter for agenda_urls, not a model property, as they do with Councilmatic core (for the Bill model). 

Am I understanding this correctly? And also, if this works well, then can we deploy your changes in the property-image-cache to production?

Issue #177, opened 2017-07-10
Board Reports: Account for new fiscal year in solr facets


reginafcompton:
It's July 2017 - time to update the facets!

<img width="402" alt="screen shot 2017-07-10 at 2 42 59 pm" src="https://user-images.githubusercontent.com/13078679/28036525-2ad4fce6-657e-11e7-954f-5a6ad1784505.png">

reginafcompton:
Added!...but #181 persists

Issue #128, opened 2017-04-21
Board reports - some do not have a title


reginafcompton:
This issue has two explanations:
* the board report does not have a "Subject" header in the ocr_full_text
* the board report does not have an ocr_full_text - likely, a scraper issue

N.B. Look at https://github.com/datamade/la-metro-councilmatic/blob/master/lametro/search_indexes.py#L29

reginafcompton:
Steps to getting ocr_full_text

The scraper grabs the “MatterTextPlain” from the API, and adds it to the “extras” field as “plain_text” in a bill
https://github.com/opencivicdata/python-legistar-scraper/blob/master/legistar/bills.py#L279
https://github.com/datamade/scrapers-us-municipal/blob/master/lametro/bills.py#L170

The import_data looks for “plain_text” and adds an “ocr_full_text” attribute to the Bill model

The problem?

 Sometimes, the text for a bill has multiple versions (http://webapi.legistar.com/v1/metro/matters/3534/versions). 

The scraper considers only the most recent one (https://github.com/opencivicdata/python-legistar-scraper/blob/master/legistar/bills.py#L285)

In the above example, an earlier version has “MatterTextPlain,” but the later version does not.

reginafcompton:
It seems like, for any case of a missing title, we can revert to the use of `identifier` +  `bill_type` - particularly, for board reports that do not have a subject, e.g., MINUTES.

reginafcompton:
@shrayshray : I'll summarize the above, here! Let me know if you have any questions.

To extract a board report title (i.e., the SUBJECT header in the board report PDF), we need the full text of that board report. This text resides on the API - and it can have multiple "versions." Our scraper looks at the most recent version. 

However, in a small collection of cases, the Metro board reports have text in the _earliest_, but not in the _most recent_ version. I suggest communicating this to whomever manages this side of the data entry. 

**Example**
API for [this board report](https://boardagendas.metro.net/board-report/2016-0741/) for a life of project budget: http://webapi.legistar.com/v1/metro/matters/3534/

This link lists the three versions of the board report text: http://webapi.legistar.com/v1/metro/matters/3534/versions

This link shows the earliest text (which has the necessary "MatterTextPlain"): http://webapi.legistar.com/v1/metro/matters/3534/texts/4356

This link shows the most recent text (which does not have the necessary "MatterTextPlain"): http://webapi.legistar.com/v1/metro/matters/3534/texts/5100

[...]

PR #196, opened 2017-08-11
Board report text

reginafcompton:
This pull request address issue #128 and #177

### DIFF
--- a/lametro/search_indexes.py
+++ b/lametro/search_indexes.py
@@ -40,13 +40,12 @@ class LAMetroBillIndex(BillIndex, indexes.Indexable):
                     for el in array_with_subject:
                         if 'SUBJECT:' in el:
                             results = item.replace('\n', '')
-
-        # Isolate text after 'SUBJECT'
-        if results:
-            before_keyword, keyword, after_keyword = results.partition('SUBJECT:')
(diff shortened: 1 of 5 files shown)


lametro/search_indexes.py
```diff
@@ -27,30 +27,26 @@ def prepare_last_action_date(self, obj):
...
+                            results = item.replace('\n', '')
+
+        # Isolate text after 'SUBJECT'
+        if results:
+            before_keyword, keyword, after_keyword = results.partition('SUBJECT:')
+            if after_keyword:
+                # Do not corrupt search results with brackets
+                if '[PROJECT OR SERVICE NAME]' not in after_keyword and '[DESCRIPTION]' not in after_keyword and '[CONTRACT NUMBER]' not in after_keyword:
+                    return after_keyword.strip()
+
```
fgregg: it's not strictly necessary (since `return` is a kind of `GOTO` but, it's easier to read if there's an else here.

reginafcompton: All right - makes sense. I added an ELSE for clarity. how does this look?

reginafcompton:
@fgregg - I refactored the redundant logic into a utils file. How does this look?

PR #197, opened 2017-08-11
Correct format for bill url

reginafcompton:


reginafcompton:
Bu g fix for bill document

Issue #154, opened 2017-05-19
Home and Events: Tag when Metro makes a change to an event


reginafcompton:
We will look at the "updated_at" (also "created_at") field for an event. If Metro updates the event within, say, two weeks of the meeting, then the event will get a tag or bold font or something to indicate that the event has been modified. 

@shrayshray Can you let us know the time boundaries for this, e.g., two weeks before an event? or keep the tag for 72 hours after an event gets updated?

shrayshray:
After an agenda is final (i.e., the PDF is available), if there are any changes to it, it should be highlighted in some way and display the text "Updated" next to the date/time.

reginafcompton:
@shrayshray 

Did we reach a consensus on when an event should be highlighted (e.g., if a change is made one, two, three weeks before a meeting)? 

Also: Moving this to June issues.

shrayshray:
@reginafcompton Yes -- it should be highlighted if there is a change to the date or time, or if the meeting is cancelled, after the Agenda is final/PDF is available.

reginafcompton:
To accomplish this we need to do the following:

* add an "updated_at" field in the EventDocument in django-councilmatic: https://github.com/datamade/django-councilmatic
* in the event detail view, add a comparison between the EventDocument updated_at and the Event updated_at

reginafcompton:
@shrayshray - we are wrapping up this issue, today. Here's a screenshot of what to expect, when an event gets updated, after the agenda has been added. 

<img width="936" alt="screen shot 2017-08-16 at 11 57 18 am" src="https://user-images.githubusercontent.com/13078679/29375418-8b3604e0-827a-11e7-9f2b-fa777aa2c362.png">


Quick clarification: This updated tag will only appear for _upcoming events_, not past events. Is that right?

shrayshray:
Perfect, thank you!

PR #202, opened 2017-08-16
Event update

reginafcompton:
@fgregg - this pull request address issue #154. It includes a couple tests, using pytest. (N.B. I had some simple tests for manual agenda upload, written with django unittest. So, I had to make some large deletes to get pytests up-and-running.)


lametro/templatetags/lametro_extras.py
```diff
@@ -144,4 +144,16 @@ def parse_agenda_item(text):
...
\ No newline at end of file
+        return ''
+
+
+@register.filter
+def updates_made(event_id):
+    event = Event.objects.get(ocd_id=event_id)
+
+    try:
+        document = EventDocument.objects.get(event_id=event_id)
```
fgregg: There is only one document for every event?

lametro/templatetags/lametro_extras.py
```diff
@@ -144,4 +144,16 @@ def parse_agenda_item(text):
...
+    except EventDocument.DoesNotExist:
+        return False
+    else:
+        return True if document.updated_at < event.updated_at else False
```
fgregg: It's clearer to do this like

```python
return document.updated_at < event.updated_at
```

and even clearer to do

```python
updated = document.updated_at < event.updated_at
return updated 
```

lametro/templatetags/lametro_extras.py
```diff
@@ -144,4 +144,16 @@ def parse_agenda_item(text):
         label, number = text.split(',')
         return number
     else: 
-        return ''
\ No newline at end of file
+        return ''
+
+
+@register.filter
+def updates_made(event_id):
```
fgregg: Please add a docstring to explain why we want to know when an event was updated after a document associated with that event?

reginafcompton:
@fgregg - changes made to the template tag. Let me know how it looks!

Issue #238, opened 2017-11-15
Home: "Listen Live" button link


shrayshray:
On the "Listen Live" button that shows on the Home page during meetings, we currently use a general link to the media player. The Board Chair has mentioned there's an issue with this because the general media player shows video streaming suggestions instead of the meeting Agenda, and this has caused concern from the public.
Example:
![ex1](https://user-images.githubusercontent.com/28063988/32865600-a4ef0162-ca18-11e7-99ae-0edb7401207b.jpg)
Omar may have found a solution. If it sounds feasible to you to implement it, we'd like to make it a high priority. 
Possible solution:
It looks like the live meeting link from InSite uses the GUID for the meeting in the player. We can get this from the API as shown below and generate the link that would be used for each meeting.  

System and safety GUID: B0A322B8-71C3-4197-8719-E9C3E8980D95
9 AM
http://metro.granicus.com/mediaplayer.php?event_id= B0A322B8-71C3-4197-8719-E9C3E8980D95

Construction GUID: 985A2FD9-06D6-4EB0-A384-27779F62AE01
10:15 AM
http://metro.granicus.com/mediaplayer.php?event_id=985A2FD9-06D6-4EB0-A384-27779F62AE01

Executive Management GUID: 3DEE4755-B576-4BCD-93BE-39FB381BAF81
11:30 AM
http://metro.granicus.com/mediaplayer.php?event_id=3DEE4755-B576-4BCD-93BE-39FB381BAF81

Examples:
 ![ex2](https://user-images.githubusercontent.com/28063988/32865821-b4e6bea6-ca19-11e7-9119-51ca23c0edfc.png)
![ex3](https://user-images.githubusercontent.com/28063988/32865855-cb38c7f8-ca19-11e7-92ef-77ee0719a895.png)

reginafcompton:
@shrayshray - this seems like a possible solution. We'll need to build out the functionality to scrape/import the GUID (which we currently do not do). Otherwise, adding the link itself should not be a heavy task. As you suggest, we'll make this issue top priority for November.

reginafcompton:
@shrayshray, we have code ready to implement this enhancement. A couple questions:

[...]

PR #239, opened 2017-11-28
Create media url from GUID, and refactor current event logic

reginafcompton:
This PR handles issue #238.

It also includes some much-needed refactoring for getting current events.


lametro/models.py
```diff
@@ -185,65 +187,23 @@ def upcoming_board_meeting(cls):
...
-        # return cls.objects.filter(start_time__gt=faketime)\
-        #           .filter(name__icontains="Board Meeting")\
-        #           .order_by('start_time').first()
 
     @classmethod
     def current_meeting(cls):
-        # For testing...
-        # meeting_time = datetime.now(app_timezone) - timedelta(days=8) + timedelta(hours=2) + timedelta(minutes=13)
-        # meeting_end_time = meeting_time - timedelta(hours=3)
-        # print(meeting_time, "TIMEEEE")
```
hancush: :-D

hancush:
Let's look at this together tomorrow. I'm not sure I understand the logic.

lametro/models.py
```diff
@@ -185,65 +187,23 @@ def upcoming_board_meeting(cls):
...
-        
+        # A board meeting lasts 3 hours, and a committee event lasts 1 hour.
+        # Find all events occuring in the next three hours, and then filter accordingly. 
         meeting_time = datetime.now(app_timezone) + timedelta(minutes=6)
         meeting_end_time = datetime.now(app_timezone) - timedelta(hours=3)
```
hancush: Your comment says find events occurring in the _next_ three hours.

`meeting_time` is six minutes from now. `meeting_end_time` is three hours ago.

So `found_events` is getting `Event` objects that start between three hours ago and six minutes from now, e.g., the last three hours, rather than the next three hours. Is this what we want?

reginafcompton: Yes, that's right. We can discuss this tomorrow - Metro had a lot of requests to get the logic like it is....it's also why the code ended up so verbose and needed refactoring.

hancush: Sounds good to me!

PR #240, opened 2017-11-29
Spaced crons

fgregg:
This increases the spacing of crontasks until we can figure out https://github.com/datamade/devops/issues/41. (I also removed crontasks from blackbox since they aren't particularly sensitive).


reginafcompton:
looks good @fgregg

PR #241, opened 2017-11-29
bring back more frequent import_datas

fgregg:
Since I've addressed: https://github.com/datamade/devops/issues/41

we can bring back more frequent import datas


reginafcompton:
@fgregg - this is terrific. (Are we okay turning off pglogical? I am not quite sure I understand what it did for the OCD API server....)

PR #248, opened 2017-12-08
Put jetty config under version control

reginafcompton:
Another step for this issue: https://github.com/datamade/django-councilmatic/issues/157

@evz, I did this for deployment. Do you think it's okay to restart jetty with every Metro deployment? It seems like that could cause unwanted blips. 

```
mv -f $APP_PATH/configs/jetty.conf /etc/default/jetty
chown root.root /etc/default/jetty
service jetty restart
```


evz:
Yeah let's see if there's an option to just reload the config without restarting.

reginafcompton:
@evz - in the Upstart file I only see options for: start, stop, restart, supervise, run|demo, and check (which just "echos" the various config arguments). What do you think about requiring a manual restart for this step? (I cannot imagine needing to adjust jetty configurations very often.)

evz:
Ooo, hey why don't we try this: https://wiki.apache.org/solr/CoreAdmin#RELOAD

As part of the deployment, just visit that URL using curl or something.

reginafcompton:
@evz - this loads a new schema file for Solr 4, so we should implement this as part of the deployment...but does it reload Jetty configuration?

evz:
Nope. I suppose if we want to do that, we'll want to do a manual restart. Changes like that seem a bit more fundamental.

reginafcompton:
New deployment pattern:

```
# Move jetty config files to correct place
mv -f $APP_PATH/configs/jetty.conf /etc/default/jetty
chown root.root /etc/default/jetty

# Reload nginx and supervisor files, so it recognizes updated configs
service nginx reload
supervisorctl update 

# Move solr schema to correct place and reload
cp $APP_PATH/solr_scripts/schema.xml /opt/solr/multicore/$SOLR_CORE/conf/
wget "http://localhost:8983/solr/admin/cores?action=RELOAD&core=$SOLR_CORE"
```

Issue #205, opened 2017-09-07
Enhancement: Ability to manually upload Agendas to Councilmatic site


shrayshray:
Ability to upload Agendas/Items to the Councilmatic site directly when Legistar is down.

reginafcompton:
Additional notes about this issue

**The problem**
At the end of August, Granicus received several reports about partial outages in Legistar and other Granicus services. See [here for a summary](http://status.granicusops.com/incidents/hknntk4pjj43) of the incidents. The underlying cause was a cut fiber line in Chicago.

The Granicus [incident history for Legistar](https://granicus.statuspage.io/history) displays a handful of problems every year, e.g., "Slowness," "Cannot login." Granicus, fortunately, consistently addresses issues in **a matter of hours**. 

Still, some of these issues have the potential to delay processes in Councilmatic, for example: (1) a Metro admin cannot login to Legistar and upload a meeting agenda, or (2) the Councilmatic scrapers fail, if trying to access Legistar during an outage. 

**Possible solution**
Metro Councilmatic includes an option to upload an actual PDF, which builds upon our current solution. An admin could upload _an agenda link_ or an actual document: https://github.com/datamade/la-metro-councilmatic/issues/170 

**Questions to consider and potential risks**
Where will we host uploaded PDFs (e.g., an `uploads` directory at the root of LA Metro)? 
Should we remove uploaded PDFs, once Legistar recovers, to prevent data duplication? 

Uploading documents to Councilmatic comes with some risks. Broadly, it creates another access point for adding data and, in turn, undermines the ideal relationship between Councilmatic and its upstream data sources. We presently have an integrated pipeline, which funnels data from a remote source into the Councilmatic site. With an additional data entry option, the model breaks down, creating more opportunities for points of failure. (Indeed, even with the well-tested agenda upload, we encountered [some unexpected behaviors](https://github.com/datamade/la-metro-councilmatic/issues/189). )

I suggest we think further about the value, necessity, and consequences of this enhancement, before implementation.

@derekeder and @fgregg - do you have additional thoughts on this?

shrayshray:
Omar and I had an extensive discussion about this based on the potential risks outlined above. We're trying to balance: 

1. Concern with the manual upload option causing any compromise to the system integrity, and, in a sense, undermining the nature of Councilmatic's relationship with Legistar. This is clearly articulated and legitimate.
[...]

PR #249, opened 2017-12-19
December

reginafcompton:
This PR handles issue #205 

Some issues of note:
- [x] to render a newly uploaded file requires running collect static. Django has suggestions for doing this: https://docs.djangoproject.com/en/2.0/howto/static-files/deployment/#serving-static-files-in-production
- [x] we'll also need to think about permissions for the uploaded files - they get assigned "datamade:datamade," but we might want "datamade:www-data"
- [x] persist PDF uploads after deployment - minimize stuff that happens during the mv to `tmp`
- [x] DELETE should remove the pdf from the `static/pdf`
- [x] can you download the static file?
- [x] remember to update this document: https://docs.google.com/document/d/1mRjCsCSsPtuBM2VO5DW1rU0KEcaLcxWNXIsWlY9QEno/edit


lametro/templates/lametro/event.html
```diff
@@ -55,9 +55,9 @@ <h4>Participants</h4>
                     </p>
                     {% endif %}
               
-                    {% if agenda_url or uploaded_agenda_url %}
+                    {% if agenda_url or uploaded_agenda_url or uploaded_agenda_pdf %}
 
-                    {% if uploaded_agenda_url and not agenda_url%}
+                    {% if uploaded_agenda_url or uploaded_agenda_pdf and not agenda_url%}
```
hancush: I'm not sure about Django templates but Python evaluates `and` before `or`, so this line would be like `if uploaded_agenda_url or (uploaded_agenda_pdf and not agenda_url)`. 

If we mean to say `if there is an uploaded URL or PDF, and there is not an agenda URL`, I think we need to put the `or` clause in parentheses.

lametro/templates/lametro/event.html
```diff
@@ -108,19 +108,64 @@ <h4>Minutes</h4>
...
+                                  <i class="fa fa-file-text" aria-hidden="true"></i> Upload an Agenda PDF
+                                </a>
+                              </h4>
+                            </div>
+                            <div id="collapsePDF" class="panel-collapse collapse {% if pdf_form.errors %}in{% endif %}" role="tabpanel" aria-labelledby="headingTwo">
+                              <div class="panel-body">
+                                {% if legistar_ok %}
+                                    <div class="alert alert-info" role="alert"><i class="fa fa-exclamation-circle" aria-hidden="true"></i> <strong>Caution!</strong> Our system does NOT detect a Legistar outage. Do not use this option.</div>
+                                {% else %}
+                                    <div class="alert alert-info" role="alert"><i class="fa fa-exclamation-circle" aria-hidden="true"></i> Our system detects a Legistar outage. Procede with this option but exercise caution.</div>
```
hancush: Procede -> Proceed

lametro/templates/lametro/event.html
```diff
@@ -108,19 +108,64 @@ <h4>Minutes</h4>
...
+                            <div class="panel-heading" role="tab" id="headingTwo">
+                              <h4 class="panel-title">
+                                <a class="collapsed" role="button" data-toggle="collapse" data-parent="#accordion" href="#collapsePDF" aria-expanded="false" aria-controls="collapsePDF">
+                                  <i class="fa fa-file-text" aria-hidden="true"></i> Upload an Agenda PDF
+                                </a>
+                              </h4>
+                            </div>
+                            <div id="collapsePDF" class="panel-collapse collapse {% if pdf_form.errors %}in{% endif %}" role="tabpanel" aria-labelledby="headingTwo">
+                              <div class="panel-body">
+                                {% if legistar_ok %}
```
hancush: Nice!

lametro/templates/lametro/event.html
```diff
@@ -158,20 +203,25 @@ <h4>Agenda upload</h4>
...
-            encoded_url = encodeURIComponent(url)
-            doc_url = base_url + '?filename=agenda&document_url=' + encoded_url
+        // Admin view: Cancel submission of PDF
+        $('#pdf-form-cancel').on('click', function() {
+            document.getElementById('pdf-check-viewer-test').src = '';
+            document.getElementById('pdf-form-input').value= null;
 
-            return doc_url 
-        };
+            $('#pdf-form-message').addClass('hidden');
```
hancush: You can select multiple IDs like `$('#pdf-form-message, #pdf-check-viewer-test...').addClass('hidden');`

lametro/tests/test_lametro.py
```diff
@@ -5,14 +5,36 @@
...
+    '''
+
+    with open('lametro/tests/test_agenda.pdf', 'rb') as agenda:
+        agenda_file = agenda.read()
+
+        agenda_pdf_form = AgendaPdfForm(files={'agenda': SimpleUploadedFile('test_agenda.pdf', agenda_file, content_type='application/pdf')})
+    
+        assert agenda_pdf_form.is_valid() == True
+
+
```
hancush: Might be good to test a non-PDF file does not validate, as well?

lametro/templates/lametro/event.html
reginafcompton: This can be simplified to: `{% if not agenda_url%}`...since we check for `uploaded_agenda_url or uploaded_agenda_pdf` in previous template logic.

lametro/templates/lametro/event.html
reginafcompton: https://www.youtube.com/watch?v=LukyMYp2noo

lametro/views.py
```diff
@@ -238,33 +240,31 @@ def get_context_data(self, **kwargs):
...
     event_doc = EventDocument.objects.filter(event_id=event.ocd_id, note__icontains='Manual upload')
-    for e in event_doc: 
+
+    for e in event_doc:
+        # Remove stored PDF from Metro app.
+        if 'Manual upload PDF' in e.note:
+            try:
+                os.remove('lametro/static/%s' % e.url )
+            except OSError:
+                pass
```
hancush: Is there any reason we might want to log this?

scripts/before_install.sh
```diff
@@ -7,11 +7,15 @@ mkdir -p /home/datamade/la-metro-councilmatic
 
 if [ "$DEPLOYMENT_GROUP_NAME" == "staging" ]
 then
+    # Preserve uploaded PDFs
+    mv /home/datamade/lametro-staging/lametro/static/pdf/agenda-*.pdf /tmp
```
hancush: Presumably the change to `after_install` is these getting moved back? (Juuuuuuust checking! 😇)

reginafcompton: Yep!

lametro/views.py
reginafcompton: No - indeed, I am being a little overcautious here, since a PDF should not reside in the static files, if a corresponding entry for an EventDocument does not exist in the database. But...just in case a PDF disappears or somehow does not get created, then the Metro admin will not see an error when they try to delete a mis-uploaded agenda.

lametro/templates/lametro/event.html
```diff
@@ -55,9 +55,11 @@ <h4>Participants</h4>
                     </p>
                     {% endif %}
               
+                    <!-- Check that the database has an agenda to display -->
                     {% if agenda_url or uploaded_agenda_url or uploaded_agenda_pdf %}
 
-                    {% if uploaded_agenda_url or uploaded_agenda_pdf and not agenda_url%}
+                    <!-- An event can have both manually uploaded and auto-imported agendas, i.e., if a Metro Admin uploads an agenda and then import_data runs. Give the Admin an option to delete the manually uploaded agenda, but ONLY if an auto-imported agenda does not exist. -->
```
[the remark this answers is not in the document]
reginafcompton: Yes, we'd need to manually delete the uploaded agendas...however, having an EventDocument with a "Manual upload" note does not interfere with the functionality of the site itself. (The conditionals favor the `agenda_url`.) So, a manually uploaded PDF could live in perpetuity in Metro Councilmatic without causing any harm.

With that said, we could automate this with a script + crontask, but automating the deletion process might be more time-consuming than having to manually delete a PDF in the very rare occasion that Metro uses this feature. Thoughts?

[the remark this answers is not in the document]
hancush: That seems fair to me!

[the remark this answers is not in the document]
hancush: (To not automate, for now.)

Issue #251, opened 2018-01-18
Home: Current meeting not showing


shrayshray:
This happened during the 1/18/18 Construction Committee, which began at 10:15. When the issue was reported and I checked the site, it was 11:18; and although the meeting was still running, it wasn't showing in the "Current" slot. I hadn't checked the site to confirm the meeting was indeed up prior to 11:18, but I assume it was based on Regina's explanation:
"Presently, the logic for current meetings is:
- a committee meeting lasts one hour
- a board meeting lasts three hours"
This is in place to account for meetings which occur in succession.

@reginafcompton your suggestion for a fix sounds good: show a committee meeting as "current," until six minutes before next meeting begins. Then, on the days with successive meetings, the "Current meeting" view will display the entire time.
Thank you!

reginafcompton:
@shrayshray - what is the maximum duration of a committee meeting? 

I need to deal with events that occur at the end of a succession of meetings - e.g., [when would the meeting at 1:00 on 1/18 conclude](https://boardagendas.metro.net/event/ad-hoc-customer-experience-committee-aad6c70c626f/) - 90 minutes? less?

(For the others, I am checking the start time of the next meeting...and building out the logic from there.)

shrayshray:
@reginafcompton I think 120 minutes would be safe for maximum meeting duration. 
Does the check for the next meeting just check the next meeting's scheduled time, or does it check whether the next meeting has started, i.e., the live link is available?
If it's possible, it would be ideal to set a 2 hour max, but also have checks for whether the current meeting is still running, and whether the next meeting has started, take precedence for determining when to stop showing the current meeting link.

reginafcompton:
@shrayshray - the logic looks at the start time of the next meeting. For example: This [9:00 am meeting](https://boardagendas.metro.net/event/system-safety-security-and-operations-4a0e05784b62/) would stop showing as current at 10:09am, when [this meeting would show as current](https://boardagendas.metro.net/event/construction-committee-0560d127d6f9/).

**Some other notes and questions**

-  If I remember correctly, the status of an event depends on [the presence of an audio link on the Legistar calendar, during the scheduled meeting time](https://metro.legistar.com/calendar.aspx). _Is that correct?_ 

[...]

PR #252, opened 2018-01-22
Refine current meeting logic

reginafcompton:
Handles #251 

@hancush - this PR profoundly overhauls the logic for retrieving a current meeting and the entire test suite. Some important changes:
- simplify the template logic to only expect a queryset
- use Legistar audio link as a indicator of an "In Progress" meetings (I have some concerns about this...but I'd like to test it on the staging site regardless)
- refactor tests in the style of LargeLots

Happy to chat about it on Monday!


.pytest_cache/v/cache/lastfailed
```diff
@@ -0,0 +1,10 @@
+{
```
hancush: I think you'll want to add `.pytest_cache/` to your `.gitignore` and delete this directory & the files in it.

lametro/models.py
```diff
@@ -191,29 +191,58 @@ def upcoming_board_meeting(cls):
 
     @classmethod
     def current_meeting(cls):
-        # Create the boundaries for discovering events (in progess) within the timeframe stipulated 
-        # by Metro.
-        # Then, filter the events according to these boundaries.
-        six_minutes_from_now = datetime.now(app_timezone) + timedelta(minutes=6)
-        three_hours_ago = datetime.now(app_timezone) - timedelta(hours=3)
+        # QA Testing....
+        # fakenow = datetime(2018,1,18,10,10)
```
hancush: Safe to strike this commented code?

lametro/models.py
```diff
@@ -191,29 +191,58 @@ def upcoming_board_meeting(cls):
...
+        # six_minutes_from_now = fakenow + timedelta(minutes=6)
+        # three_hours_ago = fakenow - timedelta(hours=6)
+
+        # We look for everything in the last 6 hours, because the maximum recorded meeting is 5.38 hours (according to the spreadsheet provided by Metro in issue #251).
+        six_minutes_from_now = datetime.now(app_timezone) + timedelta(minutes=5)
```
hancush: `five_minutes_from_now`

lametro/models.py
```diff
@@ -191,29 +191,58 @@ def upcoming_board_meeting(cls):
...
+        three_hours_ago = datetime.now(app_timezone) - timedelta(hours=6)
```
hancush: `six_hours_ago`

lametro/models.py
hancush: Actually should L194-L207 be here at all?

https://github.com/datamade/la-metro-councilmatic/blob/022b541525f26bcd1c685666f3fef966b71a11cf/lametro/models.py#L194-L207

I'd define `six_hours_ago` / `five_minutes` from now below the multi-line comment, and strike everything else.

lametro/models.py
```diff
@@ -191,29 +191,58 @@ def upcoming_board_meeting(cls):
...
-                # Change the value of "_hour_ago" for committee meetings to 1 hour.
-                if found_events.filter(name__icontains='Board Meeting'):
-                    return found_events.first()
-                else:
-                    one_hour_ago = six_minutes_from_now - timedelta(hours=1)
-                        
-                    return found_events.filter(start_time__gt=one_hour_ago).first()
+            return calculate_current_meetings(found_events, six_minutes_from_now)
+
+        # Solution 2: rely entirely on Legistar.
```
hancush: Did you mean to leave this in here?

lametro/templates/partials/meeting_details_current.html
```diff
@@ -0,0 +1,60 @@
...
+                    {% if current_meeting|length > 1 %}
+                        <p><a class="btn btn-salmon" href="{{media_url}}" target="_blank"><i class="fa fa-headphones" aria-hidden="true"></i> Listen</a></p>
+                    {% else %}
+                        <div class="row">
+                            <div class="col-xs-5">
+                                <p><a class="btn btn-salmon" href="{{media_url}}" target="_blank"><i class="fa fa-headphones" aria-hidden="true"></i> Listen</a></p>
+                            </div>
+                            <div class="col-xs-7">
+                                {% if current_meeting.first.documents.all %}
+                                    <a class="btn btn-teal" id="pdf-download-link" target='_blank' href='{#current_meeting.first.documents.all|find_agenda_url}}'><i class='fa fa-fw fa-download'></i> Get Agenda PDF</a>
```
hancush: `{#` should be double curlies, no?

lametro/utils.py
```diff
@@ -28,37 +29,50 @@ def parse_subject(text):
...
-            found_events = found_events.filter(start_time__gt=one_hour_ago)
-
-            if len(found_events) > 1:
-                return found_events
-            else:
-                return found_events.first()
-        else:
-            return found_events.first()                
\ No newline at end of file
+    # Metro provided a spreadsheet of average meeting times. The minimum average time a meeting lasts is 52 minutes: let's round to 55 and remove the 5-minute buffer.
```
hancush: Does that mean `five_minutes_from_now` in the `current_meeting` should just be now?

lametro/utils.py
```diff
@@ -28,37 +29,50 @@ def parse_subject(text):
...
+                if legistar_meeting_progress(event):
+                    return found_events
+        else: 
+            if legistar_meeting_progress(found_events.last()):
+                return Event.objects.filter(ocd_id=found_events.last().ocd_id)
+
+        return Event.objects.none()
+
+
+def legistar_meeting_progress(event):
```
hancush: Can I get a brief docstring here? Something like "The Legistar organization detail page flags ongoing meetings as 'In progress.' Ping this page to check whether a planned event is indeed happening at the stated date and time."

lametro/utils.py
```diff
@@ -28,37 +29,50 @@ def parse_subject(text):
...
+    if found_events.filter(start_time__gte=time_ago):
+        # Check if previous event is still going on in Legistar.
+        previous_meeting = found_events.filter(start_time__gte=time_ago).last().get_previous_by_start_time()
+        if legistar_meeting_progress(previous_meeting):
+            return Event.objects.filter(ocd_id=previous_meeting.ocd_id)
+
+        return found_events.filter(start_time__gte=time_ago)  
+    else:  
+        # The IF statement handles the below cases:
+        # (1) found_events includes just one event object. Example: the first committee meeting of the day - 9:00 am on 01/18/2018
```
hancush: Can we just check whether `len(found_events) == 1` and return the event or None (if it's not "In progress"), prior to doing any of the other checks?

.pytest_cache/v/cache/lastfailed
reginafcompton: Good catch!

lametro/models.py
reginafcompton: Yep - that's all junk.

lametro/models.py
reginafcompton: Yes, but I just moved it to a comment in the original issue.

lametro/utils.py
reginafcompton: Well, Metro still wants a meeting to appear fives minutes before its start time...so, actually my comment should say: "and add the 5-minute buffer."

lametro/utils.py
reginafcompton: Well, if it's not 'In progress,' then we need to check if a previous event is. That means potentially pinging Legistar twice on one page load. I am trying to avoid that.

lametro/models.py
```diff
@@ -191,29 +191,32 @@ def upcoming_board_meeting(cls):
...
+        (2) the previous meeting has ended (determined by looking for "In progress" on Legistar).
+
+        The maximum recorded meeting duration is 5.38 hours, according to the spreadsheet provided by 
+        Metro in issue #251.
+        So, to determine initial list of possible current events, we look for all events scheduled 
+        in the past 6 hours.
+
+        This method returns a list (with zero or more elements).
+        '''
+        # fakenow = datetime(2018,1,18,12,55)
```
hancush: shall we strike this?

reginafcompton: Ohhhh, yes.

Issue #269, opened 2018-03-12
`sort_name` not parsed correctly for Board Report 2015-0344


reginafcompton:
Board Report 2015-0344 returns a very long string of text for its `sort_name`. This means that something in the indexing logic does not behavior the way it's designed. 

Take a look at the utility functions called here: https://github.com/datamade/la-metro-councilmatic/blob/master/lametro/search_indexes.py#L33
https://github.com/datamade/la-metro-councilmatic/blob/master/lametro/utils.py#L8

reginafcompton:
Resolved!

PR #272, opened 2018-03-13
Improve logic to retrieve sort name

reginafcompton:
Handles issue #269


lametro/utils.py
```diff
@@ -16,24 +17,17 @@ def format_full_text(full_text):
...
-# Isolate text after 'SUBJECT'
 def parse_subject(text):
-    if text:
-        before_keyword, keyword, after_keyword = text.partition('SUBJECT:')
-        if after_keyword:
-            if '[PROJECT OR SERVICE NAME]' not in after_keyword and '[DESCRIPTION]' not in after_keyword and '[CONTRACT NUMBER]' not in after_keyword:
-                return after_keyword.strip()
-
-    return None
+    if ('[PROJECT OR SERVICE NAME]' not in text) and ('[DESCRIPTION]' not in text) and ('[CONTRACT NUMBER]' not in text):
```
hancush: Great improvement. :-)

lametro/utils.py
```diff
@@ -16,24 +17,17 @@ def format_full_text(full_text):
...
-        txt_as_array = full_text.split("..")
-        for item in txt_as_array:
-            if 'SUBJECT:' in item:
-                array_with_subject = item.split('\n\n')
-                for item in array_with_subject:
-                    if 'SUBJECT:' in item:
-                        results = item.replace('\n', '')
+        full_text_with_cue = full_text.replace('\n\n', 'NEWLINE').replace('\r\n', 'NEWLINE').replace('\n..', 'NEWLINE')
+
+        match = re.search('(SUBJECT:)(.*?)(NEWLINE|ACTION:)', full_text_with_cue.replace('\n', ' '))
```
hancush: Why do this replace here, instead of in the `full_text_with_cue` definition?

hancush:
@reginafcompton I see your point, but if we are debugging an unanticipated parsing failure, I wouldn't immediately think to check the tests for examples. I also think it's kind of hard to follow regexes, if I don't know what the string might look like.

lametro/utils.py
reginafcompton: Yes, it seems like something like this would be better:

```
clean_full_text = full_text.replace('\n\n', 'NEWLINE').replace('\r\n', 'NEWLINE').replace('\n..', 'NEWLINE').replace('\n', ' ')

match = re.search('(SUBJECT:)(.*?)(NEWLINE|ACTION:)', clean_full_text)
```

Honestly, I think there must be a cleverer use of RegEx to get the desired results. Have you had any luck with MULTILINE? https://docs.python.org/3.6/library/re.html#re.MULTILINE That presumably would eliminate the need for `replace`....but it did not behave as I expected.

hancush: you know, i'm really terrible at regex and [tend to avoid them](http://regex.info/blog/2006-09-15/247).

what is the intent here? it seems like you are just looking for "SUBJECT" and either "NEWLINE" or "ACTION".

could you do `'SUBJECT' in clean_full_text and any(token in clean_full_text for token in ['NEWLINE', 'ACTION'])` or something like that?

hancush: are new lines meaningful?

reginafcompton: Yes, I fear RegEx, too. [My previous solution](https://github.com/datamade/la-metro-councilmatic/blob/master/lametro/utils.py#L15) completely avoided RegEx, but it also missed some edge cases and was more difficult to decipher. 

I added some comments to give you a sense about how newlines help parse the subject header. @hancush - happy to chat about it!

lametro/utils.py
```diff
@@ -11,29 +12,33 @@
...
     if full_text:
-        txt_as_array = full_text.split("..")
-        for item in txt_as_array:
-            if 'SUBJECT:' in item:
-                array_with_subject = item.split('\n\n')
-                for item in array_with_subject:
-                    if 'SUBJECT:' in item:
-                        results = item.replace('\n', '')
+        clean_full_text = full_text.replace('\n\n', 'NEWLINE').replace('\r\n', 'NEWLINE').replace('\n..', 'NEWLINE').replace('\n', ' ')
+        match = re.search('(SUBJECT:)(.*?)(NEWLINE|ACTION:)', clean_full_text)
```
hancush: I think moving that last replace to the line above (and seeing examples in the docstring!) really helped with the clarity here. Text parsing sucks, haha. Nice changes, all around.

Issue #255, opened 2018-02-14
Set-up redirect for person page


reginafcompton:
It appears that at some point we had an entry for Carrie Bowen with this slug: https://boardagendas.metro.net/person/carrie-bowen-42b40b4f42e1/ A user recently tried to access her page with it.  

(Her correct URL: https://boardagendas.metro.net/person/carrie-bowen-af5a8ab39aad/)

Set up a redirect for such scenarios. [We did something similar in NYC.](https://github.com/datamade/nyc-council-councilmatic/blob/master/nyc/views.py#L26)

PR #278, opened 2018-03-20
Given bad slug, redirect to proper Person page

hancush:
Addresses #255.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,6 +8,16 @@ from councilmatic_core.models import Event, EventDocument, Bill
 from lametro.models import LAMetroPerson
 
 
+def get_uid_chunk(uid=None):
+    '''
+    Create the UID chunk like the one we append to slugs to ensure
+    they're unique.
+    '''
+    if not uid:
+        uid = str(uuid4())
+
+    return uid[:13].replace('-', '')
+
 @pytest.fixture
 @pytest.mark.django_db
 def bill(db):
@@ -81,12 +91,11 @@ def metro_person(db):
     class LAMetroPersonFactory():
         def build(self, **kwargs):
             uid = str(uuid4())
-            slug_uid = uid.replace('-', '')[:13]
 
             person_info = {
                 'ocd_id': 'ocd-person/' + uid,
                 'name': 'Wonder Woman',
-                'slug': 'wonder-woman-' + slug_uid,
+                'slug': 'wonder-woman-' + get_uid_chunk(uid),
             }
 
             person_info.update(kwargs)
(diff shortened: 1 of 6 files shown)


tests/test_person.py
```diff
@@ -0,0 +1,55 @@
...
+
+from lametro.views import LAMetroPerson
+
+
+@pytest.mark.django_db
+def test_person_page_redirects(client, metro_person, mocker):
+    person = metro_person.build()
+
+    # Patch a bunch of model attributes, because patching get_context_data
+    # breaks the url routing, for some reason.
```
reginafcompton: Ugh. That's annoying. I do not have an explanation for why that might be.

tests/test_person.py
```diff
@@ -0,0 +1,55 @@
...
+    person_of_same_name = metro_person.build()
+
+    url = reverse('person', kwargs={'slug': partial_slug})
+    rv = client.get(url)
+
+    assert rv.status_code == 404
+
+    # Assert navigating to a slug that does not exist returns a 404.
+
+    url = reverse('person', kwargs={'slug': 'the-hulk'})
```
reginafcompton: So, LargeLots is animal themed. Metro will be Marvel themed!

tests/test_person.py
```diff
@@ -0,0 +1,55 @@
...
+    # Assert navigating to a valid slug works.
+
+    url = reverse('person', kwargs={'slug': person.slug})
+    rv = client.get(url)
+
+    assert rv.status_code == 200
+
+    # Assert navigating to a partially valid slug, redirects to a valid slug.
+
+    partial_slug = '-'.join(person.slug.split('-')[:-1])
```
reginafcompton: I think testing with a partial_slug works well....but why not test with a partially correct slug instead, since that would more closely mirror the "real-world" case?

```
# Something like....
old_uuid = str(uuid4())
old_slug = partial_slug + old_uuid.replace('-', '')[:13]
```

tests/conftest.py
```diff
@@ -8,6 +8,16 @@
...
 
+def get_uid_chunk(uid=None):
+    '''
+    Create the UID chunk like the one we append to slugs to ensure
+    they're unique.
+    '''
+    if not uid:
+        uid = str(uuid4())
+
+    return uid[:13].replace('-', '')
```
reginafcompton: Perfect!

hancush:
merging! 

honk

reginafcompton:
what species of bird is that? it's too big to be native to Chicago. could be a dinosaur.

hancush:
definitely a dinosaur @reginafcompton

Issue #273, opened 2018-03-14
Finesse mobile view for Current Meeting


reginafcompton:
The mobile view does not look terrific. Let's hide the text on the buttons (i.e., only show icons) and think about changing or hiding the photo.

<img width="390" alt="screen shot 2018-03-14 at 1 58 35 pm" src="https://user-images.githubusercontent.com/13078679/37425082-2747950a-2790-11e8-905e-5373170a56b2.png">

camachoo:
In the mobile view, can we spread the button real estate to include the area under the picture? Example below:

Picture

Meeting info

Button real estate



—

Omar Camacho
Metro Los Angeles
Sr. Manager, Systems Projects, Research & Records Information Management

213.922.2335
metro.net | facebook.com/losangelesmetro | @metrolosangeles
Metro provides excellence in service and support.

reginafcompton:
That's a nice idea - I'll put something together. Should we include this in the March milestone?

shrayshray:
@reginafcompton If you think there is time to accomplish this in March, then yes, let's add to March milestones.

reginafcompton:
<img width="397" alt="37996920-faf6c7e2-31de-11e8-9f43-252f89085740" src="https://user-images.githubusercontent.com/13078679/38097114-1c77b83e-333a-11e8-9efc-9eee1bc6fa97.png">

Closing!

PR #284, opened 2018-03-27
Mobile fixes

beamalsky:
Fixes for Issue #273 and improved nav bar scaling. The new mobile Current Events view looks like they wanted, but I'm not sure the changes I made were the most straightforward.

screen shot 2018-03-27 at 4 49 46 pm

### DIFF
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -191,6 +191,12 @@ class LAMetroEvent(Event):
 
     @classmethod
     def current_meeting(cls):
+        
+        #use this line to hardcode one current event
+        return Event.objects.filter(start_time='2017-06-15 13:30:00-05')
+        #use this line to hardcode two current events
+        # return Event.objects.filter(start_time='2017-11-30 11:00:00-06')
+
         '''
         Create the boundaries for discovering events (in progess) within the timeframe stipulated 
         by Metro. A meeting displays as current if:
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -191,11 +191,6 @@ class LAMetroEvent(Event):
 
     @classmethod
     def current_meeting(cls):
-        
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -1,4 +1,4 @@
-psycopg2==2.6.2
+psycopg2==2.7.1
```
reginafcompton: You recently merged this change to master: you should not have to duplicate it here! Try to keep your local master up-to-date with the remote master, especially before checking out a new branch. Let me know if you have questions about this.

lametro/static/css/city_custom.css
```diff
@@ -258,7 +262,7 @@ div#toggleControls { margin-bottom: 1em; }
     }
 }
 
-@media (min-width: 767px) and (max-width: 991px) {
+@media (min-width: 767px) and (max-width: 1264px) {
```
reginafcompton: Nice catch!

lametro/models.py
```diff
@@ -191,6 +191,12 @@ def upcoming_board_meeting(cls):
 
     @classmethod
     def current_meeting(cls):
+        
+        #use this line to hardcode one current event
+        #return Event.objects.filter(start_time='2017-06-15 13:30:00-05')
+        #use this line to hardcode two current events
+        #return Event.objects.filter(start_time='2017-11-30 11:00:00-06')
```
reginafcompton: I'd actually like to see these comments consolidated with the "docstring" (lines 201 ff.)....something, like, "To hardcode current event(s) for testing, then return these querysets...." (etc etc etc).

requirements.txt
beamalsky: 👌🏻

Issue #266, opened 2018-03-08
[META] Reports/Agendas - Ability to search attachment text


shrayshray:
Implementing system that allows users to search for board reports based on the
text in attachments. 
We are discussing internally and will provide response on whether this functionality should be the default search type, or whether it will be an optional addition to the standard search, e.g., use a "search attachments" checkbox next to the search bar.

reginafcompton:
**Steps to add this search functionality**

(1) write a custom script that iterates over all attachments and converts Word, PDF, and Excel files into plain text. Run this script against the entire Metro database, and ultimately, integrate it into the regular data pipeline (i.e., run it every time we refresh data).

(2) configure and rebuild the search indexes to enable searching by attachment text - this entails creating a custom search template, ([as we did with NYC](https://github.com/datamade/nyc-council-councilmatic/blob/master/nyc/templates/search/indexes/nyc/bill_text.txt)), and possibly [adding/preparing a new field to the search index](https://github.com/datamade/la-metro-councilmatic/blob/master/lametro/search_indexes.py).

reginafcompton:
**Questions to answer:**
- how much of a performance hit, after we add all the new text? 
- check or uncheck attachments: search for all (by default), just attachments, or do not search attachments? 

DataMade will have a first implementation of this issue by the end of April.

hancush:
FYI – We'll need to push the completion of this issue to the May milestone.

hancush:
**Road map:** 

1. We should add attachment text as a distinct field for search (to enable optional searching), to [the `BillIndex` search index](https://github.com/datamade/la-metro-councilmatic/blob/68f42bf1530c91ae9ce87cc3eccdbc5c69dff925/lametro/search_indexes.py#L10).
    - What's the best way to do this: Concatenate all attachment text and search via bill, or create parallel index for bill documents? (The "best" way will be informed by the answers to the next two questions.)
    - Is the purpose of searching attachment text, to surface the attachment, or the bill?
    - What is the default search functionality (bill text only, attachment text only, or bill text and attachment text)?
 
2. Then, most likely, we will need to extend / refactor [the `LAMetroCouncilmaticFacetedSearchView` view](https://github.com/datamade/la-metro-councilmatic/blob/68f42bf1530c91ae9ce87cc3eccdbc5c69dff925/lametro/views.py#L705) to alter the `SearchQueryset` to search different fields, depending on user input.

hancush:
Re: our earlier questions,

[...]

PR #286, opened 2018-04-09
Add attachment converter to regular pipeline

reginafcompton:
This PR turns on cron and includes the attachment conversion script as part of the regular data pipeline. Related to https://github.com/datamade/la-metro-councilmatic/issues/266


reginafcompton:
Good question - yes, we do not run compile pdfs on staging, since we consider the production site to have the more authentic data. We do not have a staging version of metro-pdf-merger...so the compile script sends all data to the same place, e.g., https://github.com/datamade/la-metro-councilmatic/blob/master/lametro/management/commands/compile_pdfs.py#L81 (Happy to message about this further!)

Issue #263, opened 2018-03-07
Get and display the Spanish language audio


reginafcompton:
Metro will soon include Spanish-language broadcasting of events. 

DataMade will scrape these links and include them on the Boardagendas site. They will appear as a distinct button (during a current meeting) or url (on the Events page and Event detail page) - with text in Spanish language, e.g., "Escuchar."

Metro will add some test events in Legistar in April or May, and DataMade will do a test run with this data.

shrayshray:
This is high priority for April (Chair's request). We're still waiting on the device from Granicus, and internal hardware setup to support this, but have some additional info for reference which may help Datamade get a head start on initial development so we're closer to "ready-to-go" once we have everything in place an are ready to deploy ...

Spanish Language broadcast of meetings is handled as a second meeting by Granicus. See example from LA City's Granicus site: https://www.lacity.org/your-government/audiovideo/council-meeting-video
![lac-sap](https://user-images.githubusercontent.com/28063988/38119815-6e408148-3379-11e8-9790-c467c38e2e63.PNG)

Metro displays broadcast links on Legistar site, not Granicus site: https://metro.legistar.com/Calendar.aspx
Mockup of potential display:
![lam-sap-mockup](https://user-images.githubusercontent.com/28063988/38119846-93fc5b78-3379-11e8-9b8d-46c0b19b7ad6.PNG)

@reginafcompton @hancush could you remind me of the prefix used for test meetings, so they are excluded from display on the live site? I can set up some fake/test meetings once I have this info. Thank you!

reginafcompton:
@shrayshray - we did not yet decide on a prefix. We could identify test events in several ways, but I think the best strategy would be to list the `EventLocation` as "TEST".

(Since the event name correlates with the committee, I'd rather not use that to identify test events - to avoid unexpected data cascades.)

First, we'll add a filter on Councilmatic to hide these test events.
After that, you can populate Legistar with some test data.

A question - will these test events be visible on Legistar?

shrayshray:
@reginafcompton Yes, these events will be visible on Legistar.

hancush:
Hi, @shrayshray, just an FYI that I'm getting started on this, this week. I'll be in touch in the next couple of days about adding test data!

hancush:
# The Plan

## Overview

[...]

PR #287, opened 2018-04-09
hide test events on production, but leave them visible on staging

hancush:
related to #263.

adds a `SHOW_TEST_EVENT` setting that hides events with an event location of TEST when set to `False`, so we don't show test data from legistar.


evz:
So, this looks great. I'm not able to look at the encrypted files, tho, so can @jeancochrane double check that?

jeancochrane:
How about we add you to the keyring as part of this PR @evz? I don't have much context for what's going on here.

hancush:
That is indeed the plan @evz. Thanks!

PR #295, opened 2018-05-01
display english and spanish event media links correctly

hancush:
this pr:

- [x] pins la-metro to the latest version of django councilmatic to take advantage of multiple related `EventMedia`
- [x] updates template logic [here](https://github.com/datamade/la-metro-councilmatic/blob/master/lametro/templates/lametro/event.html#L17) and [here](https://github.com/datamade/la-metro-councilmatic/blob/master/lametro/templates/partials/past_event_day.html#L38) to iterate over and display `event.media_urls` rather than only one `event.media_url`
- [x] refactors `extra_context` [here](https://github.com/datamade/la-metro-councilmatic/blob/master/lametro/views.py#L49-L67) to grab both english and spanish live media links  (and make logic more obvious: a link to live media appears in legistar when a meeting is ongoing)
- [x] update all listen link text to "watch in english" and "ver en español" 

addresses #263


hancush:
blurgh, we don't import [all of the sources](https://github.com/datamade/django-councilmatic/blob/bdeff46b1e7c3e89131ba579f9976dbe72da0f62/councilmatic_core/management/commands/import_data.py#L1501) into `councilmatic`, so an additional source would require changes there, too.

hancush:
we also only add the web detail link if it exists, otherwise we just add a link to the calendar. it doesn't seem to exist, until after the meeting has passed.

fgregg:
How would we get the GUID if the meeting detail url didn't exist? Is that knowable some other way?

hancush:
@fgregg it's on the api event; we [save it in the extras](https://github.com/datamade/scrapers-us-municipal/blob/6ebb252441669f13890179b5ba14606bf977547f/lametro/events.py#L165), and [import it](https://github.com/datamade/django-councilmatic/blob/bdeff46b1e7c3e89131ba579f9976dbe72da0f62/councilmatic_core/management/commands/import_data.py#L1509) in django-councilmatic. 

we could save the spanish guid in the extras, as well, but we'd still need to make a change in django-councilmatic to facilitate > 1 event guid.

fgregg:
> @fgregg it's on the api event; we save it in the extras, and import it in django-councilmatic.

> we could save the spanish guid in the extras, as well, but we'd still need to make a change in django-councilmatic to facilitate > 1 event guid.

This seems sensible to me.

It also seems to me that we should add the spanish api event url and, when available, the spanish legistar meeting url to the OCD data when available, even if we don't import that data into councilmatic.

lametro/models.py
```diff
@@ -189,12 +192,68 @@ def get_queryset(self):
...
+        else:
+            return fallback
+
+
+    @property
+    def english_live_media_url(self):
+        guid = self.extras['guid']
+        english_url = self.BASE_MEDIA_URL + 'event_id={guid}'.format(guid=guid)
+
+        return self._valid_or_generic(english_url,
```
fgregg: For me this would be clearer if you _valid_or_gener was just _valid, and then you handled the fallback case here.

PR #300, opened 2018-05-09
Run the tests from Travis build

hancush:
This PR:

- Repairs the tests by [fixing a bug](https://github.com/datamade/la-metro-councilmatic/pull/300/files#diff-1214be6822cebe9a6279d8f370c70d30R62) in `calculate_current_events` that [only considered meetings for 45 minutes](https://github.com/datamade/la-metro-councilmatic/compare/fix_tests?expand=1#diff-1214be6822cebe9a6279d8f370c70d30L48), rather than 55 minutes, past their start time.
  - I've also [made "now" an optional kwarg](https://github.com/datamade/la-metro-councilmatic/pull/300/files#diff-1214be6822cebe9a6279d8f370c70d30R46) so we can [test the logic](https://github.com/datamade/la-metro-councilmatic/pull/300/files#diff-7491ccb4a5e2c1045607c6b5b5811018R198) with an arbitrary datetime.
- Runs the tests in `travis.yml`.
- Revises docstrings for `current_meeting`, `calculate_current_meeting`, and `test_current_committee_meeting_first`.

### DIFF
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -265,17 +265,22 @@ class LAMetroEvent(Event, LiveMediaMixin):
 
     @classmethod
     def current_meeting(cls):
-
         '''
-        Create the boundaries for discovering events (in progess) within the timeframe stipulated
-        by Metro. A meeting displays as current if:
-        (1) "now" is five minutes or less before the designated start time;
-        (2) the previous meeting has ended (determined by looking for "In progress" on Legistar).
+        Discover and return events in progress.
 
-        The maximum recorded meeting duration is 5.38 hours, according to the spreadsheet provided by
-        Metro in issue #251.
-        So, to determine initial list of possible current events, we look for all events scheduled
+        The maximum recorded meeting duration is 5.38 hours, according to the
+        spreadsheet provided by Metro in issue #251. So, to determine initial
+        list of possible current events, we look for all events scheduled
         in the past 6 hours.
+
+        A meeting displays as "current" if:
+
+        (1) it started in the last six hours, or it starts in the next five
+            minutes (determined by this method); and
+        (2) it started less than 55 minutes ago, and the previous meeting has
+            ended (determined by `calculate_current_meetings`); or
+        (3) Legistar indicates it is in progress (detemined by `calculate_
+            current_meetings`).
 
         This method returns a list (with zero or more elements).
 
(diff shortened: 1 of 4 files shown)


lametro/utils.py
```diff
@@ -43,33 +43,57 @@ def parse_subject(text):
...
+    earliest_start = found_events.earliest('start_time').start_time
+    latest_start = found_events.latest('start_time').start_time
+
+    if found_events.filter(start_time__gte=fifty_five_minutes_ago):
+        previous_meeting = found_events.filter(start_time__gte=fifty_five_minutes_ago)\
+                                       .last()\
+                                       .get_previous_by_start_time()
+
         if legistar_meeting_progress(previous_meeting):
             return LAMetroEvent.objects.filter(ocd_id=previous_meeting.ocd_id)
```
evz: The way that I am reading this is if the previous meeting is still in progress, return that one instead of that one plus the current one? Is that correct?

lametro/utils.py
```diff
@@ -43,33 +43,57 @@ def parse_subject(text):
...
+        # n.b., In reality, concurrently scheduled events happen one after the
+        # other, but we want the current meeting display to line up with the
+        # scheduled events (i.e., if it's 9:15, show all of the events slated
+        # for 9, even if only one is technically in progress).
         for event in found_events:
             if legistar_meeting_progress(event):
                 return found_events
-    else: 
+
+    else:
```
evz: The `if` here seems to be if there are multiple events that started in the last 55 minutes, the `elif` seems to be, as you descibe, there are multiple events scheduled to happen at the same time (or if there is only one). What is the `else` for? Are we ever going to follow this path or is it just defensive programming?

tests/test_events.py
```diff
@@ -91,95 +91,113 @@ def test_updates_made_false(event, event_document):
...
+        'name': 'Planning and Programming Committee',
         'start_time': '2018-01-17 2:00:00',
         'slug': 'planning-and-programming-committee-f8e135045661'
     }
     event.build(**planning_meeting_info)
 
     safety_meeting_info = {
         'ocd_id': 'ocd-event/5e84e91d-279c-4c83-a463-4a0e05784b62',
-        'name': 'System Safety, Security and Operations Committee', 
+        'name': 'System Safety, Security and Operations Committee',
```
evz: Does it matter that there is no `slug` here?

lametro/models.py
```diff
@@ -287,12 +292,12 @@ def current_meeting(cls):
 
         five_minutes_from_now = datetime.now(app_timezone) + timedelta(minutes=5)
         six_hours_ago = datetime.now(app_timezone) - timedelta(hours=6)
-        found_events = cls.objects.filter(start_time__lte=five_minutes_from_now, start_time__gte=six_hours_ago)\
+        found_events = cls.objects.filter(start_time__gte=six_hours_ago, start_time__lte=five_minutes_from_now)\
```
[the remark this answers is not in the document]
hancush: no, i just think it's easier to grok like lower limit <= start time <= upper limit

tests/test_events.py
hancush: hmmmm i don't think so

lametro/utils.py
hancush: that's the existing logic, correct. it feels sort of at odds with trying to mirror the calendar in the next block, but that's how it works now...

lametro/utils.py
hancush: it happens if you have multiple events scheduled in a day, that don't start at the same time. so like, if it's 2 p.m., and there are 9 a.m. and 12 p.m. events, they'll both come through as found events, and we test each one to see which is happening now (if any). 

i think there is a simpler way to write this logic, but i'm not sure i want to do that, in this pr... do you have any immediate thoughts?

PR #304, opened 2018-05-17
Refactor current meeting logic

hancush:
This PR:

- Refactors `LAMetroEvent.current_meeting`
- Removes methods superseded by the refactor from `lametro.utils`
- Refactors tests for `current_meeting` functionality


lametro/models.py
```diff
@@ -263,41 +264,95 @@ def upcoming_board_meeting(cls):
...
+        Hit the endpoint, and return the corresponding meeting, or an empty
+        queryset.
+        '''
+        running_events = requests.get('http://metro.granicus.com/running_events.php')
+
+        for guid in running_events.json():
+            # We get back two GUIDs, but we won't know which is the English
+            # audio GUID stored in the 'guid' field of the extras dict. Thus,
+            # we iterate.
+            meeting = cls.objects.filter(extras__guid=guid.upper())
```
fgregg: seeing `guid.upper()` suggests something is pretty wrong upstream?

lametro/models.py
```diff
@@ -263,41 +264,95 @@ def upcoming_board_meeting(cls):
...
-        To hardcode current event(s) for testing, use these examples:
-        return LAMetroEvent.objects.filter(start_time='2017-06-15 13:30:00-05')
-        return LAMetroEvent.objects.filter(start_time='2017-11-30 11:00:00-06')
+
+    @classmethod
+    def current_meeting(cls):
         '''
-        from .utils import calculate_current_meetings  # Avoid circular import
+        If there is a meeting scheduled to begin in the last six hours or in
+        the next five minutes, hit the running events endpoint. If there is a
```
fgregg: This is a pretty good explanation of what the code is doing, but it would be even better if it explained why.

In particular, if there is not streaming_meeting, then why do want to show the meetings that started within the last 20 minutes.

Is the idea that streaming_meetings is not completely reliable to indicate that a a meeting is happening right now?

lametro/models.py
hancush: guids are all-uppercase [in the api](http://webapi.legistar.com/v1/metro/events/915), but are all-lowercase in the running events endpoint. i'd prefer not to alter the source data in the scrape, so it seems appropriate to do a transformation for this instance-specific comparison. happy to hear an alternate perspective!

fgregg: yikes, okay. add a comment that explains why this thing is happening.

lametro/models.py
hancush: no, per metro, events "regularly" start (e.g., show up as streaming) later than their scheduled start time. if we are only looking for meetings five minutes in the future, a meeting that does not start on time will not appear as current. 

so, if there is not a streaming meeting, we also look for meetings scheduled to start up to 20 minutes ago (i consulted with omar on this buffer), so meetings that start late are still returned as current in the interim.

hancush: @fgregg i expounded on this in the docstring

fgregg: Okay, that's the kind of thing that would be good to have in your comment. This code would be clearer to me if you used if elif else constructs instead of returns.

```python
     if scheduled_meetings:
            streaming_meeting = cls._streaming_meeting()

            if streaming_meeting:
                current_meetings = streaming_meeting
            else:
                # explanation that sometimes a meeting stream shows up late
                twenty_minutes_ago = cls._time_ago(minutes=20)

                # '.annotate' adds a boolean field, 'is_board_meeting'. We want to
                # show board meetings first, so order in reverse, since False (0)
                # comes before True (1).
                current_meetings = scheduled_meetings.filter(start_time__gte=twenty_minutes_ago)\
                                     .annotate(is_board_meeting=RawSQL("name like %s", ('%Board Meeting%',)))\
                                     .order_by('-is_board_meeting')
      else:
           current_meetings = cls.objects.none()

      return current_meetings
```

PR #307, opened 2018-05-25
attachment text search

hancush:
# to-do

- [x] break search bar into partial upstream in django-councilmatic
- [x] pin metro to latest version of django-councilmatic
- [x] override the search bar partial to include toggle for "search all" v "search reports only"
- [x] fix behavior of buttons (though they're supposed to be radio buttons, more than one is selected in the get request)


lametro/templates/base.html
```diff
@@ -110,6 +110,20 @@
 
     {% block extra_js %}{% endblock %}
 
+    <script type="text/javascript">
+        $('label.btn').click(function toggleRadio(e) {
+            // Toggle "active" appearance first, so filtering on active works.
+            $(this).button('toggle');
```
evz: I've never seen `.button` before. Is that a jquery thing? Is it toggling a class or something?

lametro/templates/base.html
```diff
@@ -110,6 +110,20 @@
 
     {% block extra_js %}{% endblock %}
 
+    <script type="text/javascript">
+        $('label.btn').click(function toggleRadio(e) {
```
evz: Since this is on the base template, is using `label.btn` as a selector going to have unintended consequences on some pages?

solr_scripts/schema.xml
```diff
@@ -78,10 +76,10 @@
         <filter class="solr.LowerCaseFilterFactory"/>
         <filter class="solr.EnglishPossessiveFilterFactory"/>
         <filter class="solr.KeywordMarkerFilterFactory" protected="protwords.txt"/>
-        <filter class="solr.EnglishMinimalStemFilterFactory"/>
         <!-- Optionally you may want to use this less aggressive stemmer instead of PorterStemFilterFactory:
-          <filter class="solr.PorterStemFilterFactory"/>
+          <filter class="solr.EnglishMinimalStemFilterFactory"/>
```
evz: We should change the stemmers back to what they were. It might not make that big of a difference but since just in case, it's good to be consistent.

solr_scripts/schema.xml
```diff
@@ -215,6 +215,4 @@
 
   <!-- SolrQueryParser configuration: defaultOperator="AND|OR" -->
   <solrQueryParser defaultOperator="AND"/>
-  <similarity class="solr.BM25SimilarityFactory"/>
```
evz: Lets add the BM25 similarity scorer back in. The default for this version of solr is TFIDF which is not as good.

lametro/templates/base.html
hancush: it's [a bootstrap thing](https://getbootstrap.com/docs/3.3/javascript/#buttons)

hancush: i can add it in a comment!

PR #331, opened 2018-07-17
Do not alter Choate's title

reginafcompton:


lametro/templatetags/lametro_extras.py
```diff
@@ -66,7 +66,7 @@ def full_text_doc_url(url):
 def appointment_label(label):
```
hancush: i think a docstring outlining the different cases ("we get these types of inputs, and want these types of outputs...") would make this a little easier to grok! can i come back to this, after you've added that?

Issue #334, opened 2018-07-18
Home: Flag canceled events - High Priority


shrayshray:
1. Flag canceled events on the home page as "canceled" (similar to the solution on the comprehensive meetings view). 
2. On the meeting page, increase the size of and make more prominent the "canceled" flag. 
3. If possible, also use strike-through on the meeting name on the meeting page, home page, and comprehensive meeting page.

reginafcompton:
@shrayshray - I implemented the above suggestions. See below for examples of the three affected pages.

## Examples

**Meetings: I moved the "Cancelled" label to the far right and increased its size.**
<img width="620" alt="screen shot 2018-07-19 at 11 04 57 am" src="https://user-images.githubusercontent.com/13078679/42955683-abff361a-8b43-11e8-8e81-4f54fd0c591f.png">

***

**Meeting detail: I moved the "Cancelled" label inline with the event name**
<img width="831" alt="screen shot 2018-07-19 at 11 21 26 am" src="https://user-images.githubusercontent.com/13078679/42956613-0ed66158-8b46-11e8-8ea4-d6618c9c069a.png">

***

**Home**
<img width="727" alt="screen shot 2018-07-19 at 11 39 48 am" src="https://user-images.githubusercontent.com/13078679/42957508-7fc4c92a-8b48-11e8-83be-c44700ed5fa9.png">

shrayshray:
@reginafcompton this looks great and ready to go live!

reginafcompton:
It's live - [example event](https://boardagendas.metro.net/event/finance-budget-and-audit-committee-e911cf6f4bcb/).

Closing!

PR #335, opened 2018-07-19
Cancel flag

reginafcompton:
This PR handles issue #334


reginafcompton:
@hancush - could you take a look today? (It's a high priority issue for Metro, but fortunately only involves template changes.)

Issue #336, opened 2018-07-19
"Acting" should appear next to Shirley Choate's post


reginafcompton:
Shirley Choate holds the Metro post "Caltrans District 7 Director, Appointee of Governor of California", though she is the "Acting Caltrans District 7 Director" (on behalf of Carrie Bowen). 

We should handle this by adding some information to the "extras" field in the scraper, and then, importing and rendering it appropriately.

shrayshray:
Correcting Shirley Choate's title ASAP is high priority; but would it be possible to add the ability for Metro to make these type of changes via the Content Management System referenced in Issue #265 for future updates?

reginafcompton:
Here's the plan:
1. Assign a value to the extras field when getting [Post](https://github.com/opencivicdata/pupa/blob/c0dbe820821caa66f1cb02675dfdcd3df22307b7/pupa/scrape/schemas/post.py) objects [in the scraper](https://github.com/opencivicdata/scrapers-us-municipal/blob/master/lametro/people.py#L77) – this value might be `extras = {"title-prefix": "Acting"}`
2. Add an extras JSONfield to the [Post object](https://github.com/datamade/django-councilmatic/blob/master/councilmatic_core/models.py#L599) in Councilmatic. 
3. Get the extras in [`import_data`](https://github.com/datamade/django-councilmatic/blob/master/councilmatic_core/management/commands/import_data.py#L731) when importing posts
4. Filter for extras when displaying a the post of a person: committees page, committee detail pages, and person/people pages.

reginafcompton:
@shrayshray - you can see a fix for this on the staging site. I'll deploy it to production in the morning, unless you see a reason not to.

This type of change should not be handled by the CMS. It seems very simple, I know! However, if we want to maintain the paradigm of the OpenCivicData standard, then the Councilmatic site should not change a board member's post, title, or role, without that change being accurately reflected in the OCD API. As you can see, the change for this originates pretty far upstream, i.e., by populating the `extras` field in the membership model, but that insures that the data exists consistently and correctly in all places. 

Handled by https://github.com/datamade/la-metro-councilmatic/pull/352

PR #352, opened 2018-10-08
Membership extras

reginafcompton:
This PR handles the last part of https://github.com/datamade/la-metro-councilmatic/issues/336

### DIFF
--- a/lametro/views.py
+++ b/lametro/views.py
@@ -524,6 +524,7 @@ class LACommitteeDetailView(CommitteeDetailView):
             sql = ('''
               SELECT
                 p.name, p.slug, p.ocd_id,
+                m.extras,
                 array_agg(m.role) as committee_role,
                 array_agg(mm.label::VARCHAR)
                 FILTER (WHERE mm.label is not Null) as label,
@@ -545,7 +546,7 @@ class LACommitteeDetailView(CommitteeDetailView):
               WHERE m.organization_id = %s
               AND m.end_date::date > NOW()::date
               GROUP BY
-                p.name, p.slug, p.ocd_id
+                p.name, p.slug, p.ocd_id, m.extras
               ORDER BY last_name
             ''')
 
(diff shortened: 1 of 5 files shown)


lametro/views.py
```diff
@@ -423,7 +424,7 @@ def get_context_data(self, *args, **kwargs):
             WHERE m.organization_id='ocd-organization/42e23f04-de78-436a-bec5-ab240c1b977c'
             AND m.end_date >= '{0}'
             AND m.person_id <> 'ocd-person/912c8ddf-8d04-4f7f-847d-2daf84e096e2'
-            GROUP BY p.ocd_id, p.name
+            GROUP BY p.ocd_id, p.name, m.extras
```
evz: Maybe this is a nonsensical question but if a person had more than one post, would this return 2 rows for that person because of the `extras` field being part of the `GROUP BY`? Does that even happen? Does it matter?

lametro/views.py
```diff
@@ -545,7 +547,7 @@ def get_context_data(self, **kwargs):
               WHERE m.organization_id = %s
               AND m.end_date::date > NOW()::date
               GROUP BY
-                p.name, p.slug, p.ocd_id
+                p.name, p.slug, p.ocd_id, m.extras
```
evz: Same question as above

lametro/views.py
```diff
@@ -598,7 +601,7 @@ def get_context_data(self, **kwargs):
...
```
evz: Same

lametro/views.py
reginafcompton: that's true, but the same person should not have more than one post – so, if that happens, then something is wrong with the data upstream.

reginafcompton: it seems to me like the Councilmatic site should not mask such data errors? what do you think?

evz: Yeah that seems reasonable. I'm not really sure what else we'd do in this case.

Issue #351, opened 2018-09-22
Committee Pages:  Move Chief Executive Office outside members list


shrayshray:
Metro's Chief Executive Office (currently, Phil Washington), is not a board member. He attends all Board meetings, but he's not an official member; and although he does not vote in committees or regular board meetings, he's also not technically a "non-voting member" (like Shirley Choate, for example). Is it possible to create a separate section on committee pages, either below or to the right of the list of official committee members, for the CEO, based on his "Title" as "Chief Executive Officer" and without his "Committee Role"? And exclude him from the committee member list? 
![image](https://user-images.githubusercontent.com/28063988/45910897-e0a44580-bdc1-11e8-983f-e4d277e060f3.png)

reginafcompton:
Steps to take:

- [ ] for Washington's memberships, change the `role` to "Chief Executive Officer" (accomplished via the scraper)
- [ ] create new table/div with Washington's role, name, and link to detail page

reginafcompton:
@shrayshray - I have a solution ready to go, but before I deploy it to staging, I wanted to ask a couple questions:

(1) the expectation is that the `/committees` page now shows one fewer member, right? 

<img width="820" alt="screen shot 2018-10-09 at 4 05 36 pm" src="https://user-images.githubusercontent.com/13078679/46698705-65be9580-cbdd-11e8-95a5-76d2f204f971.png">

(2) I put the "CEO" subsection directly beneath the members table. What do you think? I'd like to get a photo of Phil Washington, so that it's not so bare. Would that be possible?

<img width="1181" alt="screen shot 2018-10-09 at 4 06 04 pm" src="https://user-images.githubusercontent.com/13078679/46698822-aae2c780-cbdd-11e8-969f-a9b3eef0e939.png">


(3) Would you like to see something different on the `person` detail page for Phil Washington? The table states "Member of," but that might be misleading?

<img width="1258" alt="screen shot 2018-10-09 at 4 10 18 pm" src="https://user-images.githubusercontent.com/13078679/46698934-f9906180-cbdd-11e8-8487-de3522e9bb60.png">

shrayshray:
@reginafcompton 
(1) Yes, exactly. The member count displaying as 7 instead of 6 is what prompted the request for this change.
(2) Great! Looks very good, but I agree it would be better to have a headshot for consistency. Here ya go:
[...]

PR #353, opened 2018-10-09
PersonDetail revision

reginafcompton:
This PR handles issues #351, and also refactors how we find a person's memberships on the PersonDetail.


lametro/views.py
```diff
@@ -689,8 +693,31 @@ def get_context_data(self, **kwargs):
...
+                JOIN councilmatic_core_organization AS o
+                ON o.ocd_id = m.organization_id
+                WHERE m.person_id = %s
+                AND m.end_date::date > NOW()::date
+                AND m.organization_id != %s
+                ORDER BY
+                    CASE
+                        WHEN m.role='Chair' THEN 0
+                        WHEN m.role='Vice Chair' THEN 1
+                        WHEN m.role='Member' THEN 2
```
evz: The role is never null I take it? I seem to remember PostgreSQL putting NULLs first by default.

lametro/templates/lametro/person.html
```diff
@@ -157,7 +157,7 @@ <h3>
 
             {% if request.GET.view == 'committees' %}
                 <h3>Committees</h3>
-                {% if person.chair_role_memberships or person.member_role_memberships %}
+                {% if memberships_list %}
```
evz: So much nicer! Seems like there must be less overhead for coming up with this list than the previous way, too.

reginafcompton: yeah, perhaps! I just needed template code that was easier to digest.....

lametro/views.py
reginafcompton: okay - that's a good question. Looks like the role cannot be Null, since the scraper guarantees that (i.e., assigning a value of "Member" of "Non-voting member"): https://github.com/opencivicdata/scrapers-us-municipal/blob/master/lametro/people.py#L117

So, this query should work as expected.

Issue #345, opened 2018-09-11
Board Reports: Unpublished reports showing on website!


shrayshray:
A bunch of reports which are not yet published are showing up on the site. See the first 8 reports listed: https://boardagendas.metro.net/search/
E.g., 2018-0435: https://boardagendas.metro.net/board-report/2018-0435/ Status in Legistar is “Agenda Ready”. It is NOT available on metro.legistar.com. The Agenda has not yet been published and is not available: https://boardagendas.metro.net/event/executive-management-committee-8000f2384368/

shrayshray:
Follow up on this issue ... thank you so much, @reginafcompton, for resolving it right away.
The reports which were showing up in advance of their agenda being published (they are on agendas for 9/19 and 9/20) had been manually changed in Legistar by unchecking the box "Not Viewable Via InSite". 

Six of the 8 reports were General Public Comment, which users previously manually set to "Not Viewable ...", and then unchecked weeks/months later when it was decided the policy would be to show them. So for this month's General Public Comment reports, the users got mixed up and assumed they always needed to manually uncheck the box for these reports. The other two reports were changed within about 10 minutes on the same day, so I'm assuming the user just went with the flow he/she got into with this process.

So this was a workflow issue on the Metro side ... users not clear that the box is checked by default _before_ the report's agenda is published, but it will be viewable once the agenda is published - unless the box is checked _after_ publication (and we alert Datamade to this change). We're clarifying the appropriate workflow with our users. Here's what I'm wondering: Would it be possible for the Councilmatic site to follow the same logic for viewable/hidden as InSite, and first check whether the report's agenda has been published? These reports were not visible on InSite while they were visible on the Councilmatic site, because InSite is looking at whether the reports are on a published agenda.

reginafcompton:
Thanks for this detailed report @shrayshray. I have a couple questions before suggesting a solution:

1. Can you tell me about the `MatterAgendaDate` for a Board Report on the Legistar API (e.g., http://webapi.legistar.com/v1/metro/matters/5204)? Specifically, does a Board Report have a `MatterAgendaDate`, _only_ when the agenda has been published? 

2. It looks like the `MatterStatusName` can be "Agenda Ready", even when the agenda is not published. Can you confirm? (I was looking at [this upcoming event](http://webapi.legistar.com/v1/metro/events/1422/eventitems), which contains reports listed as "Agenda Ready," though the Agenda is still a draft.)

shrayshray:
[...]

PR #357, opened 2018-10-17
Viewable reports

reginafcompton:
This PR handles the last portion of issue #345


lametro/models.py
```diff
@@ -78,6 +78,29 @@ def get_last_action_date(self):
...
+
+        (2) Does the Bill have a classification of "Board Box"? Then, show it.
+
+        (3) Is the Bill on a published agenda, i.e., an event with the status of "passed" or "cancelled"? Then, show it.
+        '''
+        if self.bill_type == "Board Box":
+            return True
+
+        events_with_bill = Event.objects.filter(agenda_items__bill_id=self.ocd_id)
+        passed_events = [event for event in events_with_bill if (event.status == 'passed' or event.status == 'cancelled')]
```
hancush: will this array ever become very large? since we only care about the presence, but not how many, etc., it may be more memory-efficient to do something like:

```python
for event in events_with_bill:
    if event.status in ('passed', 'cancelled'):
        return True
return False
```

lametro/views.py
```diff
@@ -739,6 +739,8 @@ def search(self):
             # Don't auto-escape my query! https://django-haystack.readthedocs.io/en/v2.4.1/searchqueryset_api.html#SearchQuerySet.filter
             sqs = sqs.filter_or(attachment_text=Raw(self.cleaned_data['q']))
 
+        sqs = sqs.filter(viewable=True)
```
hancush: if we always hide bills, should we index them at all?

hancush: (i.e., do we have all the info we need to omit them when we build the index?)

hancush: (i forget, does our current pipeline update the index with every scrape, or with enough frequency for this suggestion to make sense?)

lametro/models.py
reginafcompton: No, this list will never be very large –  a bill may be associated with one or two events. More than that does not align with what Metro has done historically.

lametro/views.py
reginafcompton: I asked myself this very question! 

I decided to index "hidden" bills for one reason – [the integrity script](https://github.com/datamade/django-councilmatic/blob/master/councilmatic_core/management/commands/data_integrity.py), which checks that the Councilmatic database and the Solr index have the same number of bills. If we do not index a "hidden" bill, then this script will complain about a missing bill in the Solr index. I think it's easier to modify the SearchQuerySet, rather than finding a work around with how the integrity script or `import_data` works. 

(N.B. `data_integrity` runs after every execution of `update_index`, which runs after `import_data` which runs [four times per hour.](https://github.com/datamade/la-metro-councilmatic/blob/master/scripts/lametro-crontasks))

hancush: ha, great minds. :-) that makes sense to me. maybe it would be good a drop a comment or docstring in the `search_indexes.py` file?

reginafcompton:
Ah, that `Manager` is slick. 

I thought about all the places where we render bills, and the only instance where we render _all_ bills is on the search view (which uses the sqs object). For other places, we show bills based on whether or not they have actions (i.e., [the person detail](https://github.com/datamade/la-metro-councilmatic/blob/master/lametro/views.py#L690)) or appear on an agenda (i.e.,[ the event detail](https://github.com/datamade/la-metro-councilmatic/blob/master/lametro/views.py#L148)). 

I do think I need to account for the later case. 
For the former case, I do not see how a bill that does not appear on a published agenda could have any actions. Does that sound right?

hancush:
@reginafcompton i think that that sounds right, but that a manager may still be a good idea, unless there will ever be a case where we want to show non-viewable bills.

reginafcompton:
@hancush, if we add a context manager, as you've done with the Event, then we'd end up adding the bills to the Solr index (which is not desirable, c.f. above). 
Haystack uses the Django ORM to retrieve bills: https://github.com/django-haystack/django-haystack/blob/master/haystack/indexes.py#L154 
The `_default_manager` uses the first manager defined in the class: https://docs.djangoproject.com/en/2.1/topics/db/managers/

I suspect there might be a way to build an LAMetro bill manager, so that Haystack ignores it. But, again, I cannot think of a use case in which we risk showing "hidden" bills (other than the search page)...so, again, I am wondering if we need the manager at all. Do you want to chat about this further?

PR #383, opened 2018-12-01
Don't return future events as last action dates

hancush:


hancush:
@reginafcompton should this have a test?

reginafcompton:
A test would be much appreciated! [(Here, methinks.)](https://github.com/datamade/la-metro-councilmatic/blob/master/tests/test_bills.py)

Do you want me to pick this up, or do want to do the testing?

Thanks, @hancush !

hancush:
i can do it, @reginafcompton!

hancush:
@reginafcompton tell me about it!!! those docs sure pay dividends! ^_^

PR #388, opened 2019-01-15
for past events, english is the first media option

### LINKED ISSUE #354: Meetings: List English video link first
The meeting video links listed on the [Events page](https://boardagendas.metro.net/events/) should always list "Watch in English" before  "Ver en Espanol."
image

jmithani:
This PR edits the past events to show the option to view media in English first. It uses [this nifty filter](https://docs.djangoproject.com/en/2.1/ref/templates/builtins/#dictsortreversed).

screen shot 2019-01-15 at 4 05 24 pm

Closes #354.

### DIFF
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -408,7 +408,7 @@ class LAMetroEvent(Event, LiveMediaMixin):
 
     @property
     def media(self):
-        ## makes sure that 'Watch in English' appears first
+        '''Return related LAMetroEventMedia objects such that English audio appears first'''
         query_set = LAMetroEventMedia.objects.filter(event_id=self.ocd_id)
         query_set = sorted(query_set, key=lambda media: media.label, reverse=True)
         return query_set
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -408,7 +408,7 @@ class LAMetroEvent(Event, LiveMediaMixin):
 
     @property
     def media(self):
-        ## order_by makes sure that 'Watch in English' appears first
+        ## makes sure that 'Watch in English' appears first
         query_set = LAMetroEventMedia.objects.filter(event_id=self.ocd_id)
         query_set = sorted(query_set, key=lambda media: media.label, reverse=True)
         return query_set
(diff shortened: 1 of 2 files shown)


lametro/templates/partials/past_event_day.html
```diff
@@ -35,11 +35,11 @@ <h5 class="text-default">
     </div>
 
     <div class="col-xs-3">
-        {% for media in event.media %}
-        <p>
+        {% for media in event.media|dictsortreversed:'label' %}
```
hancush: can you add [a django template comment](https://docs.djangoproject.com/en/2.1/topics/templates/#comments) here, explaining this sort?

jmithani:
i'm having some issues with this on the model level.

in theory, this should work:`LAMetroEventMedia.objects.filter(event_id=self.ocd_id).order_by('-label')` but i'm getting an error that `label` isn't an option. 

going into the shell, `LAMetroEvent.objects.filter(media__isnull=False)` gives me the error that `media` does not exist, though `dir(LAMetroEvent.objects.first())` shows `media` as a property and calling `LAMetroEvent.objects.first().media` returns a value (of `[]`).

let's look at it together tomorrow! writing this so i can remember my steps, heh

hancush:
@jmithani both [`label`](https://github.com/datamade/la-metro-councilmatic/blob/0c66a41293d99ad33cf7d4a81a2a86e3ba8f50d9/lametro/models.py#L419) and [`media`](https://github.com/datamade/la-metro-councilmatic/blob/0c66a41293d99ad33cf7d4a81a2a86e3ba8f50d9/lametro/models.py#L410) are _properties_ of their respective models, not _attributes_.

attributes are stored in the database, as columns in the model tables. properties are (often) calculated from database attributes, but they exist only in python-land. so, you cannot use properties to order a query set, because it uses the database layer, where properties aren't available.

if you want to sort by the label property, you'll have to do that in python. does that make sense?

lametro/models.py
```diff
@@ -408,7 +408,10 @@ def upcoming_committee_meetings(cls):
 
     @property
     def media(self):
-        return LAMetroEventMedia.objects.filter(event_id=self.ocd_id)
+        ## order_by makes sure that 'Watch in English' appears first
```
hancush: this comment would function better as a doc string, like: "Return related LAMetroEventMedia objects such that English audio appears first"

PR #390, opened 2019-01-16
Displaying Related Board Reports by date

### LINKED ISSUE #355: Report: Related Board Reports are out of order
Related Board Reports should be listed in descending order by date of last action.
image

jmithani:
Orders the `related_bills` in `views.py` so they are displayed in descending date order.

Closes #355.

### DIFF
--- a/lametro/views.py
+++ b/lametro/views.py
@@ -83,7 +83,7 @@ class LABillDetail(BillDetailView):
         # Create list of related board reports, ordered by descending last_action_date.
         if context['legislation'].related_bills.all():
             all_related_bills = context['legislation'].related_bills.all().values('related_bill_identifier')
-            related_bills = [ bill for bill in LAMetroBill.objects.filter(identifier__in=all_related_bills) ]
+            related_bills = LAMetroBill.objects.filter(identifier__in=all_related_bills)
             context['related_bills'] = sorted(related_bills, key=lambda bill: bill.last_action_date, reverse=True)
 
 
--- a/lametro/views.py
+++ b/lametro/views.py
@@ -4,7 +4,7 @@ from operator import attrgetter
 import itertools
 import urllib
 import json
-from datetime import date, timedelta, datetime
+from datetime import date, timedelta, datetime, MINYEAR
 from dateutil.relativedelta import relativedelta
 from dateutil import parser
 import requests
--- a/lametro/views.py
+++ b/lametro/views.py
@@ -81,10 +81,12 @@ class LABillDetail(BillDetailView):
             context['packet_url'] = None
 
         # Create list of related board reports, ordered by descending last_action_date.
+        # Thanks https://stackoverflow.com/a/2179053 for how to handle null last_action_date
         if context['legislation'].related_bills.all():
             all_related_bills = context['legislation'].related_bills.all().values('related_bill_identifier')
             related_bills = LAMetroBill.objects.filter(identifier__in=all_related_bills)
-            context['related_bills'] = sorted(related_bills, key=lambda bill: bill.last_action_date, reverse=True)
+            minimum_date = datetime.date(datetime.MINYEAR, 1, 1)
+            context['related_bills'] = sorted(related_bills, key=lambda bill: bill.last_action_date or minimum_date, reverse=True)
 
 
         return context
--- a/lametro/views.py
+++ b/lametro/views.py
@@ -85,7 +85,7 @@ class LABillDetail(BillDetailView):
         if context['legislation'].related_bills.all():
             all_related_bills = context['legislation'].related_bills.all().values('related_bill_identifier')
             related_bills = LAMetroBill.objects.filter(identifier__in=all_related_bills)
-            minimum_date = datetime.date(datetime.MINYEAR, 1, 1)
+            minimum_date = datetime(MINYEAR, 1, 1, 0, 0, 0, 0, app_timezone)
             context['related_bills'] = sorted(related_bills, key=lambda bill: bill.last_action_date or minimum_date, reverse=True)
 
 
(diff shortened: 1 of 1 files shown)


lametro/views.py
```diff
@@ -80,9 +80,11 @@ def get_context_data(self, **kwargs):
         except:
             context['packet_url'] = None
 
-        # Create list of related board reports
+        # Create list of related board reports, ordered by descending last_action_date.
         if context['legislation'].related_bills.all():
-            context['related_bills'] = [LAMetroBill.objects.get(identifier=bill.related_bill_identifier) for bill in context['legislation'].related_bills.all()]
+            related_bills = [LAMetroBill.objects.get(identifier=bill.related_bill_identifier) for bill in context['legislation'].related_bills.all()]
```
hancush: while we're in here, let's make this a bit more efficient. this approach will make a database call for each related bill. how about we make use of [the `in` filter](https://docs.djangoproject.com/en/2.1/ref/models/querysets/#in) to grab all the `LAMetroBill` objects where the identifier is in the related bills?

lametro/views.py
```diff
@@ -80,9 +80,11 @@ def get_context_data(self, **kwargs):
...
+            context['related_bills'] = sorted(related_bills, key=lambda bill: bill.last_action_date, reverse=True)
```
hancush: nice work. ;-) the sorted method is a great approach here, and for the audio links as well, because we know the thing we're sorting is going to be relatively small. a word to the wise (i learned this the hard way recently): if you're working with a larger thing, sorting can be suuuper inefficient.

lametro/views.py
```diff
@@ -80,9 +80,12 @@ def get_context_data(self, **kwargs):
         except:
             context['packet_url'] = None
 
-        # Create list of related board reports
+        # Create list of related board reports, ordered by descending last_action_date.
         if context['legislation'].related_bills.all():
-            context['related_bills'] = [LAMetroBill.objects.get(identifier=bill.related_bill_identifier) for bill in context['legislation'].related_bills.all()]
+            all_related_bills = context['legislation'].related_bills.all().values('related_bill_identifier')
+            related_bills = LAMetroBill.objects.filter(identifier__in=all_related_bills)
+            context['related_bills'] = sorted(related_bills, key=lambda bill: bill.last_action_date, reverse=True)
```
hancush: this will work great, provided that all bills have a `last_action_date`. is it possible to have no last action date?

jmithani:
it can be null, so a different field might be better. would
`ocd_created_at` (or is that just when it was put in?) or `updated_at` be
better? i got `last_date_updated` from here
<https://github.com/datamade/la-metro-councilmatic/blob/master/lametro/templates/lametro/legislation.html#L164>
.

lametro/views.py
jmithani: what have you implemented instead in those cases?

hancush:
@jmithani ocd_created_at is the date the item was scraped into the ocd api. updated_at is the last time it was updated. last_action_date is the right field to sort on, you'll just need to accomodate the nones in your sort. [this](https://stackoverflow.com/questions/2178931/in-python-sorting-on-date-field-field-may-sometimes-be-null) might get you started!

lametro/views.py
```diff
@@ -81,10 +81,12 @@ def get_context_data(self, **kwargs):
             context['packet_url'] = None
 
         # Create list of related board reports, ordered by descending last_action_date.
+        # Thanks https://stackoverflow.com/a/2179053 for how to handle null last_action_date
         if context['legislation'].related_bills.all():
             all_related_bills = context['legislation'].related_bills.all().values('related_bill_identifier')
             related_bills = LAMetroBill.objects.filter(identifier__in=all_related_bills)
-            context['related_bills'] = sorted(related_bills, key=lambda bill: bill.last_action_date, reverse=True)
+            minimum_date = datetime.date(datetime.MINYEAR, 1, 1)
```
hancush: you imported `MINYEAR` from `datetime`, so you can just access it as `MINYEAR` here.

jmithani: `minimum_date = datetime.date(MINYEAR, 1, 1)` like this?
that gets me `descriptor 'date' requires a 'datetime.datetime' object but received a 'int'`

jmithani: although, i **think** this would still return an error with the comparison, as trying to compare datetime and date objects doesn't work. 

switching to a datetime got me an error about naive and aware objects, which i'm still parsing.

lametro/views.py
```diff
@@ -81,10 +81,12 @@ def get_context_data(self, **kwargs):
...
+            context['related_bills'] = sorted(related_bills, key=lambda bill: bill.last_action_date or minimum_date, reverse=True)
```
hancush: `can't compare offset-naive and offset-aware datetimes` happens (as you might expect) when you try to compare datetime objects with a timezone and without one. the datetimes in the metro db all have the los angeles timezone associated. so, you need to [add a timezone](https://docs.python.org/3/library/datetime.html#datetime-objects) to your minimum date.

n.b., `app_timezone` is [defined in models](https://github.com/datamade/la-metro-councilmatic/blob/730abffbc065c17d43a372b671d8c3c2bdce5b13/lametro/models.py#L19) and [imported in this module](https://github.com/datamade/la-metro-councilmatic/blob/730abffbc065c17d43a372b671d8c3c2bdce5b13/lametro/views.py#L36) with the `*`, so it's available for you!

lametro/views.py
hancush: you want the datetime object, as you've already arrived at! see https://github.com/datamade/la-metro-councilmatic/pull/390/files#r248859862 for the comparison problemo.

PR #391, opened 2019-01-19
wiki 

reginafcompton:
This PR contains documentation for the la-metro-councilmatic wiki. I do not plan to merge the wiki repo into master (though I will need to merge the images), but rather copy it over to the repo's wiki. PR's are just nice places to request feedback.

***
I would like to make our original Google doc obsolete (i.e., delete it entirely, or focus it to just the after-hours policy). I need to decide which other parts of that Google doc should be moved to this Wiki. Maybe:
* something about "useful commands"
* something that outlines the after-hours policy (though maybe we should point to an internal document, rather than making that info available open source)

https://docs.google.com/document/d/1X34EU06gErdZ1psg8niSqDMk0nH-M1vsP-RNPLNUsF4

@hancush would you have time to give this a first pass? It's still missing lots of details, but I'd love to talk about it from a high-level perspective, before going any further.


wiki/debugging.md
```diff
@@ -0,0 +1,76 @@
+# Debugging data issues
+
+Many issues can arise in the Metro galaxy, from the shallowest part of the frontend to the deepest end of the backend. Tackling data inaccuracies requires a strategic approach. This guide traces an easeful, step-by-step process for efficient debugging. 
+
+DON'T PANIC.
```
hancush: ha! i would put this at the TOP. ;-)

wiki/debugging.md
```diff
@@ -0,0 +1,76 @@
...
+
+DON'T PANIC. 
+
+## Identifying the point of failure
+
+[point to the charts in the overview that describe the data pipeline]
+
+Data travels along a complicated path to arrive in Metro Councilmatic. A disruption in the data pipeline can happen at four different points. 
+
+1. Legistar - the Legistar API or web interface did not provide accurate data
```
hancush: what about a point 0: data entry by metro (or other legislative body) staff?

wiki/debugging.md
```diff
@@ -0,0 +1,76 @@
...
+DON'T PANIC. 
+
+## Identifying the point of failure
+
+[point to the charts in the overview that describe the data pipeline]
+
+Data travels along a complicated path to arrive in Metro Councilmatic. A disruption in the data pipeline can happen at four different points. 
+
+1. Legistar - the Legistar API or web interface did not provide accurate data 
+2. OpenCivicData API - the DataMade scrapers did not properly port data to the API
```
hancush: is it worth mentioning that this is also a database, underneath it all?

hancush: should we intro pupa + python-legistar here at all?

wiki/debugging.md
```diff
@@ -0,0 +1,76 @@
...
+tail -100 /var/log/councilmatic/lametro-importdata.log
+```  
+
+As with the scraper, if the import seems to be running without error, then the bug may be harder to find. A good first question: what are the most recent changes in the `django-councilmatic` codebase (the import script, but also the data modeling)? could this have caused unusual behavior?
+
+### Step 6. Investigate the display logic.
+
+You successfully confirmed that the data landed, without failure, in the Metro Councilmatic database! The bug resides somewhere in `la-metro-councilmatic` or the parent app, `django-councilmatic`. You can work through this by, first, examining the output of the view code, and then, inspecting how the templates deal with the data. Again, be sure to ask: what are the most recent changes, and could they have affected the health of the code?
+
+## Recurring problems
```
hancush: it'd be good to list these known issues up high, even if it's just some nav items with links to the explanations down here.

wiki/debugging.md
reginafcompton: that seems directly linked to Legistar. but I think I can add a clause, like, "likely because Metro staff did not enter accurate data"

wiki/debugging.md
reginafcompton: Let's mention both these things at the bottom of "Step 4" (below). These opening four points should not overwhelm the reader with too much detail, I think.

wiki/debugging.md
```diff
@@ -1,16 +1,18 @@
...
+If the bug-in-question feels like a familiar one, then read below about [recurring problems](#recurring-problems).
 
 ## Identifying the point of failure
 
 [point to the charts in the overview that describe the data pipeline]
 
 Data travels along a complicated path to arrive in Metro Councilmatic. A disruption in the data pipeline can happen at four different points. 
 
-1. Legistar - the Legistar API or web interface did not provide accurate data 
+1. Legistar - the Legistar API or web interface displays inaccuracies, likely because Metro staff did not enter correct data
```
hancush: good clarification! this suggests that the solution likely involves human error, which i'd approach differently than if it were a website malfunction.

reginafcompton:
After the last round of revision, I have a clear(er) picture of what's left to accomplish:

- [ ] add more examples to the "Debugging" section
- [ ] write the content for the "Commands to know" page
- [ ] write content for the "Deployment" page
- [ ] determine where to include info about After Hours Support (see note above)

I'll let @hancush add language and images to the Overview page. Then, we can port this to the wiki.

reginafcompton:
@hancush - this document is ready for further review when you have time! In particular, I'd appreciate a close reading of the Debugging and Commands pages.

wiki/commands.md
```diff
@@ -0,0 +1,134 @@
...
+```
+
+```bash
+# scrape all people
+# the people scraper does not have a "window" argument
+# but instead determines which people to update by looking at those visible on the web interface
+pupa update lametro people
+```
+
+## Importing data to Councilmatic
```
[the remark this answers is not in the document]
reginafcompton: Hmmmm.....I sometimes do, but most often do not. Do you have a preferred practice you'd like to instate?

reginafcompton:
@hancush - made changes: do these seem to resolve your comments? If so, I'll get cracking on the deployment section.

wiki/commands.md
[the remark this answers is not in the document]
hancush: hm. i do the same as you. but, i think it would be nice to have a reliable record of when the imports are run, whether it's via cron or manually. shall we default to logging?

wiki/commands.md
```diff
@@ -0,0 +1,138 @@
+# Commands to know
+
+The LA metro galaxy comes with several CLI commands and their various options. This section identifies some of the most significant commands, how to use them, and where to execute them. 
+
+## Scraping data from Legistar
+
+Running the scrapers can be simple or fairly involved. You can run full scrapes or "windowed" scrapes; you can run scrapes at faster or slower rates; you can run scrapes for all data or just bills, events, or people (oh, my). 
+
+**Note!** The Metro scrapers on the server run at different intervals, depending on the day. [The crontask file includes the details](https://github.com/datamade/scrapers-us-municipal/blob/master/scripts/scrapers-us-municipal-crontask#L12). (Learn how to decipher the specifics of crontasks with [the beloved crontab guru](https://crontab.guru/). Roughly, the scrapers:
```
hancush: think the closing parenthesis is missing on the crontab sentence!

wiki/commands.md
```diff
@@ -0,0 +1,138 @@
...
+
+Then, run the appropriate command.
+
+```bash
+# scrape all recently updated data
+pupa update lametro
+
+# scrape all recently updated data, but used cached pages
+# avoid fastmode for larger scrapes, since you risk scraping out-of-date data
+# instead, use th `rpm` option (see below)
```
hancush: the*

wiki/commands.md
```diff
@@ -0,0 +1,153 @@
...
+# shell into the server
+ssh ubuntu@lametro.datamade.us
+
+# visit the production directory, and launch the correct virtual environment
+sudo su - datamade
+cd lametro
+workon lametro
+
+# run the command!
+python manage.py import_data
```
hancush: i think it might be worth adding the bit to append the output to our log files for each of these commands, in case the reader doesn't make it all the way to the bottom. would you mind moving the logging section up, and adding `>> /the/appropriate/logfile.txt` to each of the import pipeline commands in this section, for easy copy/paste-ability?

wiki/debugging.md
```diff
@@ -0,0 +1,98 @@
...
+https://ocd.datamade.us/ocd-event/4b2ebe06-6e17-4602-846a-8bffd562963b
+```
+
+### Step 3. Find the source URLs.
+
+The OCD API provides links to the Legistar API and web interface. For example, the above mentioned event has source urls listed like so: 
+
+![Source URLs](https://github.com/datamade/la-metro-councilmatic/tree/master/lametro/static/images/wiki/source_url.png)
+
+Events scrape the Spanish language source (i.e., "api (spa)"), as well. In initial debugging, focus on the "api" and "web" sources – by visiting these links and checking that the data in Legistar appears as expected. Does it?
```
hancush: *sap

wiki/deployment.md
```diff
@@ -0,0 +1,23 @@
+# Deployment
+
+DataMade hosts a [staging version](https://lametro.datamade.us/) and [production version](https://boardagendas.metro.net/) of Metro Councilmatic. 
+
+Both sites reside on the [`metro-councilmatic` server](https://github.com/datamade/server-metro-councilmatic), an EC2 instance spun up in late 2018, after [the original Councilmatic server ran out of diskspace](https://github.com/datamade/devops/issues/55). Metro developers should be able to shell into this server, e.g., [for executing management commands](commands.md). If you cannot, then talk to your friendly DataMade devops team.
+
+### Patterns and practices
+
+Metro deployments follow a typical DataMade pattern: merging or pushing to master deploys to the staging site, and pushing a tag deploys to the production site. Metro Councilmatic also uses the "classic" style of DataMade deployment - or, conversely, Metro does not follow the contempary ["zero downtime" deployment pattern](https://github.com/datamade/deploy-a-site/blob/master/Zero-Downtime-Deployment.md). As a result, the Metro site raises a 500 error for a couple minutes, during the course of a codedeploy lifetime.
```
hancush: can we phrase this like, "_triggering_ metro deployments follows the typical datamade pattern... conversely, the metro deployment scripts do not use datamade's contemporary, zero-downtime pattern." helps to know which part of the deployment we're talking about, i think.

wiki/deployment.md
```diff
@@ -0,0 +1,23 @@
...
+So, when can you trigger a deployment to production? DataMade *does not* need to communicate to the folks at Metro the exact timing of deployments. However, we do take a couple precautions:
+
+1. Do not deploy during meetings, since we do not want the site to go down for a couple minutes, when users need access to event info.
+2. Do not deploy on support Fridays, since we do not want the site behave unexpectedly, when Metro admin adds data on Friday night.
+
+These practices have exceptions, of course, particularly, if Metro needs an urgent bug fix.
+
+### Solr
+
+Deployments rarely require special intervention or deviation from the standard CodeDeploy pipeline. Only one aspect of Metro Councilmatic requires attention upon deployment: regenerating the Solr schema.
```
hancush: although i think the readme has a line about it, can we say here that this is not necessary unless we have made a change to the schema, which we don't do frequently?

wiki/deployment.md
reginafcompton: nice - I like this

reginafcompton:
@hancush - another batch of edits! 

also, for info about Friday support nights, I think we should keep that as an internal document. It seems odd to outline our protocols for client relations in a publicly accessible wiki. what do you think?

jmithani:
an overview of the (template) file structure would be helpful—i got twisted up between `event.html` and `events.html` and all of the partials.

Issue #328, opened 2018-07-14
Event detail: Related board reports are not showing up


reginafcompton:
The most recent events do not show related board reports (though previous events do).

The SQL in the view code is not behaving as expected: https://github.com/datamade/la-metro-councilmatic/blob/master/lametro/views.py#L146

reginafcompton:
**Initial Problem**
Metro added agendas on July 13, around 2:30 CST. Over two dozen bills (present in Legistar) were not in the `opencivicdata` database at this time. As a result, the "Related Board Reports" did not render on Event Detail pages.

**Preliminary Solution**
I manually ran the scraper (twice), around 5:00, with the `window` option:
```
pupa update lametro bills window=7
pupa update lametro bills window=500
```
These scrapes caught nearly all of the missing bills (which were added to the OCD API database at 5:25 pm or 5:45 pm on July 13). The data import in Councilmatic subsequently added these bills to LA Metro.

Still, the related board reports did not render as expected. 

**Cause and Effect**
I turned off the scraper crontasks during the manual execution of the scraper - hence, turning off the events scraper, which creates the `related_entities` and event items upon which the [Councilmatic view](https://github.com/datamade/la-metro-councilmatic/blob/master/lametro/views.py#L146) depends. After turning on cron again, the event scraper ran, mitigating the issue (like magic).

**Root Problem (Unsolved)**
Why were these board reports missing? Many have a Legistar `MatterLastModifiedUtc` timestamp
of July 6, suggesting that the scraper should have scraped the bill on or before that date. Examples:
http://webapi.legistar.com/v1/metro/matters/5039
http://webapi.legistar.com/v1/metro/matters/5022
http://webapi.legistar.com/v1/metro/matters/5073
http://webapi.legistar.com/v1/metro/matters/5124

The Sentry logs do not show any corresponding errors, and the DataMade team [did not develop the scraper on this day](https://github.com/opencivicdata/scrapers-us-municipal/commits/master).

reginafcompton:
**Reference: List of previously missing bills**

```
[...]

PR #394, opened 2019-02-07
Add conditional for hiding bills with restricted views

reginafcompton:
This PR adds a conditional to the `is_visible` property: it explicitly hides bills marked as having a restricted view in Legistar.

It works in cooperation with https://github.com/opencivicdata/scrapers-us-municipal/pull/258, https://github.com/opencivicdata/python-legistar-scraper/pull/81, and https://github.com/datamade/django-councilmatic/pull/232

N.b., I'll get the tests passing once I can add the above-mentioned changes in Django Councilmatic to the testing requirements.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,2 +1,4 @@
 [pytest]
 DJANGO_SETTINGS_MODULE = tests.test_config
+filterwarnings =
+    ignore::django.utils.deprecation.RemovedInDjango20Warning
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -85,7 +85,7 @@ class LAMetroBill(Bill):
         This issue summarizes why that might happen: https://github.com/datamade/la-metro-councilmatic/issues/345#issuecomment-421184826
         Metro staff devised three checks for knowing when to hide or show a report:
 
-        (1) Is the view restricted, i.e., is `MatterRestrictViewViaWeb` set to True in the Legistar API? We skip these bills further upstream. https://github.com/opencivicdata/scrapers-us-municipal/pull/251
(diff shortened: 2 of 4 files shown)


reginafcompton:
### Overview

With the newly provided API token, DataMade scrapers now have access to private Metro bills. But….why do we need to scrape private bills?

We need to know which bills to show and which bills to hide on Councilmatic: the MatterRestrictedViewViaWeb field tells us which bills are public (show!) and which are private (hide!). The process: 
1. scrape bill and include the value of MatterRestrictedViewViaWeb in the extras (https://github.com/opencivicdata/scrapers-us-municipal/pull/258 and https://github.com/opencivicdata/python-legistar-scraper/pull/81)
2. import bills and indicate if they are restricted (https://github.com/datamade/django-councilmatic/pull/232)
3. hide or show bills based on this value. (this PR)

But wait. Why not just skip private bills? Sometimes, Metro will toggle bills between public and private, meaning that the Councilmatic system may have bills that were previously public. Scraper needs to access those now-private bills, so that Councilmatic can make sense of their new status.

Not knowing which bills had changed between public and private has caused issues in the past:
https://github.com/datamade/la-metro-councilmatic/issues/345#issuecomment-455683240
https://github.com/datamade/la-metro-councilmatic/issues/328

pytest.ini
```diff
@@ -1,2 +1,4 @@
 [pytest]
-DJANGO_SETTINGS_MODULE = tests.test_config
\ No newline at end of file
+DJANGO_SETTINGS_MODULE = tests.test_config
+filterwarnings =
+    ignore::django.utils.deprecation.RemovedInDjango20Warning
```
hancush: :)

Issue #379, opened 2018-11-15
Agendas: Text instructions for opening links


shrayshray:
Links in embedded Agenda PDFs only work when the CTRL key is held while clicking the link. Please add a line of text noting this above the PDF. (It should only be viewable when accessed via a desktop browser, as the embedded PDF is hidden in mobile view.) See mockup below.
![pdfinstructionsmockup](https://user-images.githubusercontent.com/28063988/48586486-f2e3c100-e8e4-11e8-92af-539e4122218d.png)

shrayshray:
This issue is specifically about the hyperlinks in Agendas on File ID's of Reports and Report attachments: 
![image](https://user-images.githubusercontent.com/28063988/52675372-84120600-2edb-11e9-9c87-6c976f1afee0.png)

reginafcompton:
Closed via #402. Thanks @jmithani !

PR #402, opened 2019-02-15
Adding line about how to access links on pdfs

jmithani:
Addresses #379 

Adds in helper text that is shown when the PDF viewer iframe is on the screen.


jmithani:
@reginafcompton done!

Issue #350, opened 2018-09-21
Overlapping text in mobile view


shrayshray:
Text on the homepage is overlapping in mobile browsers. I think this has to do with the addition of the attachment search buttons! Here's a screenshot from Omar's phone (Android). I'm having the same issue on iOS/Safari.
![screenshot_20180921-141942](https://user-images.githubusercontent.com/28063988/45910019-a71c0c00-bdba-11e8-9a8b-ffa0dc77f3ee.png)

jmithani:
Closed via #404

jmithani:
@shrayshray I deployed a fix for this. When you get a chance, can you let us know what you think?

shrayshray:
@jmithani looks great to me, thank you!!

PR #404, opened 2019-02-20
Fixing overlapping text on front page

jmithani:
Addressing #350.

- Media query added to accommodate the size of the form entry and button for xs screens
- Added `hidden-sm` on the breaks so that the text doesn't overlap on tablet sizes either


jmithani:
@reginafcompton ready for another review!

Issue #403, opened 2019-02-15
Events page does not load fast enough


reginafcompton:
Today, we noticed that the Events page on the staging site was raising a 502 error: this was caused by Nginx timing out, suggested by several `client closed connection while waiting for request`errors in the Nginx log (i.e., `/var/log/nginx/lametro-staging-councilmatic-error.log`).

## The particular problem
The database had nearly 300 TEST events, which dramatically slowed the load time. I cleared those, and the site runs fine now. (I did the same on the production database.)

## The more general problem
**The site should be able to handle this volume of events.** (At some point, we'll have hundreds more!) Let's dig into this issue, sooner rather than later. Two considerations:

- how are we building and returning the Events queryset, and can we do that more efficiently? [`prefetch_related`](https://docs.google.com/document/d/1giTexSmZ0hXd1sPBU8AUuT_O2VWgpXaGqDZnMzW05kE/edit) comes to mind. (We used [prefetch_related in Chicago Councilmatic](https://github.com/datamade/chi-councilmatic/blob/master/chicago/views.py#L28) for a similar issue.)

- Solr! Let's finally deal with issue https://github.com/datamade/la-metro-councilmatic/issues/233

@shrayshray I'd like to prioritize this in our February milestone? Would that be okay?

shrayshray:
Yes, please prioritize!

reginafcompton:
## Expensive code

**[`updates_made`](https://github.com/datamade/la-metro-councilmatic/blob/master/lametro/templatetags/lametro_extras.py#L179)** 
Iterates over all EventDocuments for every instance of an Event rendered on the Event template. 

Solution? We can gather this information by annotating the main queryset, rather than gathering this data in the template. That is, do away with the `updates_made` template tag and add that logic to the view code. 

**implementation of [`media` property](https://github.com/datamade/la-metro-councilmatic/blob/master/lametro/models.py#L413)**
Queries for all media objects, sorts them, and [iterates over them for each event.](https://github.com/datamade/la-metro-councilmatic/blob/master/lametro/templates/partials/past_event_day.html#L37-L44)

Solution? Again, it seems like we should annotate the main event queryset with media data, rather than in the template.

reginafcompton:
@shrayshray - we made good progress on this issue. I see a notable difference in load time for `/events`. I think we can finish the remaining items in March (or later). I'll put this in the March milestone for now.

PR #406, opened 2019-02-21
Faster events

reginafcompton:
This PR addresses the first checkbox discussed here: https://github.com/datamade/la-metro-councilmatic/issues/403#issuecomment-465208479

I plan to move through issue #403 iteratively, taking on each checkbox, one at a time.

### DIFF
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -409,14 +409,6 @@ class LAMetroEvent(Event, LiveMediaMixin):
         return meetings
 
 
-    @property
-    def media(self):
-        '''Return related LAMetroEventMedia objects such that English audio appears first'''
-        query_set = LAMetroEventMedia.objects.filter(event_id=self.ocd_id)
-        query_set = sorted(query_set, key=lambda media: media.label, reverse=True)
-        return query_set
-
-
 class LAMetroOrganization(Organization):
     '''
     Overrides use the LAMetroEvent object, rather than the default Event
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -422,6 +422,10 @@ class LAMetroEventMedia(EventMedia):
     class Meta:
         proxy = True
 
+    def __init__(self, *args, **kwargs):
+        super(LAMetroEventMedia, self).__init__(*args, **kwargs)
+        self.event = override_relation(self.event)
+
     @property
     def label(self):
         '''
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -8,6 +8,7 @@ from django.db import models, connection
 from django.db.models.expressions import RawSQL
 from django.utils import timezone
 from django.contrib.auth.models import User
+from django.db.models import Max, Min, Prefetch, Case, When, Value
 
 import requests
 
@@ -206,6 +207,19 @@ class LAMetroEventManager(models.Manager):
             return super().get_queryset().exclude(location_name='TEST')
 
         return super().get_queryset()
+
+    def with_media(self):
+        mediaqueryset = LAMetroEventMedia.objects.annotate(
+            olabel=Case(
+                When(note__endswith='(SAP)', then=Value(0)),
+                output_field=models.CharField(),
+            )
+        ).order_by('-olabel')
+
+        events_with_media = Event.objects\
+                                 .prefetch_related(Prefetch('media_urls', queryset=mediaqueryset))
+
+        return events_with_media
 
 
 class LiveMediaMixin(object):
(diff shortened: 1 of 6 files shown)


lametro/models.py
```diff
@@ -417,23 +417,6 @@ def media(self):
         return query_set
 
 
-class LAMetroEventMedia(EventMedia):
```
reginafcompton: EventMedia queried from `prefetch_related` are not the child class (LAMetroEventMedia). [Apparently, that not yet possible in Django.](85aab34288b5e51fb90031cbaf9616fcfea53ae1) 

So, I moved this to a template filter (for now).

lametro/models.py
```diff
@@ -409,31 +409,6 @@ def upcoming_committee_meetings(cls):
         return meetings
 
 
-    @property
-    def media(self):
```
reginafcompton: Replacing this with a template filter that sorts the media_urls set in the template, without hitting the database, again.

lametro/views.py
```diff
@@ -299,7 +301,9 @@ def get_context_data(self, **kwargs):
 
         # If all meetings
         elif self.request.GET.get('show'):
-            all_events = LAMetroEvent.objects.order_by('-start_time')
+            all_events = LAMetroEvent.objects\
+                                     .order_by('-start_time')\
+                                     .prefetch_related('media_urls')
```
hancush: do we see the same gains if we use the `Prefetch` object to return `LAMetroEventMedia` objects? this works in the shell:

```python
from django.db.models import Prefetch

LAMetroEvent.objects.filter(SOME FILTER).prefetch_related(Prefetch('media_urls', queryset=LAMetroEventMedia.objects.all()))
```

lametro/models.py
reginafcompton: Update! I am doing the ordering of media_urls in the original queryset.

lametro/views.py
reginafcompton: We do get the same gains! Building off of that....do you think it's too abstruse to order the Event Media when prefetching? I implemented how that would work here:

https://github.com/datamade/la-metro-councilmatic/pull/406/files#diff-803b5091212eac625d7bc1c2dd933c21R322

reginafcompton:
@hancush - ready for additional review!

lametro/views.py
```diff
@@ -271,8 +283,24 @@ class LAMetroEventsView(EventsView):
...
+
+        We prefetch EventMedia (i.e., 'media_urls') in the Events querysets: 
+        this makes for a more efficient page load. 
+        We also order the 'media_urls' by label, ensuring that links to SAP audio
+        come after links to English audio. 'mediaqueryset' facilitates 
+        the ordering of prefetched 'media_urls'.
+        '''
         context = super(LAMetroEventsView, self).get_context_data(**kwargs)
 
+        mediaqueryset = LAMetroEventMedia.objects.annotate(
```
hancush: how would you feel about moving this prefetch up into the existing event manager [as an extra method](https://docs.djangoproject.com/en/2.1/topics/db/managers/#adding-extra-manager-methods), e.g. `with_media`? that way we have an easy way to access, without repeating code. 

with that change, you can then get the events with media prefetched, like so:

```python
LAMetroevent.objects.with_media().filter(...)   # etc.
```

as far as i can tell, we still get the gains? (i've been looking at the django debug toolbar – are you monitoring a different way? this was my q from semaphor, btw!)

reginafcompton: sounds excellent! (yeah, I was not pleased about the level of DRY-ness here.) 

And yes, I am using django debug toolbar!

lametro/views.py
```diff
@@ -130,15 +130,19 @@ def post(self, request, *args, **kwargs):
 
     def get_context_data(self, **kwargs):
         context = super(EventDetailView, self).get_context_data(**kwargs)
-        # Create URL for packet download.
+        
+        # Get the event with prefetched media_urls in proper order. 
         event = context['event']
+        event = LAMetroEvent.objects.with_media().filter(ocd_id=event.ocd_id).first()
+        context['event'] = event
```
hancush: this looks like a good case for overriding [the `get_object` method](https://ccbv.co.uk/projects/Django/2.1/django.views.generic.detail/DetailView/#get_object)!

reginafcompton: GOOD IDEA

reginafcompton:
THANK YOU HANNAH! YOU TOO

Issue #400, opened 2019-02-14
Fiscal year facet does not display correctly for the current year


reginafcompton:
<img width="1257" alt="screen shot 2019-02-14 at 3 46 12 pm" src="https://user-images.githubusercontent.com/13078679/52816721-df74fd00-306f-11e9-94ea-7148c51f77d4.png">

reginafcompton:
Closed via #418

PR #418, opened 2019-03-12
Dynamically prepare facet for legislative session

reginafcompton:
This PR addresses issue #400.

I moved the calculations for the legislative session from the template logic to `search_indexes.py`. 

Notably, his allows us to remove a once-custom partial. (N.b., [I used the delightful diff checker](https://www.diffchecker.com/diff) to compare between the old Metro partial and [the Django-Councilmatic partial.](https://github.com/datamade/django-councilmatic/blob/b25523132ba857d3c7f3c5d9886b8076b4ef21a4/councilmatic_core/templates/partials/search_filter_item.html))

I also wrote some tests for the prepare functionality in `search_indexes.py`.


tests/conftest.py
```diff
@@ -43,6 +44,21 @@ def build(self, **kwargs):
 
     return BillFactory()
 
+@pytest.fixture
+@pytest.mark.django_db
+def legislative_session(db):
```
hancush: i'm totally fine with this fixture and don't require any changes! just a question for science: is there a reason you didn't use the factory pattern you've used elsewhere for this particular fixture?

reginafcompton: "for science"...heh.

We use factory-style fixtures whenever we instantiate more than one instance of a model within a single test, [e.g., here](https://github.com/datamade/la-metro-councilmatic/blob/master/tests/test_bills.py#L26). The tests for this PR do not require that of legislation sessions, and until the tests do, then I think we can leave the `legislative_session` fixture as a simple one.

Issue #233, opened 2017-11-10
Events: Remove Solr dependency 


reginafcompton:
Currently, we discover minutes related to a Board Meeting using the Solr search engine: https://github.com/datamade/la-metro-councilmatic/blob/master/lametro/templatetags/lametro_extras.py#L101

Let's find a way to do this without using Solr.

shrayshray:
Per Reginia's email 11/9/2017, it's anticipated this will be addressed by the end of January.

reginafcompton:
@shrayshray - this issue does not directly relate to issue #232 (which we'll address by the end of January). Rather, this issue makes a change to the Solr-dependency on the events page.

reginafcompton:
We have a fix for this! PR #424. @hancush will deploy it to production tomorrow and close the issue!

hancush:
update: i'm out of the office today, will deploy this (and #425) on monday!

hancush:
fixed a bug in the implementation of this issue in #429, insodoing also debugged some mysterious behavior in the import in https://github.com/datamade/django-councilmatic/pull/243. 

tested successfully locally, but the aforementioned mysterious behavior ("complete" data imports actually being incomplete) is precluding testing on staging because agenda items for this year's board meetings have not been imported, so the associated minutes are not associated with the events and are not showing up. this will be fixed next week (and will also close #430).

in the meantime, because we've tested successfully locally and the production database is in good shape, i think it is safe to deploy this to production.

hancush:
aaaannnnd, it's deployed! @shrayshray, do things look in order to you? (that is, do board meetings that should have minutes, have minutes?)

shrayshray:
@hancush looks good to me!

PR #424, opened 2019-04-03
Remove Solr dependency on events page

reginafcompton:
## Overview

This PR revises our strategy for finding minutes associated with Board Meetings on the Events listview. 

Some events do not have corresponding minutes in an EventDocument: previously, we queried Solr to discover those minutes. Thanks to [updates in django-councilmatic](https://github.com/datamade/django-councilmatic/blob/master/councilmatic_core/management/commands/import_data.py#L986), the `ocr_full_text` contains what we need.

This PR also moves the `get_minutes` logic into the LAMetroEvent class, rather than a template tag. 

Altogether, this new approach saves us around 200 queries. 

#### pytest
This PR includes three tests for three simple cases, i.e., when (1) `event_minutes` should return none, (2) `event_minutes` should return an EventDocument url, and (3) `event_minutes` should return a board report url. 

Execute `pytest tests/test_events.py`, and wait for tests to pass.

#### run the code
* `python manage.py runserver`
* visit http://127.0.0.1:8000/events/
* look for the Minutes hyperlink next to Board Meetings

Handles #233, but also relates to #403

### DIFF
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -460,7 +460,7 @@ class LAMetroEvent(Event, LiveMediaMixin):
 
 
     @property
-    def event_minutes(self):
+    def board_event_minutes(self):
         '''
         This method returns the link to an Event's minutes. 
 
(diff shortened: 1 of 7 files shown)


lametro/models.py
```diff
@@ -458,6 +459,31 @@ def upcoming_committee_meetings(cls):
         return meetings
 
 
+    @property
+    def event_minutes(self):
```
hancush: is it true that this only ever returns values for board meetings (and not subcommittee meetings, etc.)? if so, could we:

1. check whether the event is a board meeting up top, and return none if it isn't, to avoid making queries when we don't need to?
2. rename this to `board_meeting_minutes`, to make it a bit more explicit?

hancush: oh, i see we check whether it's a board meeting in the template, so we won't save queries for this use case, but it would be nice to bundle that all into this property, so it's clear if we want to make use of it later.

reginafcompton: Agreed! (And what I originally had: https://github.com/datamade/la-metro-councilmatic/pull/424/commits/6884be4d4371ce28ef28cb3635dd1a189320bf72#diff-154e85485bdcc3826d208aa9dbc8d8e3R471)

PR #427, opened 2019-04-05
Rotate old AWS keys

jeancochrane:
## Overview

LA Metro is still using old AWS credentials attached to evz's account in order to query the property image cache. Rotate these credentials for a new pair that is created for a separate IAM user and that only has read/write access to the two PIC buckets.

### Checklist

## Testing Instructions

 * Swap out your settings for the production settings:

```bash
blackbox_cat configs/settings_deployment.production.py.gpg > councilmatic/settings_deployment.py
```

* Is it safe to actually refresh the PIC cache? If so, run the management command and confirm that it works with the new credentials:

```
python manage.py refresh_pic
```

* If not, edit the `refresh_pic.py` management command in your virtualenv to comment out [the line that deletes keys from the cache](https://github.com/datamade/django-councilmatic/blob/8d64575cf93470d13db99db9f4d42afd3a67cc5c/councilmatic_core/management/commands/refresh_pic.py#L33), and then run the command

Connects https://github.com/datamade/ops/issues/579


hancush:
btw, for the sake of writing this down so i actually remember it, [the `property-image-cache` flow](https://github.com/datamade/property-image-cache/blob/379416015e70783a4b7cb81d5f4b8cc82ec7386b/app.py#L60) is this: look for the requested image in the cache. if it doesn't exist, add it. then return the cached image. (naturally, there's some exception handling in there, as well.)

`refresh_pic` removes cached data for things that may have changed. see regina's [excellent explanation](https://github.com/datamade/django-councilmatic/blob/8d64575cf93470d13db99db9f4d42afd3a67cc5c/councilmatic_core/management/commands/refresh_pic.py#L39-L53) of how we choose what to remove. 

so, it's "safe" to run the command in that it won't break anything. but i commented out the delete line anyway, because deleting things unnecessarily can add a second or two of load time to the next viewer of any affected pages (while the requested documents are added to the cache).

Issue #436, opened 2019-05-02
Capture Subject's GUID in LA Metro data model


hancush:
**Warning:** Thinking "out loud" ahead.

Each bill subject has an associated GUID that should be captured in the Councilmatic database, in order to facilitate lookups between the Councilmatic and Semaphore systems. (Subject labels are not guaranteed to be consistent.)

That means we need to incorporate the GUID into the data model. There are a couple of options for doing this:

1. [Multi-table inheritance](https://docs.djangoproject.com/en/2.2/topics/db/models/#multi-table-inheritance). In other words, create a Metro-specific `Subject` subclass with an associated `guid`. 
2. Create a lookup table with a model like `SubjectGuid`, which has:
    - A foreign key to a `Subject` and a character field with the associated GUID.
    - ✅ Character fields for both the subject name and GUID, linkable to the core `Subject` by name.

Note that the `import_data` script deletes and re-imports subjects each time a bill is imported, in order to capture the removal of subjects, so both the multi-table inheritance and the lookup table with a foreign key would require changes upstream, or subclassing the `import_data` and removing implicated `LAMetroSubject` or `SubjectGuid` objects, so deletions can be made from the core `Subject` table, without raising an exception about orphaned references.

However, a lookup table containing the distinct subject and GUID pairings would not require additional changes and, because new topics from SmartLogic are only ported to Legistar once per day, it would only need to be updated once per day, to capture any changes and to remove any unused pairs. We could even take a blunt approach and delete and reimport all pairings from the matter indexes list. This approach would have the added benefit of removing the need to orchestrate this process with a scrape/data import. It also would be an obvious candidate for a callback to be run when SmartLogic has completed its sync with Legistar.

hancush:
Related to #432.

hancush:
Pre-requisite of #437.

hancush:
N.b., the actual task here is: Write a Django model to function as a lookup table between distinct subject-GUID pairs. This model should have fields for the subject name and GUID. 

It would also be good to think about constraints that would help us ensure the data looks the way we expect. For example, subject names may be unique. (Relatedly, we might also think about constraints for the core `Subject`, too, like all subject names appear in the lookup table. This can happen a separate issue.)

jmithani:
Closed by #438

PR #438, opened 2019-05-03
Adding functionaility to refresh Subject-GUID relationship

jmithani:
## Overview

This PR:

- Adds in a new model `SubjectGuid` to act as a lookup model between `Subject` names and Semaphore GUIDs.
- Creates a `refresh_guid` management command that clears out all unused Subject/GUID pairs, then repopulates with current topic list.
- Creates an endpoint to run the `refresh_guid` command.

### Checklist

## Testing Instructions

 * Migrate the database.
 * Navigate to the `http://127.0.0.1:8000/refresh-guid/` endpoint. Confirm it is a 403 response.
 * Copy the `REFRESH_KEY` from `settings_deployment.py.example` into your local `settings_deployment.py`.
 * Access the endpoint now, passing in the `REFRESH_KEY` as a parameter. Confirm you receive a 200 response.

- In the terminal, run `python manage.py refresh_guid`
- Confirm items are deleted and updated via a quick exploration in the django shell.

Handles #436


lametro/views.py
```diff
@@ -797,3 +799,16 @@ def metro_login(request):
...
+@csrf_exempt
+def refresh_guid_trigger(request, refresh_key):
+    try:
+        if refresh_key == settings.REFRESH_KEY:
+            management.call_command('refresh_guid')
+            return HttpResponse(200)
+    except AttributeError:
+        if not refresh_key:
+            print('You need a refresh_key in your local deployment settings files to access this.')
+        else:
```
hancush: good instinct, but it would be more appropriate to pair this `else` with the `if` statement checking whether the given key is correct.

councilmatic/settings_deployment.py.example
```diff
@@ -43,6 +43,9 @@ CACHES = {
 # Set this to flush the cache at /flush-cache/{FLUSH_KEY}
 FLUSH_KEY = 'super secret junk'
 
+# Set this to allow updating of /refresh-guid/{REFRESH_KEY}
```
hancush: ah, i forgot: you'll want to add `REFRESH__KEY` to the encrypted settings files for the staging and production site!

hancush:
@jmithani since it's support night, let's merge this on monday! that way, we don't have changes hanging out on staging that we might not want in production yet, in the event that we need to deploy a fix this evening. this is probably an over-abundance of caution.

jmithani:
roger that! will wait until after the weekend.

Issue #433, opened 2019-04-15
Add subject autocomplete to the search bar


hancush:
### Behavior

1. When a user enters a search term, auto-suggest topics based search bar input
2. If the user issues a plain text query, e.g., types "bus" and hits enter/clicks the magnifying glass, search functionality is unchanged
3. If the user selects a suggested topic, return results matching the selected subject facet, rather than the results of a free text search

### Implementation

- [x] Add a JavaScript event listener to the search bar that POSTs queries as they're typed to SmartLogic's `hints` endpoint, and populates a dropdown of autocomplete suggestions from the response. Each option should display the label, and have a value equal to the topic's GUID. 
- [x] Write an adapter that accepts a GUID and returns the corresponding label in the Councilmatic database. _N.b., using this adapter, rather than querying with the label from the response, hedges against unexpected behavior in the event that a label has changed in Semaphore since the last sync with Legistar/Councilmatic._
- [x] When a user selects an option, feed the GUID to the adapter to construct and issue a search on the subject facet.

jmithani:
Adapter implementation plan:

1. Create `/topic/` endpoint to query with a GUID.
2. From given GUID, `SubjectGuid.objects.get(guid=<param>)`.
3. Return the selected `SubjectGuid` name.

hancush:
@jmithani sometimes, there will be duplicate guids in the lookup table. if that is the case, you will want to grab the name of the subject that also appears in the main `Subject` model. can you update point two to reflect this?

hancush:
n.b., haystack defaults to returning no results if a search is conducted without a query term. to enable topic search, we will want to override this to return all results: https://github.com/django-haystack/django-haystack/blob/802b0f6f4b3b99314453261876a32bac2bbec94f/haystack/forms.py#L41

PR #445, opened 2019-05-20
Add autocomplete search with SmartLogic's Semantic Enhancement Search API

jmithani:
## Overview

This PR:

- Bundles [Devbridge's jQuery autocomplete library](https://github.com/devbridge/jQuery-Autocomplete).
- Implements a Django view to retrieve a JSON web tokens for SmartLogic's Semantic Enhancement Search API.
- Implements autocomplete functionality drawing suggestions from the SES API on the home page and search results page.

### Checklist

## Testing Instructions

- If needed, regenerate the Solr Schema
- Update your `settings_deployment` with the SmartLogic API keys
- In one terminal window `docker-compose up solr-staging`, in another run the server
- Begin typing a topic, e.g. Metro Green Line
- Confirm autocomplete appears and is functional
- Choose a topic, confirm it goes to the appropriate search page when selected

Handles #433


lametro/templates/lametro/index.html
```diff
@@ -136,5 +136,34 @@ <h2>
...
+          onSelect: function(suggestion) {
+            $.when(
+              $.get('/topic/', { 'guid': suggestion.data }))
+                .then(function(response) {
+                  if (response.status_code == 200) {
+                    var base = '/search/?selected_facets=topics_exact%3A';
+                    var url = base + response.subject_safe;
+                    window.location.href = url;
+                  } else {
+                    console.log(response)
```
jmithani: @hancush what should the exception behavior be here you think?

hancush: we don't have great patterns for logging in a useful way in javascript. we have sentry configured to capture exceptions in our routes, so i think don't do anything if we don't have a 200 status code, for now.

lametro/views.py
```diff
@@ -811,3 +815,102 @@ def refresh_guid_trigger(request, refresh_key):
...
+            print('Could not authenticate with SmartLogic: {}'.format(e))
+            return None
+        except Exception:
+            raise
+        else:
+            data = json.loads(response.content.decode('utf-8'))
+            return data['access_token']
+
+
+def fetch_topic(request):
```
hancush: this strikes me as an excellent opportunity to test. can you write tests for each of these cases?

1. there is a single subject with the given guid
2. there are multiple subjects with the given guid
3. there are no subjects with the given guid

let me know if you have any questions that need to be answered before you get started.

lametro/views.py
```diff
@@ -811,3 +815,102 @@ def refresh_guid_trigger(request, refresh_key):
...
+    response['guid'] = guid
+
+    try:
+        subject = SubjectGuid.objects.get(guid=guid)
+        subject = subject.name
+        response['subject_safe'] = urllib.parse.quote(subject)
+        response['status_code'] = 200
+    except MultipleObjectsReturned:
+        subjects = [s.subject for s in SubjectGuid.objects.all()]
+        for s in subjects:
```
hancush: only one of the subject names for a duplicate guid should appear in the subjects table, so you can do this more concisely like `Subject.objects.get(subject__in=[list of names for given guid]`).

lametro/views.py
```diff
@@ -811,3 +815,102 @@ def refresh_guid_trigger(request, refresh_key):
...
+    response = {}
+    response['guid'] = guid
+
+    try:
+        subject = SubjectGuid.objects.get(guid=guid)
+        subject = subject.name
+        response['subject_safe'] = urllib.parse.quote(subject)
+        response['status_code'] = 200
+    except MultipleObjectsReturned:
+        subjects = [s.subject for s in SubjectGuid.objects.all()]
```
hancush: why are you iterating over all `subjectguid` objects?

hancush:
```suggestion
        subjects = [s.name for s in SubjectGuid.objects.all()]
```

lametro/templates/lametro/index.html
```diff
@@ -136,5 +136,30 @@ <h2>
             collapseActivity();
             return false;
         });
+
+        $('#search-bar').autocomplete({
+          serviceUrl: '{% url "autocomplete" %}',
+          paramName: 'query',
+          transformResult: function(response) {
+            var response = JSON.parse(response);
+            console.log(response);
```
hancush:
```suggestion
```

lametro/views.py
jmithani: okay, good to know!

lametro/views.py
jmithani: a mistake!

jmithani: a mistake!

tests/conftest.py
```diff
@@ -185,3 +185,47 @@ def build(self, **kwargs):
...
+@pytest.fixture
+@pytest.mark.django_db
+def subject_guid(db, subject):
+    class SubjectGuidFactory():
+        def build(self, **kwargs):
+            related_subject = subject.build()
+            guid = '0000-0-0000'
+
+            subject_guid_info = {
+                'id': randrange(10000),
```
hancush: just popping in: django will set the id for you, so you don't need to set one here. 👌

lametro/views.py
```diff
@@ -811,3 +815,100 @@ def refresh_guid_trigger(request, refresh_key):
...
+    except MultipleObjectsReturned:
+        subjects = [s.name for s in SubjectGuid.objects.filter(guid=guid)]
+        subject = Subject.objects.get(subject__in=subjects)
+        subject = subject.subject
+        response['subject_safe'] = urllib.parse.quote(subject)
+        response['status_code'] = 200
+    except ObjectDoesNotExist:
+        subject = ''
+        subject_safe = ''
+        response['status_code'] = 404
```
hancush: adding the status code to this response is a great touch. enables a simple, straightforward check when we use this response in the template.

tests/test_fetch_topic.py
```diff
@@ -0,0 +1,49 @@
...
+
+    '''
+    This tests returning a single Subject matching the GUID
+    '''
+
+    s_g = subject_guid.build()
+
+    response = client.get('/topic/', { 'guid': s_g.guid })
+    response = json.loads(response.content.decode('utf-8'))
+    assert response['status_code'] == 200
```
hancush: how about also checking that the subject that came back is the subject from your subject guid?

tests/test_fetch_topic.py
```diff
@@ -0,0 +1,49 @@
...
+    canonical_subject = subject.build()
+
+    canonical_subject_guid_info = { 'name': canonical_subject.subject }
+    old_subject_guid_info = {'name': 'Old Subject'}
+
+    canonical_subject_guid = subject_guid.build(**canonical_subject_guid_info)
+    old_subject_guid = subject_guid.build(**old_subject_guid_info)
+
+    response = client.get('/topic/', { 'guid': canonical_subject_guid.guid })
+    assert response.status_code == 200
```
hancush: ditto checking the subject here.

lametro/views.py
```diff
@@ -811,3 +815,100 @@ def refresh_guid_trigger(request, refresh_key):
...
+    response['guid'] = guid
+
+    try:
+        subject_guid = SubjectGuid.objects.get(guid=guid)
+        subject = subject_guid.name
+        response['subject_safe'] = urllib.parse.quote(subject)
+        response['status_code'] = 200
+    except MultipleObjectsReturned:
+        subjects = [s.name for s in SubjectGuid.objects.filter(guid=guid)]
+        subject = Subject.objects.get(subject__in=subjects)
```
hancush: can you add a comment here for future us that we wrote this with the assumption that there should only ever be one of the two subjects associated with a duplicate subjectguid in use in the main subject table? if this is not the case, it will be nice to have that reminder. :-)

lametro/templates/lametro/index.html
```diff
@@ -136,5 +136,29 @@ <h2>
...
+              suggestions: $.map(response.termHints, function(d) {
+                return {'value': d.name, 'data': d.id};
+              })
+            }
+          },
+          onSelect: function(suggestion) {
+            $.when(
+              $.get('/topic/', { 'guid': suggestion.data }))
+                .then(function(response) {
+                  if (response.status_code == 200) {
```
hancush: this is such an improvement over the jquery ui autocomplete! great find. can you add this autocomplete functionality to the search bar on the search page, too?

lametro/static/css/city_custom.css
```diff
@@ -128,6 +128,32 @@ div#google_translate_element {
...
+  padding: 2px 5px;
+  white-space: nowrap;
+  overflow: hidden;
+}
+
+.autocomplete-selected { background: #F0F0F0; }
+
+.autocomplete-suggestions strong {
+  font-weight: normal;
+  color: #3399FF;
```
hancush: the default blue is a little obtrusive – how about we highlight the matching portion with the same salmon/red color we use for page links/the search button?

lametro/templates/partials/search_bar.html
```diff
@@ -20,3 +22,34 @@
         <input type="radio" name="search-reports" autocomplete="off" {% if 'search-reports' in request.GET %} checked{% endif %}> Search reports only
     </label>
 </div>
+
+{% block extra_js %}
```
hancush: putting this block here doesn't have quite the effect we want: it puts the javascript in the middle of the page. 

in order to add this to the end of the full search markup, you need to extend [the default `search.html` template](https://github.com/datamade/django-councilmatic/blob/master/councilmatic_core/templates/search/search.html). this will involve creating a `search.html` file at the same path in the `lametro/templates`, and putting these script elements inside an extended `extra_js` block. 

there's more guidance in [the django docs on template inheritance](https://docs.djangoproject.com/en/2.0/ref/templates/language/#template-inheritance). one thing in particular you will find useful in adding content to a block is this:

> If you need to get the content of the block from the parent template, the `{{ block.super }}` variable will do the trick. This is useful if you want to add to the contents of a parent block instead of completely overriding it.

(i just learned this looking for the docs to add to this comment! 🙃)

hancush:
btw, logging for testing purposes: here are current councilmatic subjects that are also in smartlogic, i.e., can be used to test the autocomplete flow:

```
lametro_councilmatic=# select distinct subject from lametro_subjectguid join councilmatic_core_subject on lametro_subjectguid.name = councilmatic_core_subject.subject;
                subject
---------------------------------------
 Title VI Requirements
 All Door Boarding
 Metro Green Line
 2ND QUARTER PERFORMANCE
 ENA (EXCLUSIVE NEGOTIATION AGREEMENT)
 RIDERSHIP INITIATIVES
 Measure R
 Measure M
(8 rows)
```

lametro/templates/partials/search_bar.html
jmithani: thank you! it didn't occur to me that the base search template would be a part of councilmatic. lacking that it seemed this was the best place over `base.html`.

lametro/templates/lametro/index.html
```diff
@@ -136,5 +136,40 @@ <h2>
...
+                    return {'value': d.name, 'data': d.id};
+                  }
+                )
+              };
+            }
+          },
+          onSearchError: function(query, jqXHR, textStatus, errorThrown) {
+            return textStatus + ' : ' + errorThrown;
+          },
+          onSelect: function(suggestion) {
```
jmithani: if 'No topics found' is clicked, it populates the search bar but doesn't lead anywhere.

lametro/templates/lametro/index.html
```diff
@@ -136,5 +136,40 @@ <h2>
...
+            $.when(
+              $.get('/topic/', { 'guid': suggestion.data }))
+                .then(function(response) {
+                  if (response.status_code == 200) {
```
jmithani: this doesn't appear to be working, locally or on staging.

lametro/views.py
```diff
@@ -811,3 +815,119 @@ def refresh_guid_trigger(request, refresh_key):
...
+        headers = {'Authorization': 'Bearer ' + self.access_token}
+
+        try:
+            response = requests.get(url, headers=headers)
+        except HTTPError as e:
+            print('Could not communicate with SmartLogic: {}'.format(e))
+        except Exception:
+            raise
+        else:
+            if response.status_code == 200:
```
jmithani: The issue here was when searching for "con," there was a 200 response from SmartLogic—it was the HTML of an error page. We handle the `json.JSONDecodeError` that causes here.

lametro/views.py
```diff
@@ -811,3 +815,119 @@ def refresh_guid_trigger(request, refresh_key):
...
+        session = self.request.session
+
+        if not session.get('smart_logic_token'):
+            session['smart_logic_token'] = self._generate_token()
+
+        return session['smart_logic_token']
+
+    def render_to_response(self, context):
+        '''
+        Return response as JSON.
```
jmithani: This helper needed to be updated after our changes to `get_queryset()`.

lametro/templates/lametro/index.html
jmithani: JK, i was just using a topic that wasn't synced.

lametro/templates/lametro/index.html
```diff
@@ -141,17 +141,28 @@ <h2>
...
+              return {
+                suggestions: $.map(res.termHints, function(d) {
+                    return {'value': d.name, 'data': d.id};
+                  }
+                )
+              };
             }
           },
+          onSearchError: function(query, jqXHR, textStatus, errorThrown) {
+            return textStatus + ' : ' + errorThrown;
```
hancush: what happens to this message? (does it get console logged or put somewhere we can see it if we try to reproduce the problem?)

lametro/templates/lametro/index.html
```diff
@@ -141,17 +141,28 @@ <h2>
...
           transformResult: function(response) {
-            var response = JSON.parse(response);
-            return {
-              suggestions: $.map(response.termHints, function(d) {
-                return {'value': d.name, 'data': d.id};
-              })
+            var res = JSON.parse(response);
+            if (res.status_code == 500 || res.termHints.length < 1) {
+              return {
+                suggestions: [{'value': 'No topics found', 'data': 'na'}]
```
hancush: nice catch/update! teeny point of clarification: does this lib let you return an array of strings? if so, would be good to do that here, to avoid creating dummy data if we don't need to.

lametro/templates/lametro/index.html
hancush: i think i'd prefer it to do nothing if clicked. is that possible?

lametro/views.py
hancush: nice catch!

hancush: can you leave a comment to that effect in the except block?

lametro/templates/lametro/index.html
jmithani: i think i dug into this with no success last week, but i'll look again with fresh eyes

lametro/templates/lametro/index.html
jmithani: it's not required and we're not using it—i went ahead and removed it since the errors weren't being utilized.

lametro/static/js/autocomplete.js
```diff
@@ -0,0 +1,39 @@
...
+    paramName: 'query',
+    transformResult: function(response) {
+      var res = JSON.parse(response);
+      if (res.status_code == 500 || res.termHints.length < 1) {
+        return {
+          suggestions: ['No topics found']
+        };
+      } else {
+        return {
+          suggestions: $.map(res.termHints, function(d) {
```
jmithani: This addresses the acronym request.

lametro/views.py
```diff
@@ -811,3 +815,123 @@ def refresh_guid_trigger(request, refresh_key):
     except AttributeError:
         print('You need a refresh_key in your local deployment settings files to access this.')
     return HttpResponse(403)
+
+
+class SmartLogicAPI(ListView):
+    api_key = SMART_LOGIC_KEY
+    query_format = 'https://cloud.smartlogic.com/svc/' \
+        + SMART_LOGIC_ENVIRONMENT + '/ses/CombinedModel/hints/{query}.json?FILTER=AT=System:%20Legistar'
```
jmithani: This filters to only suggest SES topics used in Legistar.

PR #446, opened 2019-05-21
Add server-side querying of SmartLogic API

hancush:
## Overview

This PR adds a pass-through route to query SmartLogic. Implementing this server-side allows us to keep our API credentials from being exposed in client-side code. 
 
### Checklist

### Notes

Some APIs, like Google Maps, include options to restrict requests by domain, so it is not a risk to expose keys in the source code. The SmartLogic API does not include this option, so in consultation with them, we have identified a server-side implementation as the best course of action. If this implementation proves too slow, the next-best option is to build an endpoint that exposes the expiring token, and do the querying in JavaScript, client-side.

## Testing Instructions

- Add the new SmartLogic values to your `settings_deployment.py` file. (Ping me for the API key.)
- Go to the homepage and try typing a value. Confirm that suggestions are auto-populated as you type.

Related to #433.


councilmatic/settings_deployment.py.example
```diff
@@ -80,3 +80,6 @@ LOGGING = {
 
 # Set to False in production!
 SHOW_TEST_EVENTS = True
+
+SMART_LOGIC_KEY = 'smartlogic api key'
+SMART_LOGIC_ENVIRONMENT = '0ef5d755-1f43-4a7e-8b06-7591bed8d453'
```
jmithani: based on what you said, i assume this doesn't need to be secret—just double-checking

hancush: ya, this is just the url. i made it configurable because i assume we are working in a test environment and we'll need to update it when we are ready to go live.

Issue #440, opened 2019-05-14
Home page: Display message when there is no live stream


shrayshray:
Unless people visit boardagendas.metro.net during a Board meeting, they might not be aware they can watch meetings live. 
When there is not a meeting in session, please display the following text in place of the "Watch in English"/"Watch in Spanish" buttons: "No meetings are in session at this time, The links to watch a meeting live, in English or Spanish, will be posted here once the meeting begins." 
Something like this:
![20190514-livestream-message](https://user-images.githubusercontent.com/28063988/57736204-01285500-765c-11e9-8c3f-eedf6c4b63fa.jpg)

jmithani:
deployed!

PR #447, opened 2019-05-28
Adding information about livestream to homepage when there is not a meeting in session

jmithani:
## Overview

This PR adds in text to explain where to find a livestream for the next board meeting when there isn't one currently in session.

### Checklist

### Demo

When there is an upcoming meeting scheduled...
Screen Shot 2019-05-28 at 11 29 49 AM

Switching out the partial for `meeting_details_current.html`, we can see the text is replaced by a link to the live stream.

Screen Shot 2019-05-28 at 11 35 12 AM

## Testing Instructions

The notice should be visible on the homepage.

Handles #440

### DIFF
--- a/lametro/templates/partials/meeting_details_next.html
+++ b/lametro/templates/partials/meeting_details_next.html
@@ -22,6 +22,11 @@
                       <i class="fa fa-fw fa-clock-o"></i> {{ meeting.start_time | date:"g:i a"}}<br/>
                       <i class="fa fa-fw fa-map-marker"></i> {{ meeting.location_name}}<br />
                   </p>
+                  {% if meeting.documents.all %}
+                    <div class="col-xs-7">
+                      <a class="btn btn-teal" id="pdf-download-link" target='_blank' href='{{meeting.documents.all|find_agenda_url}}'><i class='fa fa-fw fa-download'></i> Get Agenda PDF</a>
+                    </div>
+                  {% endif %}
                 </div>
               </div>
             </div>
@@ -32,11 +37,7 @@
             <p class="small" style="padding-top: 10px;"><strong>No meetings are in session at this time. The links to watch a meeting live, in English or Spanish, will be posted here once the meeting begins.</strong></p>
 
             <!-- Links to media url and PDF download -->
-            {% if meeting.documents.all %}
-              <div class="col-xs-7">
-                <a class="btn btn-teal" id="pdf-download-link" target='_blank' href='{{meeting.documents.all|find_agenda_url}}'><i class='fa fa-fw fa-download'></i> Get Agenda PDF</a>
-              </div>
-            {% else %}
+            {% if not meeting.documents.all %}
               <p class="small"><em>Agenda will be posted no later than 72 hours prior to the start of the meeting, or 24 hours for Special Board meetings.</em></p>
             {% endif %}
 
--- a/lametro/templates/partials/meeting_details_next.html
+++ b/lametro/templates/partials/meeting_details_next.html
@@ -6,42 +6,41 @@
         <div class="row" style="margin-top: 1em;">
             <div class="col-xs-6">
                 <img src='/static/images/Gateway03RT.jpg' class='img-responsive img-rounded' title='Los Angeles Gateway Plaza' alt="Los Angeles Gateway Plaza" />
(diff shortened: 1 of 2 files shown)


lametro/templates/partials/meeting_details_next.html
```diff
@@ -32,6 +32,7 @@ <h4>
                         {% else %}
                           <br>
                           <p class="small"><em>Agenda will be posted no later than 72 hours prior to the start of the meeting, or 24 hours for Special Board meetings.</em></p>
+                          <p><strong>No meetings are in session at this time, The links to watch a meeting live, in English or Spanish, will be posted here once the meeting begins.</strong></p>
```
hancush: can you make this `.small` like the preceding `<p>`?

hancush: also: let's display this message, regardless of whether there are agenda documents.

jmithani:
@hancush i moved it to be under the photo, thoughts? i don't think it's perfect but it looks nice on desktop.

Screen Shot 2019-05-29 at 9 28 43 AM
Screen Shot 2019-05-29 at 9 29 56 AM

hancush:
hm, good thought, @jmithani. the alignment is still a little wonky to me. what if we put both notices in a new row element beneath the columns containing the image / meeting details?

jmithani:
@hancush that definitely looks a lot better!
Screen Shot 2019-05-29 at 10 06 23 AM
Screen Shot 2019-05-29 at 10 06 44 AM

lametro/templates/partials/meeting_details_next.html
```diff
@@ -6,42 +6,41 @@
...
+              </div>
             </div>
+        </div>
+
+        <div class="row">
+          <div class="col-xs-12">
+            <p class="small" style="padding-top: 10px;"><strong>No meetings are in session at this time. The links to watch a meeting live, in English or Spanish, will be posted here once the meeting begins.</strong></p>
+
+            <!-- Links to media url and PDF download -->
+            {% if meeting.documents.all %}
```
hancush: when a link to the agenda becomes available, i think it would be good to keep it front and center in the meeting info col, rather than down here. what do you think?

lametro/templates/partials/meeting_details_next.html
```diff
@@ -8,39 +8,40 @@
...
+                    <span class="small text-muted">{{ meeting.description }}</span>
+                  </h4>
 
+                  <!-- Meeting info -->
+                  <p class="small text-muted">
+                      <i class="fa fa-fw fa-calendar-o"></i> {{ meeting.start_time | date:"D n/d/Y"}}<br/>
+                      <i class="fa fa-fw fa-clock-o"></i> {{ meeting.start_time | date:"g:i a"}}<br/>
+                      <i class="fa fa-fw fa-map-marker"></i> {{ meeting.location_name}}<br />
+                  </p>
+                  {% if meeting.documents.all %}
```
hancush: looking at the original, it seems like you're missing an enclosing `div.row` around this block! once you add that i'm happy to approve.

jmithani: pushed! @hancush

lametro/templates/partials/meeting_details_next.html
```diff
@@ -2,46 +2,49 @@
...
+              <i class="fa fa-fw fa-calendar-o"></i> {{ meeting.start_time | date:"D n/d/Y"}}<br/>
+              <i class="fa fa-fw fa-clock-o"></i> {{ meeting.start_time | date:"g:i a"}}<br/>
+              <i class="fa fa-fw fa-map-marker"></i> {{ meeting.location_name}}<br />
+            </p>
+            {% if meeting.documents.all %}
+            <div class="row">
+              <div class="col-xs-7">
+                <a class="btn btn-teal" id="pdf-download-link" target='_blank' href='{{meeting.documents.all|find_agenda_url}}'><i class='fa fa-fw fa-download'></i> Get Agenda PDF</a>
               </div>
+              {% endif %}
```
hancush: oops! if you open the row div inside the if statement, you should close it in there, too. otherwise you prematurely close the parent div if the if condition isn't satisfied. 
Screen Shot 2019-05-29 at 5 23 45 PM

Issue #348, opened 2018-09-18
Board Reports: Revise Columns on History


shrayshray:
In the "History" section of Board Report pages:
![image](https://user-images.githubusercontent.com/28063988/45721399-d59cab80-bb5c-11e8-90c6-4002073e8ba2.png)
1. Please change the header "Committee" to "Meeting", and
2. Add column for "Meeting Status".

No. 2 will help us provide greater clarity in cases where a meeting was cancelled but the report was routed to another meeting at a later date.

jmithani:
Hi @shrayshray, just wanted to confirm I understand your request. For "Meeting Status," are you referring to the status of the meeting that occurred on the specified date (wherein this report was discussed)? Or do you mean something else?

Thanks!

shrayshray:
@hancush hmm, i guess this is kind of confusing. Maybe we don't need the meeting status column. Would it be possible though to indicate when a meeting was cancelled, similar to the way it's shown on the meetings page? 
![image](https://user-images.githubusercontent.com/28063988/58656203-b3a11e80-82d0-11e9-81be-f9e182a58fc7.png)

One more thing -- Right now, the Meeting names just link to the page About the committee. Could this be changed so the link goes to the page for the actual meeting? E.g., the 9/20/17 Finance, Budget & Audit Committee meeting.
![image](https://user-images.githubusercontent.com/28063988/58656428-20b4b400-82d1-11e9-870e-323f52b95c8a.png)

jmithani:
Sure thing, we can replace those links with ones to the specific meeting, and we can do the strike-through too for cancelled meetings.

jmithani:
@shrayshray

@hancush and I have done some exploration of the database, and we can't find any registered actions on reports slated for cancelled meetings. (Is there an example report you had in mind you could share?) To track these actions, we would need to look at the agendas of the cancelled meetings to see if something was to be discussed then. That change is larger. Would you still like us to prioritize this?

jmithani:
There is no current use case for this tracking bills that were on the agendas of cancelled meetings. Closing this issue until that becomes a need.

shrayshray:
Here is an example: https://boardagendas.metro.net/board-report/2018-0606/
It was on the FINANCE, BUDGET AND AUDIT COMMITTEE agenda for OCTOBER 17, 2018 (cancelled) and the EXECUTIVE MANAGEMENT COMMITTEE agenda OCTOBER 18, 2018 (not cancelled).

[...]

PR #450, opened 2019-05-30
Updating report actions table with more information

jmithani:
## Overview

- [x] Changes "Committee" to "Meeting"
- [x] Links to relevant event instead of committee overview page
- [x] Displays "Cancelled" a la the events page

### Checklist

### Demo

Screen Shot 2019-05-31 at 2 54 51 PM
Screen Shot 2019-05-31 at 2 54 59 PM

### Notes

I used template tags instead of a manager because I wanted the function to occur per-object, not on a QuerySet. This seemed like the least complex way to do that.

## Testing Instructions

- Navigate to a piece of legislation
- Hard refresh the page
- Confirm link goes to appropriate event agenda
- Confirm event status is correct
- Repeat with a few different pages

Handles #348


lametro/models.py
```diff
@@ -7,7 +7,7 @@
 from django.conf import settings
 from django.db import models, connection
 from django.db.models.expressions import RawSQL
-from django.core.exceptions import ObjectDoesNotExist
+from django.core.exceptions import ObjectDoesNotExist, MultipleObjectsReturned
```
hancush:
```suggestion
from django.core.exceptions import ObjectDoesNotExist
```

lametro/templatetags/lametro_extras.py
```diff
@@ -2,6 +2,7 @@
 from django.template.defaultfilters import stringfilter
 from django.utils.html import strip_entities, strip_tags
 from django.utils import timezone
+from django.core.exceptions import MultipleObjectsReturned
```
hancush: this totally works, just a fun trivia bit: you can access this and the `ObjectDoesNotExist` exceptions directly from any django model, like `LAMetroEvent.MultipleObjectsReturned`.

lametro/templatetags/lametro_extras.py
```diff
@@ -175,5 +176,28 @@ def get_highlighted_attachment_text(context, ocd_id):
...
+    try:
+        event = LAMetroEvent.objects.get(participants__entity_type='organization',\
+                                        participants__entity_name=action.organization, \
+                                        start_time__date=action.date.date())
+    except MultipleObjectsReturned:
+        bill = action.bill()
+        event = LAMetroEvent.objects.get(participants__entity_type='organization',\
+                                        participants__entity_name=action.organization, \
+                                        start_time__date=action.date.date()) \
+                                        .get(agenda_items__bill=bill)
```
hancush: this seems like a smart filter to apply all the time. is there a reason to only check if there are multiple objects returned?

lametro/templatetags/lametro_extras.py
```diff
@@ -175,5 +176,28 @@ def get_highlighted_attachment_text(context, ocd_id):
     attachment_text = ' '.join(d.full_text for d in bill.documents.all() if d.full_text)
 
     highlight = ExactHighlighter(context['query'])
-    
+
     return highlight.highlight(attachment_text)
+
+
+def get_event_occured(action):
```
hancush: this name is a little confusing to me. how about more descriptive, like `get_action_event`?

lametro/templates/lametro/legislation.html
```diff
@@ -108,7 +108,12 @@ <h3 class="no-pad-bottom"><i class='fa fa-fw fa-list-ul'></i> History</h3>
                                         {% endif %}
                                     </td>
                                     <td class="small">
-                                        {{action.organization.link_html|safe}}
+                                      {% if action|get_event_occured|get_event_status == 'cancelled' %}<strike>{% endif %}
```
hancush: calling this twice will lead to duplicate queries. how about using some `with` logic to grab the event? here's an example from the index page:

https://github.com/datamade/la-metro-councilmatic/blob/25ab4a6489e0862b06d714c8a7dc529130dcb08f/lametro/templates/lametro/index.html#L72-L74

(the partial can then reference the meeting object like `{{ meeting }}`.)

hancush:
(did i mention this change looks dope?)

lametro/templatetags/lametro_extras.py
hancush: clarification: why do you only check to see whether the bill is on an event agenda if multiple events are returned? it seems like a good thing to filter on regardless.

jmithani: right now in the first block, trying to use `action.bill()` doesn't work because it tries to call `LAMetroBill` and that throws an error. i could condense it this way:
```
bill = action.bill()
event = LAMetroEvent.objects.filter(participants__entity_type='organization',\
                                        participants__entity_name=action.organization, \
                                        start_time__date=action.date.date()) \
                                        .get(agenda_items__bill=action._bill)
```

i think i misunderstood the privacy of `._bill` and avoided using it.

hancush: what is the difference between `LAMetroBill` and `_bill`? what error does `action.bill()` throw?

jmithani: here's the error:
```
----> 1 action.bill()
TypeError: 'LAMetroBill' object is not callable
```

[`action._bill`](https://github.com/datamade/django-councilmatic/blob/master/councilmatic_core/models.py#L522) is a field that links to the underlying `Bill` object. `action.bill()` calls the subclass ([`override_relation()`](https://github.com/datamade/django-councilmatic/blob/master/councilmatic_core/models.py#L34)), if any, instead—`LAMetroBill` in this case.

**ALL THAT BEING SAID**..............the problem was me _calling_ `action.bill()` instead of using it properly, like `action.bill`.

hancush:
@jmithani can you provide an example link to a bill with actions for a cancelled meeting?

jmithani:
@hancush i spent some time looking and couldn't find any either.

hancush:
@jmithani i am thinking it's because there aren't actions to record if a meeting didn't happen...! if we want to include when an item is scheduled for consideration during a meeting, i think we'd need to look for events where the bill appears on the agenda. that change is a little more involved. might be good to post an update in the issue explaining that the meeting status will be a bigger change than we thought and asking them if they'd still like us to prioritize it. what do you think?

jmithani:
@hancush that makes sense to me. also maybe there is an example they have in mind that can be shared.

jmithani:
There is no use case for this—the UI changes will never be seen. Closing until further development is warranted.

lametro/models.py
```diff
@@ -153,6 +153,56 @@ def board_report(self):
...
+
+        for action in actions:
+            event = LAMetroEvent.objects.filter(\
+                participants__entity_type='organization',\
+                participants__name=action.organization,
+                start_time__date=action.date)
+
+            
+            action_dict = {
+                'date': action.date_dt.date(),
```
hancush, line 182: `date_dt` changed in https://github.com/datamade/django-councilmatic/pull/272 -- it now returns a date object! (You may have to rebase your branch off master, then rebuild the app image with `docker-compose build app` to capture the upgraded django-councilmatic dependency to test this locally.)

lametro/templates/lametro/legislation.html
```diff
@@ -24,7 +24,6 @@ <h1>
                     {% if legislation.inferred_status %}
                         {{ legislation.inferred_status | inferred_status_label | safe }}
                     {% endif %}
-                    <a href="rss/" title="RSS feed for Legislation {{legislation.friendly_name}}" target="_blank"><i class="fa fa-rss-square" aria-hidden="true"></i></a>
```
hancush, line 27: Why remove this?

lametro/templates/lametro/legislation.html
```diff
@@ -98,20 +97,25 @@ <h3 class="no-pad-bottom"><i class='fa fa-fw fa-list-ul'></i> History</h3>
                             </tr>
                         </thead>
                         <tbody>
-                            {% for action in actions %}
+                            {% for event in events %}
```
hancush, line 100: I think `action` as a variable name may be a little bit more precise, since there can be more than one item corresponding to the same event (one being scheduled, with subsequent items for actions taken during the meeting). What do you think?

lametro/templates/lametro/legislation.html
```diff
@@ -98,20 +97,25 @@ <h3 class="no-pad-bottom"><i class='fa fa-fw fa-list-ul'></i> History</h3>
...
-                                        <span class='text-{{action.label}}'>{{action.description | remove_action_subj}}</span>
+                                        <span class='text-default'>{{event.description}}</span>
                                         {% if action.related_organization %}
                                             to
                                             {{action.related_organization.link_html|safe}}
                                         {% endif %}
                                     </td>
                                     <td class="small">
-                                        {{action.organization.link_html|safe}}
+                                        {% if action_event.status == 'cancelled' %}<strike>{% endif %}
```
hancush, line 113: What does `action_event` refer to?

lametro/templates/lametro/legislation.html
```diff
@@ -98,20 +97,25 @@ <h3 class="no-pad-bottom"><i class='fa fa-fw fa-list-ul'></i> History</h3>
...
+                            {% for event in events %}
                                 <tr>
                                     <td class='nowrap text-muted small'>
-                                        <span datetime='{{action.date_dt | format_date_sort }}'>{{action.date_dt|date:'n/d/y'}}</span>
+                                        <span datetime='{{event.date}}'>{{event.date}}</span>
                                     </td>
                                     <td class="small">
-                                        <span class='text-{{action.label}}'>{{action.description | remove_action_subj}}</span>
+                                        <span class='text-default'>{{event.description}}</span>
                                         {% if action.related_organization %}
```
hancush, line 107: Looks like you need to update this section to use your dictionary, as well -- it's operating as if we are still dealing with `Action` objects.

lametro/views.py
```diff
@@ -97,6 +96,8 @@ def get_context_data(self, **kwargs):
             .order_by('-latest_date')
 
         context['related_bills'] = related_bills
+        
+        context['events'] = self.get_object().actions_and_agendas
```
hancush, line 100: There are several calls to `self.get_object()` in this method. What do you think about doing one at the beginning of the method, then using the result for the subsequent operations, that way we aren't hitting the database every time it's referenced?

hancush: Another opportunity for clean up: `self.get_object().actions.all()` is called twice (and we can eliminate the first call, because you're replacing the definition of actions here).

hancush (changes requested):
This is looking awesome, @fatima3558! Your queries make sense to me, and the corresponding items are showing up _almost_ as I'd expect, e.g.,

Screen Shot 2020-10-02 at 11 27 24 AM

Looks like there are a few more updates needed in the template, and I left some suggestions inline for cleaning up the context data method. Look forward to re-reviewing. :-)

hancush:
Superseded by #656.

Issue #461, opened 2019-06-27
highlight matched tags in search results


hancush:
if the full text of a tag appears in a search query, highlight that tag.

jmithani:
Merged in #466

Issue #459, opened 2019-06-27
add the option to toggle tag results on/off to the search result page


hancush:
<img width="1440" alt="Screen Shot 2019-06-27 at 11 50 14 AM" src="https://user-images.githubusercontent.com/12176173/60284901-4cc45480-98d2-11e9-9047-e9140bd47bc5.png">

jmithani:
Merged in #466

Issue #458, opened 2019-06-27
issue keyword search, not topic search, from autosuggested terms


hancush:
we currently execute a topic search if a user clicks an autosuggested term. sometimes, reports are tagged but do not contain a search term, and vice versa. this leads to inconsistent results, depending on the type of search issued.

- [ ] instead of executing a topic search when a term is selected, execute a plain text search. that way, all initial searches are keyword searches.
- [x] confirm that topics are indexed with a bill's text, so both tagged and keyword results are returned from a keyword search. if they aren't, add them.

hancush:
it may be a good idea to implement the hints reset functionality described in #460, as part of this issue.

PR #466, opened 2019-07-03
Add small fixes to autocomplete search functionality

hancush:
## Overview

This PR:

* Executes text search on autocomplete select (#458)
* Adds result type filter to result page (#459)
* Highlights tags matching query (#461)

### Checklist

### Demo

Screen Shot 2019-07-03 at 12 44 34 PM
_Result type filter and highlighted topic._

### Notes

There are some tradeoffs to the way I implemented topic-only search.

The current implementation will perform case-insensitive search on the topics, but it will also include topics that only contain the term, e.g., "bus" matches "BUS" but it also matches "BUS IMPROVEMENTS". 

We could also search on `topics_exact`, however the search becomes case sensitive. I think we could get around this by defining a custom string field type that lowercases things before comparing them in the Solr index, then searching on `topics_exact`, but I want to check with Shelly and Omar about case sensitivity before we go that more involved route.

## Testing Instructions

* Refresh your local subject GUIDs: `python manage.py refresh_guid`
* Run the app: `python manage.py runserver`
* Search "Metro Green Line" and click the autosuggested topic. Confirm that a text search is run.
* On the results page, confirm that the "Metro Green Line" tag is highlighted in relevant results. Take note of the total number of results
* On the results page, click "Keyword only" and confirm that results tagged "Metro Green Line" do not appear in the search. Take note of the number of keyword results.
* On the results page, click "Topic only" and confirm that only results tagged "Metro Green Line" appear in the search. Take note of the number of topic results.
* Confirm that the number of keyword results plus the number of topic results equals the number of total results.

### DIFF
--- a/lametro/templates/base.html
+++ b/lametro/templates/base.html
@@ -106,15 +106,14 @@
     <script type="text/javascript">
         $('label.btn.btn-search-option').click(function toggleRadio(e) {
             // Toggle "active" appearance first, so filtering on active works.
-            // .button docs: https://getbootstrap.com/docs/3.3/javascript/#buttons
-            $(this).button('toggle');
+            $('label.btn.btn-search-option.active').removeClass('active');
+            $(this).addClass('active');
 
             var selected_input = $(this).children('input')[0];
             $(selected_input).attr('checked', true);
 
-            var unselected_input = $('label.btn').not('.active')
-                                                 .children('input')[0];
-            $(unselected_input).removeAttr('checked');
+            var unselected_input = $('label.btn').not('.active').children('input')[0];
+            $(unselected_input).attr('checked', false);
         });
     </script>
 
--- a/lametro/templates/search/search.html
+++ b/lametro/templates/search/search.html
@@ -1,6 +1,279 @@
 {% extends "search/search.html" %}
 
 {% load staticfiles %}
+{% load extras %}
+
+{% block content %}
+
+    <br/>
+    <div class='jumbotron' id='search-jumbotron'>
+        <!-- Add partial, so the title above search can be easily customized for site. -->
+        {% include 'partials/search_header.html' %}
+
+        <form method="get" action=".">
+            <div class="input-group site-intro-search">
+
+                {% for facet, selections in selected_facets.items %}
+                    {% for s in selections %}
+                        <input name="selected_facets" type="hidden" class="form-control" value="{{facet}}_exact:{{s}}">
+                    {% endfor %}
+                {% endfor %}
+
+                {% if 'sort_by=date' in request.get_full_path %}
+                    <input name="sort_by" type="hidden" class="form-control" value="date">
+                {% endif %}
+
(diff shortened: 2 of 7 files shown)


lametro/templates/base.html
```diff
@@ -106,15 +106,14 @@
     <script type="text/javascript">
         $('label.btn.btn-search-option').click(function toggleRadio(e) {
             // Toggle "active" appearance first, so filtering on active works.
-            // .button docs: https://getbootstrap.com/docs/3.3/javascript/#buttons
-            $(this).button('toggle');
+            $('label.btn.btn-search-option.active').removeClass('active');
```
hancush: for some reason this was not working on my machine, so i updated it.

lametro/templates/search/search.html
```diff
@@ -1,6 +1,279 @@
 {% extends "search/search.html" %}
 
 {% load staticfiles %}
+{% load extras %}
+
+{% block content %}
```
hancush: i had to override the entire content block in order to add the result type filter. i think this is ok, especially since we will probably be doing additional search customizations for metro moving forward. will leave comments where i actually made changes!

lametro/templates/search/search.html
```diff
@@ -1,6 +1,279 @@
...
+                {% endif %}
+
+                {% with formatted_q=request.GET.q|remove_question %}
+                    {% include 'partials/search_bar.html' %}
+                {% endwith %}
+            </div>
+        </form>
+    </div>
+
+    <div class="row" style="margin-left:unset;">
```
hancush: the left side of the search results is not even with the left side of the search bar element. this fixes that.

lametro/templates/search/search.html
```diff
@@ -1,6 +1,279 @@
...
+        <div class="col-sm-4">
+            <p id="result-type-filter">
+                Result type:
```
hancush: this result type filter block is new.

lametro/views.py
```diff
@@ -717,6 +719,11 @@ def search(self):
             # Don't auto-escape my query! https://django-haystack.readthedocs.io/en/v2.4.1/searchqueryset_api.html#SearchQuerySet.filter
             sqs = sqs.filter_or(attachment_text=Raw(self.cleaned_data['q']))
 
+        if self.result_type == 'keyword':
+            sqs = sqs.exclude(topics__iexact=Exact(self.cleaned_data['q']))
```
hancush: the `Exact` input type looks for the full phrase: https://django-haystack.readthedocs.io/en/v2.4.1/inputtypes.html#exact

jmithani:
this looks really good @hancush!! great job! ✨

for the highlighted tag, i didn't register the color change at first. what do you think about also bolding the tag text? that's a deviation from the other highlights, but the tag is much smaller text.

Screen Shot 2019-07-16 at 3 51 07 PM

hancush:
@jmithani thank you! i'm into that suggestion. will push a change tomorrow.

PR #491, opened 2019-09-11
Adding tags to sidebar on legislation page

jmithani:
## Overview

This adds in a list of a report's tags to the detail page.

### Checklist

### Demo

Screen Shot 2019-09-11 at 3 53 21 PM

## Testing Instructions

- Search for a piece of legislation.
- Confirm that when you click on it, the tags displayed in the search result match those in the sidebar.
- Click on the tag links and confirm they open corresponding searches in a new tab.

Handles #XXX

### DIFF
--- a/lametro/templates/lametro/legislation.html
+++ b/lametro/templates/lametro/legislation.html
@@ -146,15 +146,15 @@
                         <thead>
                             <tr>
                                 <th>Board Report</th>
-                                <th class='no-wrap'>Last action</th>                            
+                                <th class='no-wrap'>Last action</th>
                             </tr>
                         </thead>
                         <tbody>
                             {% for bill in related_bills %}
                                 <tr>
                                     <td class='text-muted small'>
-                                        <a href="{% url 'lametro:bill_detail' bill.slug %}">{{ bill.identifier}}</a> 
-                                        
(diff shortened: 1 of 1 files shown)


lametro/templates/lametro/legislation.html
```diff
@@ -169,6 +168,22 @@ <h3 class="no-pad-bottom"><i class="fa fa-file-text" aria-hidden="true"></i> Rel
                 </div>
                 <div class="divider"></div>
             {% endif %}
+
+            {% if legislation.topics %}
+            <br>
```
jmithani: i moved the `<br>` to inside the conditionals to prevent two of them stacking up.

lametro/templates/lametro/legislation.html
```diff
@@ -169,6 +168,22 @@ <h3 class="no-pad-bottom"><i class="fa fa-file-text" aria-hidden="true"></i> Rel
...
+              <h3><i class="fa fa-tag" aria-hidden="true"></i> Tags</h3>
```
hancush: should we stick with "topics" for consistency?

jmithani: good q! i tried that and it didn't work—it looks like `attachments` and `related_bills` are separate context variables. 

i'll add in a context variable for `topics` for consistency.

Issue #493, opened 2019-09-13
Clarify distinction between selecting a topic and performing a keyword search


hancush:
Add a sentence or help text above the search bar, e.g., "Press enter to perform a keyword search, or select from suggested topics as they appear" and add the prompt to perform a keyword search when no suggestions are found.

shrayshray:
@hancush I don’t see the new instructions in the dropdown, it still only says “No topics found” when there are no matching concepts. Am I missing something?

hancush:
Nope, @shrayshray, I am. I've done a little housekeeping with our development streams and the autocomplete deployment on staging should now reflect the most current changes. 

There's one expected bug: If you enter a non-preferred term in the search bar, the preferred term suggestion won't include your query. For example, if you enter "HOT", you'll see "High occupancy toll lanes" instead of "High occupancy toll lanes (HOT)" in the suggestions. I broke this when I implemented the new suggestion strategy (#494) but will fix it, once we have your input on whether the change has improved the quality of the suggestions!

PR #497, opened 2019-09-19
Add clarifying search instructions

jmithani:
## Overview

Changes the language of the autocomplete feature.

### Checklist

### Demo

Screen Shot 2019-09-19 at 11 33 35 AM

## Testing Instructions

- Go to the homepage. 
- Confirm the instructions are above the search bar and are legible.
- Search for something nonsensical, and confirm the language in the suggestions box is correct.
- Execute a search.
- Confirm the instructions appear above the search bar on the search results page.

Handles #493

### DIFF
--- a/lametro/templates/lametro/index.html
+++ b/lametro/templates/lametro/index.html
@@ -7,12 +7,13 @@
             <div class='col-sm-10 col-sm-offset-1'>
                 <span class="hidden-xs hidden-sm"><br/><br/></span>
                     <h1 class="home-header">Los Angeles County Metropolitan Transportation Authority</h1>
-                <span class="hidden-xs hidden-sm"><br/></span>
                 <p class="h3">
                     <span class="text-bg">Search Metro Board Reports (2015 – present)</span>
                 </p>
-                <p style="color: #ffffff;font-weight: bold;">
-                  Press enter to perform a keyword search, or select from suggested topics as they appear.
+                <p>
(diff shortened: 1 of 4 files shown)


lametro/templates/lametro/index.html
```diff
@@ -11,6 +11,9 @@ <h1 class="home-header">Los Angeles County Metropolitan Transportation Authority
                 <p class="h3">
                     <span class="text-bg">Search Metro Board Reports (2015 – present)</span>
                 </p>
+                <p style="color: #ffffff;font-weight: bold;">
```
hancush: i'm not sure i like the bold text, but i understand it's tough to read with the background image. what do you think about increasing the opacity of the white in the gradient overlay, in order to fade the image a bit, and using normal font weight here instead?

hancush: it'd also be preferable not to use inline style, if possible. does `text-bg` do the trick?

lametro/templates/partials/search_header.html
```diff
@@ -1,3 +1,8 @@
 <p class="h4">
     Search Metro Board Reports
-</p>
\ No newline at end of file
+</p>
+<p>
+  <small>
```
hancush: Screen Shot 2019-09-19 at 1 35 32 PM

this is actually quite large on my screen. what do you think about a smaller font size?

PR #510, opened 2019-11-20
2.5 models, views, templates and ancillary files

fgregg:
## Overview

This PR makes changes to the models, views, templates, urls and ancillary files for the councilmatic 2.5 upgrade. This PR does not include the map view for the board members page, which will come in a separate PR.

### Notes

This PR depends on changes to the scrapers and django-councilmatic. I believe they can be reviewed separately

- https://github.com/opencivicdata/scrapers-us-municipal/pull/299
- https://github.com/datamade/django-councilmatic/pull/260

Also, we are bit out of synch with LA metro master, we should discuss how we want to handle that.

## Testing Instructions

Move `councilmatic/settings_deployment.py.example` to `councilmatic/settings_deployment.py`.

```bash
docker-compose up scrapers
docker-compose run --rm -e DJANGO_MANAGEPY_MIGRATE=off app python manage.py migrate
convert_attachment_text
docker-compose run --rm -e DJANGO_MANAGEPY_MIGRATE=off app python manage.py update_index --batch-size=100
docker-compose up
```


lametro/templates/lametro/event.html
```diff
@@ -17,8 +17,8 @@ <h1>
                     {{event.name}}
                 {% endif%}
 
-                {% for media in event.media_urls.all %}
-                    <a class="btn btn-salmon" href="{{ media.url }}" target="_blank"><i class="fa fa-headphones" aria-hidden="true"></i> {{ media.label }}</a>
+                {% for media in event.media.all %}
+                    <a class="btn btn-salmon" href="{{ media.links.all.0.url }}" target="_blank"><i class= "fa fa-headphones" aria-hidden="true"></i> {% if media.note == 'Audio (SAP)' %}Ver en Español{% else %}Watch in English{% endif %}</a>
```
hancush: I preferred having this functionality defined as a property of the media model. What was your rationale for moving it into the template?

lametro/templates/partials/past_event_day.html
```diff
@@ -63,7 +63,7 @@ <h5 class="text-default">
         {% endif %}
     </div>
     <div class="col-xs-1">
-        {% with minutes=event.board_event_minutes %}
+        {% with minutes=event.minutes.0.links.all.0.url %}
```
hancush: This is a long chain of index and attribute access. Could it be moved up into a revised `board_event_minutes` attribute of the Event model? Also, is it safe to assume that the first link to the first minutes file will always be what we want here?

lametro/views.py
```diff
@@ -142,110 +150,56 @@ def get_context_data(self, **kwargs):
...
+
+        agenda_with_board_reports = event.agenda\
+            .filter(related_entities__bill__versions__isnull=False)\
+            .annotate(int_order=Cast('order', IntegerField()))\
+            .order_by('int_order')
+        
+        # Find agenda link.
+        if event.documents.all():
+            for document in event.documents.all():
+                if "Agenda" in document.note:
```
hancush: Should we break after one of these conditions is met?

lametro/views.py
```diff
@@ -493,46 +373,45 @@ class LAMetroAboutView(AboutView):
...
-    def get_context_data(self, **kwargs):
-        context = super().get_context_data(**kwargs)
+    def get_queryset(self):
+        '''
+        We only want committees that have at least one member who is not
+        the CEO. We also want to not count the CEO in the committee
+        size
 
-        with connection.cursor() as cursor:
+        '''
```
hancush: Bravo! 👏

lametro/views.py
```diff
@@ -548,65 +427,29 @@ def get_context_data(self, **kwargs):
...
-
-            columns = [c[0] for c in cursor.description]
-            results_tuple = namedtuple('Member', columns)
-            objects_list = [results_tuple(*r) for r in cursor]
-
-            context['ad_hoc_list'] = objects_list
-
-            context['ceo'] = Person.objects.filter(memberships__role='Chief Executive Officer',                                   memberships___organization=committee.ocd_id).first()
+        try:
+            ceo = Membership.objects\
```
hancush: Since you're using this in a few places, `ceo` might also make a good custom method on the `Membership` manager (returning the Membership or None), or maybe on the `Person` model. Thoughts?

lametro/views.py
```diff
@@ -548,65 +427,29 @@ def get_context_data(self, **kwargs):
...
+                .get(post__role='Chief Executive Officer',
+                     end_date_dt__gt=Now())\
+                .person
+        except Membership.DoesNotExist:
+            ceo = None
+
+        non_ceos = committee.all_members\
+            .annotate(index=Case(
```
hancush: Likewise, this annotation could be a manager method, e.g., `with_display_index` (since you are sorting differently here vs. in line 501). Or, if you don't mind sorting by index and name in both instances, you could roll it all up into `in_display_order`.

councilmatic/settings_jurisdiction.py
```diff
@@ -47,12 +47,13 @@
...
 BOUNDARY_SET = ['la-metro-supervisory-districts', 'la-metro-committee-districts', 'city-of-la']
 
-MAP_CONFIG = {
-    'center': [34.0522, -118.2437],
-    'zoom': 10,
-    'color': "#54afe8",
-    'highlight_color': '#eb6864'
-}
+MAP_CONFIG = False
+# MAP_CONFIG = {
```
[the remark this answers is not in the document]
fgregg: I just commented this out to facilitate handling the map in a separate PR. Please let me know if you'd like to change.

docker-compose.yml
```diff
@@ -61,6 +52,18 @@ services:
       SOLR_LOG_LEVEL: INFO
     restart: on-failure
 
+  scrapers:
```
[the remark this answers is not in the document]
fgregg: Can I save changes to README to integrate the whole of changes?

lametro/templates/partials/past_event_day.html
fgregg: doing this way takes advantage of the prefetching. Doing it as view or model attributes leads to a lot of queries and really slows this page down a lot. Maybe that's fine if we are caching.

lametro/views.py
fgregg: yup, good idea.

lametro/views.py
fgregg: good idea.

lametro/views.py
fgregg: I'd like to handle this, and some other things in https://github.com/datamade/django-councilmatic/issues/261

Can this wait till then.

lametro/templates/partials/past_event_day.html
fgregg: lmk.

lametro/models.py
```diff
@@ -120,23 +134,29 @@ def get_last_action_date(self):
         events that have already occurred, so the last action date is not in the
         future.
         '''
-        actions = Action.objects.filter(_bill_id=self.ocd_id)
-        last_action_date = ''
+        try:
+            return self.actions.last().date_dt
+        except AttributeError:
+            return None
```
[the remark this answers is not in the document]
fgregg: I'd like to handle that here: https://github.com/opencivicdata/scrapers-us-municipal/issues/300

lametro/models.py
```diff
@@ -145,38 +165,31 @@ class Meta:
...
-        represent a jurisdiction; that's not the case w la metro. just
-        need to know whether member is current or not...
-        '''
-        m = self.latest_council_membership
-        if m:
-            end_date = m.end_date
-            today = timezone.now().date()
-            return True if today < end_date else False
+        # We want to exclude memberships like 1st chair and just
+        # get the memberships confer membership to the org
```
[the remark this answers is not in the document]
fgregg: i think it's all pretty confusing, but sure!

lametro/models.py
```diff
@@ -145,38 +165,31 @@ class Meta:
...
+        #
+        # see https://github.com/opencivicdata/python-opencivicdata/issues/129
+        primary_memberships = city_council_memberships.filter(Q(role='Board Member') |
+                                                              Q(role='Nonvoting Board Member'))
+
+        if primary_memberships.count():
```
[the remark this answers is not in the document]
fgregg: cool.

lametro/models.py
```diff
@@ -191,39 +204,40 @@ def latest_council_seat(self):
         pass
 
     @property
+    def board_office(self):
```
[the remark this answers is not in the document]
fgregg: good idea, seems like that could be done in another PR?

lametro/models.py
```diff
@@ -191,39 +204,40 @@ def latest_council_seat(self):
...
-
-            bills = LAMetroBill.objects.filter(ocd_id__in=bill_ids)
-
-        return bills
-
-
-class LAMetroEventManager(models.Manager):
+        qs = LAMetroBill.objects\
+            .defer('extras')\
+            .filter(actions__organization__classification='committee')\
```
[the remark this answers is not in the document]
fgregg: it evaluates to the same thing? 

Do we have a style guide for this type of thing?

lametro/models.py
```diff
@@ -248,14 +262,15 @@ def with_media(self):
...
+        mediaqueryset = EventMedia.objects.annotate(
             olabel=Case(
                 When(note__endswith='(SAP)', then=Value(0)),
                 output_field=models.CharField(),
             )
         ).order_by('-olabel')
 
-        return self.prefetch_related(Prefetch('media_urls', queryset=mediaqueryset))
+        return self.prefetch_related(Prefetch('media', queryset=mediaqueryset))\
+                   .prefetch_related('media__links')
```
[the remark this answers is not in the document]
fgregg: I'm not sure. I'd like to keep the prefetching together if that's okay.

lametro/templates/lametro/event.html
fgregg: prefetching.

lametro/models.py
```diff
@@ -459,50 +473,25 @@ def upcoming_committee_meetings(cls):
         return meetings
 
 
-    @property
-    def board_event_minutes(self):
```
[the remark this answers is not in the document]
fgregg: it's replaced by https://github.com/datamade/la-metro-councilmatic/pull/510/files/677148a15b34f359049ccc7e2cdee1866626d6f4#r350338234

councilmatic/settings.py
```diff
@@ -119,3 +119,7 @@
 HAYSTACK_SIGNAL_PROCESSOR = 'haystack.signals.RealtimeSignalProcessor'
 
 ADV_CACHE_INCLUDE_PK = True
+
+import socket
```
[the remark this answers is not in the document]
fgregg: okee doke.

lametro/models.py
```diff
@@ -46,16 +61,17 @@ def get_queryset(self):
...
-                                            .filter(Q(related_agenda_items__event__status='passed') | \
-                                                    Q(related_agenda_items__event__status='cancelled') | \
-                                                    Q(bill_type='Board Box'))\
-                                            .distinct()
+        filtered_qs = super().get_queryset()
+        # .exclude(restrict_view=True)\
+        #                                     .filter(Q(related_agenda_items__event__status='passed') | \
+        #                                             Q(related_agenda_items__event__status='cancelled') | \
+        #                                             Q(bill_type='Board Box'))\
+        #                                     .distinct()
```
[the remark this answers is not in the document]
fgregg: will do.

lametro/templates/partials/past_event_day.html
fgregg: [this article](https://blog.mounirmesselmeni.de/2017/02/06/accessing-prefetched-objects-inside-your-models-methods/) suggests a way of taking advantage of prefetching while having method encapsulation. It's not pretty, but might be an option if we care about non-cached performance.

lametro/models.py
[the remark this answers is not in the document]
fgregg: yes, replaced by this. https://github.com/datamade/la-metro-councilmatic/pull/510/files/677148a15b34f359049ccc7e2cdee1866626d6f4#r350338234

lametro/views.py
fgregg: actually, I'm not sure. the current code doesn't do that. seems like we might want to have more than once of these in the view context.

fgregg:
@hancush thank you for the review! I've addressed some of your very good suggestions. Others, at your approval, I would like to defer to other PRs.

I do think that we should resolve the infelicities around pre-fetching in this PR, but I require your more detailed guidance.

lametro/views.py
fgregg: done!

councilmatic/settings_jurisdiction.py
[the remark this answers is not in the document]
hancush: Hm, I think it's ok in this case. Thanks!

docker-compose.yml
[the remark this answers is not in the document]
hancush: Sure!

lametro/models.py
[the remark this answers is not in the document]
hancush: Will that come in with https://github.com/opencivicdata/scrapers-us-municipal/pull/299?

lametro/models.py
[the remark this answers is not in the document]
hancush: Hmmmmmmmmmm, TIL that chained `filter` calls are _not_ always the same as combining terms in `filter` call. So, chaining filter calls here will return bills with committee actions **and**  bills with actions from current members, not bills with committee actions from current members. 

The language is a little confusing, so here's another example [from the docs](https://docs.djangoproject.com/en/dev/topics/db/queries/#spanning-multi-valued-relationships) and [a StackOverflow thread](https://stackoverflow.com/questions/8164675/chaining-multiple-filter-in-django-is-this-a-bug). 

It seems like you want results to meet both filter conditions, yes? If so, you'll need to combine the terms.

lametro/models.py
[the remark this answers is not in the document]
hancush: Totally fine, I was just wondering – for science! 🔬

lametro/views.py
hancush: It seems like this particular annotation is specific to Metro, no?

lametro/models.py
```diff
@@ -236,6 +233,18 @@ def committee_sponsorships(self):
 
         return qs
 
+    @classmethod
```
hancush: It feels more conventional to define as a method on a Person manager, rather than on the Person class itself. What was your thinking with the `classmethod`?

fgregg: I suppose this seemed more conventional to me.

1. I think about methods on managers returning querysets (I would expect that `LAMetroPerson.objects.ceo()` would return a queryset)
2. seems similar to `upcoming_board_meeting` or `current_meeting` on the LAMetroOrganization classes.

lametro/templates/lametro/event.html
hancush: Hm, I'm a little confused why returning a proxy model (LAMetroEventMedia) instead of the base model (EventMedia) interferes with the prefetching in this instance, since we aren't applying any additional filters to the queryset. What have I missed?  🙃

lametro/models.py
hancush: Ah, ok! Sure, that makes sense to me.

lametro/views.py
fgregg: you are right. I'd actually like to handle this in the scraper (through the extras field)

lametro/models.py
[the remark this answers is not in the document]
fgregg: yes. it's not written yes, but yes.

lametro/models.py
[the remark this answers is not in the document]
fgregg: This is giving me what I want.

Screenshot_2019-11-25 Kathryn Barger - Metro Board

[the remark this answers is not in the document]
fgregg: combining them into the same select produces the same SQL.

lametro/templates/lametro/event.html
fgregg: it's not about the proxy model.

its that we have prefetch media and media links, If I did something like

`media.url` that would probably look something like this.

```python
class LAMetroEventMedia(...):
    def url(self):
        return self.links.get().url
```

calling that method would trigger a new query, and I wouldn't get the advantage of prefetching.

lametro/views.py
hancush: Generally, I'm in favor of moving things upstream, but this feels much more related to display than the legislative data itself. Beyond that, it would be nice to keep display order ephemeral, so we can change it without writing and running a database migration. Are you strongly committed to handling it in the scraper?

fgregg: okay, sure. can I handle moving this to the manager in separate PR.

fgregg: sure, makes sense. Can I handle moving the annotations to the manager in a separate PR?

lametro/templates/partials/past_event_day.html
hancush: Ah yes, that article is an old frenemy... What if you moved the document prefetching into a manager method akin to `with_media` and used that manager method in a revised `board_event_minutes` attribute? (Do let me know if this is starting to feel like beating a dead  horse – appreciate you taking the time to talk through this with me!)

lametro/views.py
hancush: Sure thing. Thank you!

lametro/templates/partials/past_event_day.html
hancush: Oh, no, that won't work, will it – you'd still call the prefetch it for every event. Ok, I can tolerate the template code. Again, thanks for humoring me.

lametro/models.py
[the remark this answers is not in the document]
hancush: Sure, sounds good.

lametro/templates/lametro/event.html
hancush: Ah! I was referring to the label display logic, not the URL reference. Sorry for the confusion!!

lametro/models.py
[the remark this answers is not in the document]
hancush: That's very mysterious to me. If combining them produces the same SQL, I'd prefer you combine them to avoid the potential for subtle bugs, should the chained method function as advertised in other circumstances – that is:

> Successive filter() calls further restrict the set of objects, but for multi-valued relations, they apply to any object linked to the primary model, not necessarily those objects that were selected by an earlier filter() call.

lametro/templates/lametro/event.html
fgregg: oh, it seemed like pure display logic. 

I suppose something like

`media.is_spanish` could be a bit nicer than `media.note == 'Audio(SAP)'`

fgregg: or event

`media.language == 'Spanish'`

lametro/models.py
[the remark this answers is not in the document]
fgregg: okee doke, done.

lametro/templates/lametro/event.html
hancush: I liked the original property, which returned the display label, because you could make one change to update the label everywhere.

https://github.com/datamade/la-metro-councilmatic/blob/5f650c07c4a4ec665e6f3afc3284efa6cee50f90/lametro/models.py#L493-L502

What do you think?

lametro/models.py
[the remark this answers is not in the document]
hancush: From the bottom of my superstitious heart, thanks very much.

lametro/templates/lametro/event.html
fgregg: i think that model's shouldn't be responsible for display logic.

hancush: Ok, how about a template tag then?

fgregg: the previous model attribute was only used in two templates. I think the clarity of where the display is defined is worth this level of duplication.

hancush: Ok. I'm happy to agree to disagree on this minor point. No change required here.

PR #513, opened 2019-11-26
handle map

fgregg:
## Overview

This PR makes handles the map for the board member page and the individual members pages.

## Testing Instructions

Move `councilmatic/settings_deployment.py.example` to `councilmatic/settings_deployment.py`.

If setting up for the first time.
```bash
docker-compose up scrapers
docker-compose run --rm -e DJANGO_MANAGEPY_MIGRATE=off app python manage.py migrate
convert_attachment_text
docker-compose run --rm -e DJANGO_MANAGEPY_MIGRATE=off app python manage.py update_index --batch-size=100
docker-compose up
```

If you have existing local repo
```bash
docker-compose down
docker-compose build app
docker-compose up
```

Check out the board members page, and ensure that hover on the map and hover on the entry on the table works as expected. For example on the City of LA layer, hovering on Garcetti should highlight the map but that the map should not highlight for anyone else. 

Check detail pages.

### DIFF
--- a/data/scripts/ocd2geojson.py
+++ b/data/scripts/ocd2geojson.py
@@ -1,0 +1,22 @@
+import sys
+import json
+
+ocd_map = json.load(sys.stdin)
+geojson = {'type': "FeatureCollection",
+            'features': []}
+
+name = sys.argv[1]
+
+try:
+    for obj in ocd_map['objects']:
+        feature = {'type': "Feature"}
+        feature['geometry'] = obj['shape']
+        feature['properties'] = {'name': obj['name']}
+        geojson['features'].append(feature)
+except KeyError:
+        feature = {'type': "Feature"}
+        feature['geometry'] = ocd_map
+        feature['properties'] = {'name': name}
+        geojson['features'].append(feature)
+
+json.dump(geojson, sys.stdout)
--- a/lametro/templates/lametro/person.html
+++ b/lametro/templates/lametro/person.html
@@ -6,7 +6,7 @@
 {% block title %}{{ person.name }}{% endblock %}
 
 {% block extra_css %}
-    {% if map_geojson %}
+    {% if MAP_CONFIG %}
         <link rel="stylesheet" href="{% static 'css/leaflet.css' %}" />
     {% endif %}
 {% endblock %}
@@ -56,7 +56,7 @@
                 <p class="small">{{ member_bio | safe }}</p>
             {% endif %}
 
-            {% if map_geojson %}
+            {% if MAP_CONFIG %}
                 <hr />
                 <h4>
                     {% if person.current_district %}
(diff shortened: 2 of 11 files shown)


data/scripts/ocd2geojson.py
```diff
@@ -0,0 +1,22 @@
+import sys
+import json
+
+ocd_map = json.load(sys.stdin)
+geojson = {'type': "FeatureCollection",
+            'features': []}
+
+name = sys.argv[1]
+
+try:
```
hancush: This could use some additional context, namely: What are the potential shapes of the data from the OCD API? Why is there more than one?

lametro/views.py
```diff
@@ -341,7 +341,50 @@ class LABoardMembersView(CouncilMembersView):
 
     def map(self):
 
-        return {}
+        maps = {'map_geojson': {'type': 'FeatureCollection',
```
hancush: Can you call this `map_geojson_districts` for consistency?

fgregg: there's a bunch of template code that expects there to be an object called `map_geojson` I'm happy to make this change, but it is a bigger change.

data/scripts/ocd2geojson.py
fgregg: i don't think we'll re-use this code except to create boundary.geojson this one time. 

If boundaries change, we'll not stuff them into ocd the API got produce boundary.geojson. 

What about just adding a README in the data directory that gives this context.

hancush: Sounds good to me!

fgregg:
I don't have insight into the speed issue, I suspect it may be addressed by the caching. Let's wait and see on that?

lametro/views.py
hancush: Hm, I can live without it, if it's going to take that much work.

fgregg:
only the mayor should be highlight.

lametro/views.py
fgregg: it's not too much work, just wanted to flag that it's not a one-line change.

hancush: Got it. Holler at me when it's ready!

hancush:
@fgregg Okey doke, I agree we can wait and see what caching does for us.

fgregg:
responded to your requests.

lametro/templates/lametro/person.html
```diff
@@ -191,7 +191,7 @@ <h3>Committees</h3>
 
 {% block extra_js %}
 
-    {% if map_geojson %}
+    {% if map_geojson_districts %}
```
hancush: Why check for `MAP_CONFIG` everywhere but here?

fgregg: because it looks like this section should actually only be loaded if the the districts exist.

to be honest, this code looks like it's has bug, as it would make more sense to load the relevant map for all the types of of board members, not just the commissioners. however, this current behavior

Issue #361, opened 2018-10-20
Board member detail: Missing map geojson


reginafcompton:
On the person detail, the maps do not render, for at least two members, e.g., 

https://boardagendas.metro.net/person/james-butts-dc85ae14f195/
https://boardagendas.metro.net/person/sheila-kuehl-9c9abd7a6ff8/

This seems to be because the [`person.latest_council_membership.post.shape`](https://github.com/datamade/django-councilmatic/blob/master/councilmatic_core/views.py#L393) evaluates to False, i.e., the membership post does not have a `shape`.

shrayshray:
It looks like Sheila Kuehl's area is rendering now on her page. James Butts' is still not. It's high priority to have his map area render, as he is Chair of the Board of Directors this year. Is there anything you need from Metro's side to address this?

hancush:
The work to resolve this has been done in the upgrade: #518.

hancush:
This change is now available on the test site! https://lametro-upgrade.datamade.us

PR #517, opened 2019-12-09
Repair map display on person detail pages

hancush:
## Overview

This PR fixes the display of maps on the person detail pages. Sorry to hop out of my lane, @fgregg, but I was looking #361 (specifically whether it was addressed in the upgrade) and arrived at this change while debugging. Thought I'd submit a PR to save you the trouble!

### Checklist

### Demo

N/A

### Notes

N/A

## Testing Instructions

 * If you haven't already, run a person scrape: `docker-compose run --rm scrapers pupa update lametro people --rpm=0`
 * Start the application: `docker-compose up app`
 * Navigate to the board members page and confirm the map still looks and behaves as expected: http://localhost:8011/board-members
 * Navigate to each board member detail page and confirm that all of them render a map, with the exception of John Bulinski, who represents the governor/state, i.e., does not have a shape.

Related to #361.

### DIFF
--- a/lametro/templates/lametro/person.html
+++ b/lametro/templates/lametro/person.html
@@ -6,7 +6,7 @@
 {% block title %}{{ person.name }}{% endblock %}
 
 {% block extra_css %}
-    {% if MAP_CONFIG %}
+    {% if map_geojson %}
         <link rel="stylesheet" href="{% static 'css/leaflet.css' %}" />
     {% endif %}
 {% endblock %}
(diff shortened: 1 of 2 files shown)


fgregg:
great catch. This doesn't quite work because it will created nested feature collections for the big map (i think it will render fine, but adds size to the that object).

I was thinking of a solution using the geojson serializer https://docs.djangoproject.com/en/3.0/ref/contrib/gis/serializers/

hancush:
Superseded by #518.

PR #518, opened 2019-12-10
person map using django serialize

### LINKED ISSUE #361: Board member detail: Missing map geojson

fgregg:
## Overview

Restores maps to individual person pages. closes #361

### Checklist

### Notes

@hancush [had a mixin strategy](https://github.com/datamade/la-metro-councilmatic/pull/517) that factored out some code and was used in both the person map and the board map. This code uses some builtin code from django, and scans a bit easier to me.

## Testing Instructions

 * If you haven't already, run a person scrape: `docker-compose run --rm scrapers pupa update lametro people --rpm=0`
 * Start the application: `docker-compose up app`
 * Navigate to the board members page and confirm the map still looks and behaves as expected: http://localhost:8011/board-members
 * Navigate to each board member detail page and confirm that all of them render a map, with the exception of John Bulinski, who represents the governor/state, i.e., does not have a shape.

Handles #361

### DIFF
--- a/lametro/views.py
+++ b/lametro/views.py
@@ -38,6 +38,7 @@ from django.views.generic import TemplateView
 from django.http import HttpResponse, HttpResponseRedirect, HttpResponsePermanentRedirect, HttpResponseNotFound
 from django.shortcuts import render_to_response, redirect
 from django.core import management
+from django.core.serializers import serialize
 from django.views.generic import View
 from django.views.decorators.csrf import csrf_exempt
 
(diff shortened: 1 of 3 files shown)


lametro/views.py
```diff
@@ -517,12 +518,21 @@ def dispatch(self, request, *args, **kwargs):
...
         context = super().get_context_data(**kwargs)
         person = context['person']
 
-        context['qualifying_post'] = person.current_council_seat.post.acting_label
+        council_post = person.current_council_seat.post
+
+        context['qualifying_post'] = council_post.acting_label
+
+        if council_post.shape:
+            context['map_geojson'] = serialize('geojson',
```
fgregg: i'm tempted to also try a serialize solution for the boardmember map, but it's not so straightforward. Might think a bit more about that,

fgregg:
cool. thanks! one thing that @jeancochrane and I talked about as a nice approach at some point would be to have the geojson be served be separate route. that way the html would be a lot smaller and the page could start rendering faster. put that in the parkinglot.

hancush:
And indeed a more compelling reason to get DRF involved.

Issue #508, opened 2019-11-18
Restore filtering of test events in 2.5 upgrade


fgregg:
Need to restore the filtering of test events in the 2.5 upgrade.

hancush:
It seems like this was already done in https://github.com/datamade/la-metro-councilmatic/commit/45256e00c0905899adfd54230bf5bc88c44eda58#diff-154e85485bdcc3826d208aa9dbc8d8e3R252, but we do need to restore bill filtering still!

PR #526, opened 2019-12-19
Restore bill filtering

hancush:
## Overview

This PR restores filtering private bills from display, namely:

https://github.com/datamade/la-metro-councilmatic/blob/9a8c0dd232aac453696281d47d256fcb5d98d0d4/lametro/models.py#L49-L55

It also re-enables the test for the manager, which should now be passing.

### Checklist

## Testing Instructions

 * Handled by re-enabled bill manager test.

Handles #508 (I think).

### DIFF
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -63,14 +63,21 @@ class LAMetroBillManager(models.Manager):
         when getting bill querysets. Otherwise restricted view bills
         may slip through the crevices of Councilmatic display logic.
         '''
-        filtered_qs = super().get_queryset()\
-                             .exclude(restrict_view=True)\
-                             .filter(Q(eventrelatedentity__agenda_item__event__status='passed') | \
-                                     Q(eventrelatedentity__agenda_item__event__status='cancelled') | \
-                                     Q(bill_type='Board Box'))\
-                             .distinct()
+        qs = super().get_queryset()
 
-        return filtered_qs
+        qs = qs.exclude(
+            restrict_view=True
+        ).annotate(board_box=Case(
+            When(bill__extras__local_classification='Board Box', then=True),
+            When(bill__classification__contains=['Board Box'], then=True),
+            default=False,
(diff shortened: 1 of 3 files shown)


lametro/models.py
```diff
@@ -63,9 +63,21 @@ def get_queryset(self):
         when getting bill querysets. Otherwise restricted view bills
         may slip through the crevices of Councilmatic display logic.
         '''
-        filtered_qs = super().get_queryset()
+        qs = super().get_queryset()
+
+        qs = qs.exclude(
+            restrict_view=True
+        ).annotate(board_box=Case(
+            When(bill__extras__local_classification='Board Box', then=True),
```
hancush: We used to be able to filter on `bill_type`, but that's been migrated to a [property of the Councilmatic bill object](https://github.com/datamade/django-councilmatic/blob/85e518ab6fb912a2bfa574e4e53abd82a2d1f412/councilmatic_core/models.py#L396).

You have to filter on properties in Python, which would require iterating over all bills. An operation that heavy didn't seem acceptable to me, so I wrote this annotation to approximate checking whether the property was Board Box. The weirdest thing is the check on `bill__classification`, which is an ArrayField. Docs for querying ArrayFields are here: https://docs.djangoproject.com/en/3.0/ref/contrib/postgres/fields/#contains.

tests/test_bills.py
```diff
@@ -86,7 +85,7 @@ def test_bill_manager(bill,
     '''
     bill_info = {
         'classification': [bill_type],
-        'extras': {'restrict_view': restrict_view},
+        'restrict_view': restrict_view,
```
hancush: This is [an attribute of the Councilmatic model now](https://github.com/datamade/django-councilmatic/blob/85e518ab6fb912a2bfa574e4e53abd82a2d1f412/councilmatic_core/models.py#L386).

Issue #520, opened 2019-12-17
Deploy to a staging site


fgregg:
In order to test and QA the changes to LA Metro and the related repositories, we should set up CD for a staging site. 

As the changes are far ranging, it would be best if it was separate from the existing staging infrastructure. It's also likely that we want to take advantage of the containerization we've already done either by using heroku or docker-compose on an EC2.

hancush:
This is done! https://la-metro-councilmatic-staging.herokuapp.com/

Except, there's no data.

hancush:
We're ready to stage this using our AWS infrastructure. The steps for that are as follows:

- [ ] Spin up a new Metro server
- [ ] Deploy the application to that server, so the database is created and migrated
- [ ] Configure the database to accept remote connections from the OCD server
- [ ] Update the scrapers to scrape into the new database
- [ ] Make any adjustments to the Metro ETL script
- [ ] Bake out event and person identifiers, and `scp` them to the new server
- [ ] Run the identifier migration management command

hancush:
For the record, the new EC2 instance is called `LA Metro Councilmatic - Upgrade`. Forest and I both have shell access, and Postgres has been configured to listen for and accept connections from the OCD database. The configs [live here](https://github.com/datamade/server-la-metro-councilmatic-upgrade).

The legacy Metro-related applications in CodeDeploy and Sentry are called `lametro-councilmatic`. I've created new ones called `la-metro-councilmatic` for consistency with this repository. 

When we're ready to turn off the lights on the legacy deployment, we should:

- [ ] Terminate the `metro-councilmatic` EC2 instance
- [ ] Delete the `lametro-councilmatic` CodeDeploy application
- [ ] Delete the `lametro-councilmatic` and `la-metro-councilmatic-staging` Sentry projects

hancush:
Also, the legacy deployment scripts don't use our zero-downtime strategy. I've adapted them to generate an SSL cert on deployment, however the domain is still hard-coded into a few places:

- The Nginx configuration
- The `settings_deployment.deployment_group.py` file
- The `after_install.sh` script

We'll want to update these to match our desired domains for staging and production when we do the cut over.

[...]

PR #529, opened 2019-12-26
Deploy upgrade to Heroku

hancush:
## Overview

This PR adds artifacts to deploy the upgraded application to Heroku.

Handles #520.


scripts/release.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash -x
+echo "Running release.sh"
+
+echo "Migrating database"
+python manage.py migrate --noinput
```
hancush: I moved this from the entry point to the release phase, because:

> If a release phase task fails, the new release is not deployed, leaving your current release unaffected.

https://devcenter.heroku.com/articles/release-phase

Migrations are explicitly named as good candidates for the release phase for this reason.

councilmatic/settings.py
```diff
@@ -12,29 +12,75 @@
...
 
+# Database and Solr connections
+DATABASES = {}
+
+DATABASES['default'] = dj_database_url.parse(
+    os.getenv('DATABASE_URL', 'postgis://postgres@postgres:5432/lametro'),
+    conn_max_age=600,
+    ssl_require=True if os.getenv('POSTGRES_REQUIRE_SSL') else False
+)
+DATABASES['default']['ENGINE'] = 'django.contrib.gis.db.backends.postgis'
```
hancush: Needed to set this manually. The Heroku `DATABASE_URL` always uses the plain Postgres engine, regardless of whether you've installed PostGIS.

Dockerfile
```diff
@@ -17,6 +17,6 @@ RUN pip install --upgrade pip && \
 
 COPY . /app
 
-RUN DATABASE_URL='' python manage.py collectstatic --noinput
+RUN mv scripts/etl.sh ../bin/metro_etl
```
hancush: This defines a `metro_etl` task that's executable in a one-off dyno like `heroku run -a la-metro-councilmatic-staging metro_etl`.

Issue #524, opened 2019-12-17
Preserve url and identifiers for people and events


fgregg:
We want to preserve the slugs of people and events. 

Some options to do this

1. extract the relevant data from our old OCD database and import it into the new councilmatic db.  Some care would need to be taken to appropriately trigger the [signal handlers](https://github.com/datamade/django-councilmatic/blob/2.5/councilmatic_core/signals/handlers.py) to create the slugs and other attribute we want. 

2. we write a script that assumed an existing councilmatic 2.5 database and, using data derived from the old setup updated the slugs.

PR #530, opened 2020-01-14
Add management command to migrate slugs from legacy Councilmatic database

hancush:
## Overview

This PR adds a Makefile to extract slugs and uniquely identifying fields from the legacy Councilmatic database, and defines a management command to join those slugs with people and events in an existing 2.5 database. The full pipeline can be run from any computer with shell access to the Metro server. This means, when we're ready to deploy the upgrade, we can shell into the server where it's deployed, forwarding our SSH credentials (`ssh -A ...`), to run this migration, saving us the trouble of configuring the legacy database to accept remote connections, etc.

### Checklist

### Notes

- Luckily, there are < 50 people in the Councilmatic database, and they all have unique names, so I was able to match them using name alone.
- I've included logic to ensure that all objects in the upgraded database are updated, but I don't include logic to ensure all objects from the legacy database were matched to an object in the upgraded database. I did this so I could run the migration on a subset of data on my local machine. It also seems more meaningful that all the new data is updated – there could be stale records in the legacy database that don't need to be migrated, e.g. Happy to hear a counter argument here, though. 
- I didn't fully containerize this pipeline because our application image does not include Make, and I don't think a one-time (or limited-use) operation is worth introducing another dependency and increasing the size of the image. Again, happy for alternative thoughts.

## Testing Instructions

_Assuming you are on a computer with shell access to the Metro server:_

 * If you haven't already, import some data into your local database: `docker-compose run --rm scrapers`
 * Retrieve the existing lists of person and event slugs from the Metro server: `make`
 * Update the slugs in your local database: `docker-compose run --rm app python manage.py migrate_identifiers`

Handles #524.


fgregg:
Looks good @hancush. 

Since you will usually end up iterating through all the people and events anyway, I think I might just build a dictionary right away, instead of building it as you go in `councilmatic_entity` would be make the code a bit more straightforward.

However, since this is not going to be a regular script that we need to deal with, I'm happy with your structure as is.

fgregg:
i liked your Keyed models, that's a nice idea!

PR #532, opened 2020-01-15
Merge master into upgrade and refactor autocomplete backend

hancush:
## Overview

This PR brings changes from master into the upgrade. It also refactors the backend approach for autocomplete to work with the upgrade model changes (namely, there is no longer a separate Councilmatic Subject). The refactor is contained to [this commit range](https://github.com/datamade/la-metro-councilmatic/pull/532/files/59c3d1f08d1f5c4f81769cc6a08397f65a55197b..b8b9fe6b93c99a3bd03f338c0d1422d8e803b7c2). I would recommend reviewing that diff alone, since the rest has already been reviewed.

### Checklist

### Demo

N/A

### Notes

N/A

## Testing Instructions

 * `blackbox_cat configs/settings_deployment.staging.py`, then copy the `SMART_LOGIC_*` settings into your local deployment settings.
* If you don't have a local database, run a scrape: `docker-compose run --rm scrapers`.
* Run the management command to sync subjects with their GUIDs: `docker-compose run --rm app python manage.py refresh_guid`.
* Run the application and navigate to the homepage. Type `Metro Red Line` into the search box and confirm you are shown suggestions. Click `Metro Red Line` in the drop down and confirm you're redirected to the search results page and the results meet your expectations.


lametro/models.py
```diff
@@ -687,3 +684,22 @@ def related_files(self):
...
+    guid = models.CharField(max_length=256, blank=True, null=True)
+
+    def __str__(self):
+        if self.guid is not None:
+            return '{0} ({1})'.format(self.name, self.guid)
+
+        else:
+            return self.name
+
+    @property
```
fgregg: isn't a blocker, but would we want to have the relation between the subject and bills be a foreign key instead of a dynamic property?

tests/test_fetch_topic.py
```diff
@@ -0,0 +1,26 @@
+import pytest
```
fgregg: you don't seem to be using most of the imports you have here.

lametro/models.py
hancush: Yes, a `ManyToMany` field would be more appropriate. That said, this functions as a lookup table between subject and GUID. We don't have any need to get at bills this way right now, so I removed the property. We can add and populate the relationship in the future, if such a need arises.

tests/test_fetch_topic.py
hancush: Removed.

lametro/management/commands/refresh_guid.py
```diff
@@ -14,39 +17,41 @@ def handle(self, *args, **options):
         scraper.retry_attempts = 0
         scraper.requests_per_minute = 0
 
-        all_topics = scraper.topics()
+        current_topics = set(chain(*Bill.objects.values_list('subject', flat=True)))
```
fgregg: how often are we going to to run this, it seems like this could be pretty slow, and could be moved into SQL.

lametro/templatetags/lametro_extras.py
```diff
@@ -154,5 +152,49 @@ def get_highlighted_attachment_text(context, id):
...
+    return False
+
+@register.filter
+def sort_by_index(array, index):
+    '''
+    Sort a list of tuples by an item in that tuple, e.g., sort a facet listing
+    like [(topic, count), (topic, count)].
+
+    Django's dictsort template filter supports this natively >= 1.10, but for
+    now, we're still on Django < 1.10.
```
fgregg: we are not django 1.10 anymore. 

this is not a blocker. just observing.

lametro/management/commands/refresh_guid.py
hancush: It gets run once a day, and we haven't had any issues so far with execution time. Run locally, it only takes a few seconds.

PR #618, opened 2020-05-18
Expose staging Solr instance on 0.0.0.0

jeancochrane:
## Overview

This PR exposes the staging Solr instances on 0.0.0.0 so that it can be accessed externally.

### Checklist

### Notes

- In order to test this PR I made the following changes manually to staging:
    - Exposed ports 8986 and 8985 in the EC2 firewall to the LA Metro dashboard instance (note that the ports are not publicly accessible)
    - Updated `docker-compose.deployment.yml` in the `lametro-staging` directory according to this PR
    - Updated `settings_deployment.py` in the `lametro-staging` directory according to this PR
    - Restarted Solr and confirmed I could continue to connect to the staging instance
    - Confirmed I could connect to the staging instance from the LA Metro dashboard instance

## Testing Instructions

 * Confirm that searches continue to behave as normal on the staging server given that the config changes have already been applied manually


jeancochrane:
Yeah, my plan was to manually restart the production Solr instance when we next deploy to production. As far as I understand it, deploying this to staging won't affect the production configs because staging and production are two separate directories on the same server. Does that sound right to you?

hancush:
@jeancochrane I'd like to be able to deploy to production without having to remember to restart the Solr instance. Can we roll out changes to production configs for now and reapply them when we are ready for launch?

jeancochrane:
@hancush Yep, I can go ahead and make that change now if it's OK with you. Are you OK letting LA Metro know that we'll need a brief window of downtime? I don't think we need to plan for more than 15 minutes.

hancush:
@jeancochrane Sorry, I meant can we roll changes _out of this PR_, and apply them to production when we are ready to launch?

jeancochrane:
Ah, yes! Sorry, my misunderstanding. I went ahead and reset the production instance to use the loopback interface as its host in 5f53c669dda6abc7644fd5b4486e1e3337ab0066.

Issue #562, opened 2020-03-02
Do we need to retain the solr_scripts directory?


hancush:
It seems like use the Solr config in `solr_configs` for our search indexes. (That is, it's the one in the volume we mount onto our Solr containers.) As best I can tell, we don't use the one on the `solr_scripts` directory anywhere. Is it safe to remove it?

Issue #463, opened 2019-06-27
investigate enriched faceting from smartlogic tags


hancush:
metro has created several tag groupings that would be useful to include as facets. these include:

- lines/ways, e.g., "bus line", "rail line"
- phase
- project type
- location
- motion by
- topics/subjects

these will all be imported into legistar as matterindexes. there are some examples in the test api now.

hancush:
Hi, @shrayshray! Is there any way to add information about what grouping an index belongs to the indexes listing in the Legistar API? Otherwise, I bet we can grab it from the Semaphore API. Is a grouping equivalent to a class in Semaphore?

In any case, the way topics are represented in our data is changing in the upgrade, so we will probably want to wait on implementation here until those changes come in.

hancush:
Hey, @shrayshray – want to reup the question here: Is there any way to add what grouping a matter index belongs to, to the Legistar API?

shrayshray:
@hancush These facets are based on concept Class.

i.	Lines / Ways - Class: “Transportation Method” and all of its Subclasses & “System: Legistar” equals true
ii.	Phases - Class: “Transportation Phase” and all of its Subclasses & “System: Legistar” equals true
iii.	Projects - Class: “Project” and all of its Subclasses & “System: Legistar” equals true
iv.	Locations - Class: “Location” and all of its Subclasses & “System: Legistar” equals true
v.	Significant Dates - Class: “Significant Dates” and all of its Subclasses & “System: Legistar” equals true
vi.	Motion by -  Class: “Board Member” and all of its Subclasses & “System: Legistar” equals true
vii.	Topics/Subjects (REPLACES: current “Filter by Topic”) – Any term/tag not covered above & “System: Legistar” equals true

hancush:
@shrayshray I know that it's _possible_ for a concept to have more than one class, but in practice, will any of the concepts in the taxonomy you describe ever belong to more than one of the classes we're using for faceting? For example, can a concept be both a line and a location?

shrayshray:
@hancush it's highly unlikely we'll have any concepts with two classes which are also two different facets. I can't think of any existing examples, or any cases where this might occur. On any concepts I've seen with two classes assigned, only one of them is a facet.

hancush:
@shrayshray Very helpful! Thank you!

hancush:
@shrayshray @camachoo New Lines / Ways and Phase facets are live on the staging site! (As a bonus, I've also fixed #592.) 

[...]

Issue #462, opened 2019-06-27
allow for multiple values of report type


hancush:
metro bills can often include many kinds of report. as such, metro has defined report types as smartlogic tags, so that a report can be classified as multiple types.

- [ ] source a canonical list of report type tags from metro.
- [ ] define some property on the bill model to return the appropriate array of tags.
- [ ] investigate whether haystack index needs to be updated to accommodate the change in model. see: [faceting docs](https://django-haystack.readthedocs.io/en/v2.4.1/faceting.html#switch-to-the-facetedsearchview-and-facetedsearchform).

hancush:
This feels related to #463... @shrayshray, how is a term classified as a report type in Semaphore? We can likely extract this information from their API, but again, we'd want to wait on implementation until the changes to topics in our end are made in the upgrade.

hancush:
Same here, @shrayshray – any guidance on how terms are classified as report types in Semaphore?

shrayshray:
@hancush the report types we want to use are in the Concept Scheme "Board Report Types".
Here's a note Omar and I dug up from way back about this facet:

> Report Types (REPLACES: current “Filter by Report Type”) - Concept Scheme: “Board Report Types” & “System: Legistar” equals true

hancush:
@shrayshray I pulled concepts in the given concept scheme using this SES API query: https://cloud.smartlogic.com/svc/0dcee7c7-1667-4164-81e5-c16e46f2f74c/ses/CombinedModel/concepts/Board Report Types.json?FILTER=AT=System: Legistar

Here are the results: https://docs.google.com/spreadsheets/d/1d1qiYxcq9UunMAwG9s1Ghb2AIszIUpCvq3QhwbPCms8/edit?usp=sharing

The top of the list looks good, but some of the terms further down the list don't seem to belong. Any advice for pulling the correct list of terms from the API?

If you want to use a fixed set of board report types, instead, can you provide a list? Many thanks!

hancush:
Update: Connected with Steve, who gave me some super helpful guidance. Have enough to move forward here, hope to stage this change in the first half of next week!

PR #622, opened 2020-05-27
Add enriched faceting (and supporting infrastructure)

hancush:
## Overview

This PR:

- Adds a classification to `LAMetroSubject`
- Populates subject classifications during the overnight sync with SmartLogic
- Updates the bill type facet to accommodate multiple types (#462), draw from SmartLogic classification
- Adds additional facets for Lines / Ways and Phase (#463)
- Removes unused Solr schema (#562)

### Checklist

### Demo

Optional. Screenshots, `curl` examples, etc.

### Notes

- [x] Should tags beneath a search result include tags in all facets, or just the topics facet? (Depending on the answer, make the appropriate refactor to show and highlight tags appropriately.)
- [x] Verify that the `Board Report Type` class has been pushed to the production SmartLogic instance prior to launch. (It was added after initial attempts to query the Board Report Type concept schema returned unexpected results in https://github.com/datamade/la-metro-councilmatic/issues/462#issuecomment-632785657.)


hancush:
@shrayshray Good morning! I'm making good progress here, and I wanted to check with you and @camachoo about something. 

Right now, each search results include all tags for the given board report.

Screen Shot 2020-05-28 at 9 26 06 AM

With the enriched faceting, some of these tags are broken out into their own categories, e.g., Metro Gold Line, the transportation phases, etc. **Should the categorical tags be included with the search results,** or only the tags that don't belong to a category, i.e., "Topics / Subjects" (as defined here https://github.com/datamade/la-metro-councilmatic/issues/463#issuecomment-628901699)?

Let me know what you think!

shrayshray:
@hancush hmm ... I like the direction you're thinking for this. It could be helpful to have tags categorized by facet, especially considering the only real order we're providing at the moment is alphabetical. My sense though is that breaking them out by facet here might be a little much for the limited space. What do you think about leaving the tags here on the search results as is, but break them out by facet on the individual report pages?

hancush:
Good thought, @shrayshray. Let me see what I can put together.

PR #637, opened 2020-08-03
Handle private bills during event packet build

hancush:
## Overview

Sometimes event agenda items can be associated with matters that are private (or otherwise do not meet criteria for display on the board agendas site). When this happens, it causes event packet building to fail! See: https://github.com/datamade/la-metro-dashboard/issues/46. This PR handles that exception. 

### Checklist

### Notes

We run the risk of this happening everywhere we access bills via a foreign key relationship that has been overridden to use `LAMetroBill`. (IIRC, there are 3-4 places, including this one.) I'll open a separate issue to audit that, since this patch is time sensitive.

FWIW, I don't see a solution to handle this further upstream, e.g., [in `django-proxy-overrides`](https://github.com/datamade/django-proxy-overrides/blob/master/proxy_overrides/base.py), though I'd be happy to discuss, if one seems obvious to you, @fgregg!

## Testing Instructions

 * Merge this PR, and wait for the latest `staging` tag to build in Dockerhub.
 * Trigger the `hourly_processing` DAG on the dashboard and confirm the `compile_pdfs` task succeeds.

Handles https://github.com/datamade/la-metro-dashboard/issues/46

### DIFF
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -740,8 +740,15 @@ class EventPacket(Packet):
 
         for item in agenda_items:
             for entity in item.related_entities.filter(bill__isnull=False):
-                bill_packet = BillPacket(bill=entity.bill)
-                related.extend(bill_packet.related_files)
+                try:
+                    related_bill = entity.bill
+                except LAMetroBill.DoesNotExist:
+                    # If this exception occurs, the related bill is private.
+                    # Skip it.
+                    continue
+                else:
+                    bill_packet = BillPacket(bill=related_bill)
+                    related.extend(bill_packet.related_files)
 
         return related
 
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -744,12 +744,12 @@ class EventPacket(Packet):
                     related_bill = entity.bill
                 except LAMetroBill.DoesNotExist:
                     # We configure event agenda items to return LAMetroBill
-                    # objects. Sometimes, agendas items concern bills that do
+                    # objects. Sometimes, agenda items concern bills that do
                     # not meet criteria for display, e.g., the bill is private
                     # or it does not appear on a published agenda. (See the
                     # LAMetroBill manager for an exhaustive list of display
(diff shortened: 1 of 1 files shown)


lametro/models.py
```diff
@@ -740,8 +740,15 @@ def related_files(self):
 
         for item in agenda_items:
             for entity in item.related_entities.filter(bill__isnull=False):
-                bill_packet = BillPacket(bill=entity.bill)
-                related.extend(bill_packet.related_files)
+                try:
+                    related_bill = entity.bill
+                except LAMetroBill.DoesNotExist:
+                    # If this exception occurs, the related bill is private.
```
fgregg, line 746: let's give a little more context on why it might happen that an item has a handle on a bill but, we can't get access it.

fgregg:
if we are going to scrape private & test bills, then this solution makes sense.

iirc, the reason that we scrape private bills is so that sometimes a bill was public and we need to make it private so we need it to update that status.

that makes me think about addressing the deletion problem: https://github.com/opencivicdata/pupa/issues/295

if that was addressed, then we could not scrape private bills and get the behavior we want.

we might still need the code you are writing here, because we would have test bills which we also no not want to show.

lametro/models.py
fgregg: the could also happen if it was test item too right?

lametro/models.py
```diff
@@ -740,8 +740,15 @@ def related_files(self):
...
+                    # Skip it.
```
fgregg, line 747: could this also happen for test bills?

hancush: Good catch – yes! It can happen if the bill does not meet the criteria for display, only one of which is because it's private:

https://github.com/datamade/la-metro-councilmatic/blob/763d58553f4dd2d2041d8abd6e9e12a6fa25346d/lametro/models.py#L51-L58

I'll clarify this comment.

Issue #638, opened 2020-08-03
Board Report: Incorrect dates in History


shrayshray:
On [2015-0957](https://boardagendas.metro.net/board-report/2015-0957/), in the "History" section below the PDF, the the dates shown for the meetings are off by one day. 
The Regular Board meeting was on 7/23/15, but shows as 7/22/15, and the Construction Committee was on 7/16/15 but shows as 7/15/15.
![image](https://user-images.githubusercontent.com/28063988/89229650-aea3c300-d596-11ea-87a4-6725d5ff539c.png)

hancush:
Hm, curious! Is this happening for all board reports, or just this one, @shrayshray?

shrayshray:
@hancush It's happening on all reports.

hancush:
Did some spelunking!

The underlying problem is that actions have a date (string) that we cast to date time, which is stored in UTC (midnight on the given date). When we go to display the date time in the template, because we have timezone support enabled, the UTC date time is localized. California is several hours behind UTC, so the date is displayed as the day _before_ the correct date.

Dates for bill actions (and memberships) do not come to us with times, so to me, it seems more appropriate to cast them to dates to avoid this issue. That looks like this: https://github.com/datamade/django-councilmatic/compare/2.5...hcg/dates-to-dates.

Unfortunately, I think this is going to be a bit of an annoying change to make because we do a lot of date time comparisons in the app, so those comparisons are going to need to be updated, if we proceed. With that in mind, I want to sanity check this change with @fgregg, before I go in and start updating all these references.

hancush:
Alternatively, we could use [the `utc` template tag](https://docs.djangoproject.com/en/3.1/topics/i18n/timezones/#utc) when we display the dates. Would still require what I'd guess are a fair few updates, and we'd have to remember to use it when we display dates from bill actions and memberships, but may ultimately be less hairy.

hancush:
Going to move forward with the original idea: Cast date strings to dates, and coerce them to date times only where necessary.

hancush:
@shrayshray Just deployed a patch to address this to the staging site. Looks like action dates for the board report you linked are behaving as expected: https://lametro-upgrade.datamade.us/board-report/2015-0957/. Would you mind helping us spot check action dates for a few other board reports, and let me know if it looks good?

shrayshray:
@hancush spot check checks out - looks great!

hancush:
Thank you, @shrayshray. We're live!

PR #651, opened 2020-09-15
Upgrade to django-councilmatic 2.6.0

hancush:
## Overview

This PR upgrades `django-councilmatic` to capture https://github.com/datamade/django-councilmatic/pull/272, which handles #638. It also updates one test affected by that change.

### Checklist

## Testing Instructions

N/A – just an FYI! (I'll cruise the staging site after this deploys to double check that the fix addresses #638 and that nothing is broken.)

Handles #638.


fatima3558 (commented):
makes sense! 👍

Issue #652, opened 2020-09-17
Add new Public Comment info


shrayshray:
Please add the following info to the Public Comment sections which show during the meeting, and hide before and after. Omar is confirming whether the eComment option will continue to be available and will report back.

Participant Instructions
The conference begins at 10:00 AM Pacific Time on September 24, 2020; you may join the conference 5 minutes prior.
Dial-in: 888-251-2949 or 215-861-0694
English Access Code: 8231160#
Spanish Access Code: 4544724#
Need an international dial-in number?
Need assistance with your audio? Please dial 888-796-6118.

hancush:
Hey, @shrayshray! Obviously the date and time in this message will change depending on the event. Are the dial-in and access codes consistent, or will they depend on the event, as well?

shrayshray:
@hancush my understanding based on Omar's email is the dial-in and access codes will be consistent.

hancush:
@camachoo, is there a generic link I should use for the international dial-in number? The one in your original email does not work when I add it to the site: https://www.teleconference.att.com/servlet/glbAccess?process=1&accessNumber=8882512949&accessCode=8231160&data=02|01|CamachoO@metro.net|a3703f3f0bb242d9b6fb08d85b25ec57|ab57129bdbfd4cacaa77fc74c40364af|0|0|637359565723000140&sdata=ft51LLU3MCgbfVr/ajVShtcy/5D3p1/V3+wV0ktFyR8=&reserved=0

hancush:
Will wait for guidance from Omar on text changes. In the meantime, I've added the dial-in info to the block that shows during the meeting.

Here's what event headings will look like at each stage.

### Before meeting, no agenda:

![Screen Shot 2020-09-17 at 2 39 21 PM](https://user-images.githubusercontent.com/12176173/93519320-ba073f80-f8f3-11ea-8ac9-15d8a1a14f81.png)

### Before meeting, agenda posted:

![Screen Shot 2020-09-17 at 2 38 05 PM](https://user-images.githubusercontent.com/12176173/93519328-bb386c80-f8f3-11ea-885d-35c6cb29793f.png)

### During meeting:

![Screen Shot 2020-09-17 at 2 43 21 PM](https://user-images.githubusercontent.com/12176173/93520266-297d2f00-f8f4-11ea-87be-f22b9137bece.png)

### After meeting:

[...]

PR #654, opened 2020-09-29
Handle daylight savings time in upcoming event check

hancush:
## Overview

This PR uses `localize()` instead of passing a `pytz` timezone to the datetime constructor. This is needed because:

> Unfortunately using the tzinfo argument of the standard datetime constructors ‘’does not work’’ with pytz for many timezones.

Specifically is, it doesn't work with daylight time. It also adds `freezegun` so we can mock datetime to test that `is_upcoming` works in a variety of cases.

### Checklist

## Testing Instructions

 * There's not a great way to test this locally! The automated tests check the logic. Think we'll have to rely on Metro to QA this once it's deployed.

Handles #652

### DIFF
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -409,6 +409,23 @@ class LAMetroEvent(Event, LiveMediaMixin, SourcesMixin):
         return timezone.now() + relativedelta(**kwargs)
 
 
+    @staticmethod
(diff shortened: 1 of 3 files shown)


fatima3558 (approved):
looks good!

PR #656, opened 2020-10-09
Include appearing on agenda in board report history, update columns

hancush:
## Overview

This PR:

- Adds an `actions_and_agendas` property to the `LAMetroBill` model that includes information about events and agendas, their times, descriptions, and related organizations
- updates the `legislation.html` template (for the `LAMetroBill` model) with the data from this property

This PR pulls the relevant commits from #450, which it supersedes. (Let's squash and merge when it comes in, to get rid of the mysterious duplicate commits.)

### Checklist

## Testing Instructions

* Navigate to a piece of legislation
* Hard refresh the page
* Confirm link goes to appropriate event agenda
* Confirm event status is correct
* Repeat with a few different pages

Handles #348


lametro/models.py
```diff
@@ -153,6 +153,56 @@ def board_report(self):
...
+                'description': action.description,
+                'event': event,
+                'organization': action.organization
+            }
+            
+            data.append(action_dict)
+
+        for event in events:
+            # Use a description of "SCHEDULED"
+            related_entity = self.eventrelatedentity_set.get(agenda_item__event=event)
```
hancush, line 192: The event related entity associates the bill with an event agenda. You're right (and it's a bit confusing!) that `organization` is a possible related entity, but we don't define it, i.e., this is coming back as null in the template. Instead, you can get at the organization via `event.participants` (see [`EventParticipant`](https://github.com/opencivicdata/python-opencivicdata/blob/ad30a91a243fd8784955b2440e79f23ba6fe7512/opencivicdata/legislative/models/event.py#L143)).

lametro/templates/lametro/legislation.html
```diff
@@ -24,6 +24,7 @@ <h1>
                     {% if legislation.inferred_status %}
                         {{ legislation.inferred_status | inferred_status_label | safe }}
                     {% endif %}
+                <a href="rss/" title="RSS feed for Legislation {{legislation.friendly_name}}" target="_blank"><i class="fa fa-rss-square" aria-hidden="true"></i></a>
```
hancush, line 27: Oops, this got removed in https://github.com/datamade/la-metro-councilmatic/pull/645 and I must have missed the change when I prepped this branch for you initially. You can remove this! 

```suggestion
```

lametro/templates/lametro/legislation.html
```diff
@@ -97,20 +98,19 @@ <h3 class="no-pad-bottom"><i class='fa fa-fw fa-list-ul'></i> History</h3>
...
-                                        {% if action.related_organization %}
-                                            to
-                                            {{action.related_organization.link_html|safe}}
-                                        {% endif %}
+                                        <span class='text-default'>{{action.description}}</span>
                                     </td>
                                     <td class="small">
-                                        {{action.organization.link_html|safe}}
+                                            {{ action.organization.link_html|safe }}
+                                        {% if action.description == 'CANCELLED' %}
```
hancush, line 111: `CANCELLED` is an event status, so you should check that, rather than the action description, to determine whether to include the `<strike>` element and label. It also looks like you're missing the opening strike element. You're looking for something like:

```html
{% if the meeting is cancelled %}<strike>{% endif %}
Organization link
{% if the meeting is cancelled %}
</strike>
<!-- label -->
{% endif %}
```

See Jasmine's commit: https://github.com/datamade/la-metro-councilmatic/pull/450/commits/7de05bfbec2588432d8b2cdcd906a459b1cdad09#diff-73a27e2aba0b040dc38e92e1f11e8b87R114

lametro/views.py
```diff
@@ -99,6 +99,8 @@ def get_context_data(self, **kwargs):
 
         context['related_bills'] = related_bills
 
+        context['events'] = bill.actions_and_agendas
```
hancush, line 102: Would you mind calling this context variable `actions`, by extension of https://github.com/datamade/la-metro-councilmatic/pull/450#discussion_r498922714?

lametro/templates/lametro/legislation.html
```diff
@@ -97,20 +98,19 @@ <h3 class="no-pad-bottom"><i class='fa fa-fw fa-list-ul'></i> History</h3>
...
-                                        <span class='text-{{action.label}}'>{{action.description | remove_action_subj}}</span>
-                                        {% if action.related_organization %}
-                                            to
-                                            {{action.related_organization.link_html|safe}}
-                                        {% endif %}
+                                        <span class='text-default'>{{action.description}}</span>
                                     </td>
                                     <td class="small">
-                                        {{action.organization.link_html|safe}}
+                                            {{ action.organization.link_html|safe }}
```
hancush, line 110: Shelly asked that we link to the meeting, rather than the organization. Luckily, people, organizations, and events all have [a `link_html` property](https://github.com/datamade/django-councilmatic/blob/ce203941540ccf70d4671823b49b254c3b4d35b4/councilmatic_core/models.py#L348), so you just need to change the reference here.

```suggestion
                                            {{ action.event.link_html|safe }}
```

lametro/templates/lametro/legislation.html
```diff
@@ -97,20 +98,19 @@ <h3 class="no-pad-bottom"><i class='fa fa-fw fa-list-ul'></i> History</h3>
                             </tr>
                         </thead>
                         <tbody>
-                            {% for action in actions %}
+                            {% for action in events %}
                                 <tr>
                                     <td class='nowrap text-muted small'>
-                                        <span datetime='{{action.date_dt | format_date_sort }}'>{{action.date_dt|date:'n/d/y'}}</span>
+                                        <span datetime='{{action.date}}'>{{action.date}}</span>
```
hancush, line 104: Looks like you dropped the formatting for display here.

```suggestion
                                        <span datetime='{{action.date}}'>{{action.date|date:'n/d/y'}}</span>
```

hancush (commented):
This is looking really slick, @fatima3558! I noticed one weirdness in the ORM query, but otherwise, the remaining changes are small. 

One question: Do you have a bill in your local DB that you can test this on? I've been looking at http://localhost:8001/board-report/2016-0023/, which appears on an agenda and also has an action, to see whether things are rendering as expected.

fatima3558:
yes, I've been using http://localhost:8001/board-report/2020-0420/, which we came up with as a good option to check the changes while we were pairing.

lametro/templates/lametro/legislation.html
fatima3558: @hancush would it be appropriate to change the name of the column here as well?

hancush: Yes, great call (and good catch)! One of the original requests was to change the column heading from Committee to Meeting: https://github.com/datamade/la-metro-councilmatic/issues/348#issuecomment-668258972

fatima3558 (approved):

lametro/models.py
```diff
@@ -153,6 +153,56 @@ def board_report(self):
...
+                'description': "SCHEDULED",
+                'event': event,
+                'organization': org
+            }
+
+            data.append(event_dict)
+
+        sorted_data = sorted(data, key=lambda x: x['date'])
+
+        return data
```
hancush, line 204: 😅

```suggestion
        return sorted_data
```

lametro/models.py
```diff
@@ -153,6 +153,56 @@ def board_report(self):
...
+            'organization': <Organization>
+        }
+        '''
+        actions = self.actions.all()
+        events = LAMetroEvent.objects.filter(agenda__related_entities__bill=self)
+
+        data = []
+
+        for action in actions:
+            event = LAMetroEvent.objects.filter(\
```
hancush, line 175: `filter()` returns a queryset – we want to attach the resulting event instance, not the queryset, to the action dict. I would expect this to only ever return one result. How about using `get()` here?

```suggestion
            event = LAMetroEvent.objects.get(\
```

hancush (commented):
Nice work, @fatima3558! Looked like the event name wasn't showing up properly for actions, and the history appeared out of order, on the test page.

Screen Shot 2020-10-14 at 9 36 49 AM

Suggested a couple of small changes inline that I think should resolve the issues. Can you make them, view the page, and confirm the history is in order and every item has a link? Then I'd be happy to bring this in.

lametro/models.py
fatima3558: ahhhh dang 😅  thanks for catching this!

fatima3558:
made the changes you suggested, confirmed that each action has a meeting link, and confirmed that the actions are in chronological order

hancush:
Woo, @fatima3558! Actions are now in date order:

Screen Shot 2020-10-14 at 10 30 38 AM

However, there's a little snafu when there are two items on the same date – conceptually, it would make most sense for "Scheduled" to appear _before_ any actions on the same date. Can you implement a solution such that the items are ordered by date *and* description? If there are multiple actions on a given date, "Scheduled" should be ordered first, followed by all subsequent actions in alphabetical order.

When applied to the example bill, the order would look like:

1. 8/19 Scheduled
2. 8/19 Recommended for approval
3. 8/27 Scheduled
4. 8/27 Approved

Let me know if you have any questions!

PR #661, opened 2020-10-15
Include appearing on agenda in board report history, update columns, sort scheduled first

fatima3558:
## Overview
This PR
* Adds an actions_and_agendas property to the LAMetroBill model that includes information about events and agendas, their times, descriptions, and related organizations
* Updates the legislation.html template (for the LAMetroBill model) with the data from this property

### Checklist

## Testing Instructions

* Navigate to a piece of legislation
* Hard refresh the page
* Confirm link goes to appropriate event agenda
* Confirm event status is correct
* Repeat with a few different pages

Handles #348

### DIFF
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -212,15 +212,10 @@ class LAMetroBill(Bill, SourcesMixin):
 
             data.append(event_dict)
 
-        scheduled_first = []
-        for d in data:
-            if d['description'] == 'SCHEDULED':
-                scheduled_first.append(d)
-        for d in data:
-            if d['description'] != 'SCHEDULED':
-                scheduled_first.append(d)
-
-        sorted_data = sorted(scheduled_first, key=lambda x: x['date'])
+        sorted_data = sorted(data, key=lambda x: (x['date'], x['description'].upper() if x['description'] == 'SCHEDULED' else x['description'].lower()))
+        
+        import pdb
+        pdb.set_trace()
 
         return sorted_data
 
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -212,7 +212,8 @@ class LAMetroBill(Bill, SourcesMixin):
 
             data.append(event_dict)
 
-        sorted_data = sorted(data, key=lambda x: (x['date'], x['description'].upper() if x['description'] == 'SCHEDULED' else x['description'].lower()))
+        sorted_data = sorted(data, key=lambda x:(x['date'],\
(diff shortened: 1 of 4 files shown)


lametro/models.py
```diff
@@ -166,6 +166,64 @@ def board_report(self):
...
+
+        scheduled_first = []
+        for d in data:
+            if d['description'] == 'SCHEDULED':
+                scheduled_first.append(d)
+        for d in data:
+            if d['description'] != 'SCHEDULED':
+                scheduled_first.append(d)
+
+        sorted_data = sorted(scheduled_first, key=lambda x: x['date'])
```
hancush, line 223: This is clever, but I worry that the first sort won't always persist when we apply the second. What if we used a tuple to sort the data, instead? Here's an example of how tuple sorts work in Python:

```python
>>> sorted([(1, 'A'), (1, 'a'), (0, 'A'), (0, 'AP')])
[(0, 'A'), (0, 'AP'), (1, 'A'), (1, 'a')]
```

You can see that Python sorts are [ascii-betical](https://en.wikipedia.org/w/index.php?title=ASCII&redirect=no#Order), i.e., uppercase letters come before lowercase letters. 

Would something like this do the trick?

```python
lambda x: (x['date'], x['description'].upper() if x['description'] == 'SCHEDULED' else x['description'].lower()
```

hancush (changes requested):

lametro/models.py
```diff
@@ -166,6 +166,60 @@ def board_report(self):
...
+            }
+
+            data.append(event_dict)
+
+        sorted_data = sorted(data, key=lambda x:(x['date'],\
+            x['description'].upper() if x['description'] == 'SCHEDULED' else x['description'].lower()))
+        
+        import pdb
+        pdb.set_trace()
+
```
hancush, line 220:
```suggestion
```

lametro/models.py
```diff
@@ -166,6 +166,60 @@ def board_report(self):
...
+            event_dict = {
+                'date': event.start_time.date(),
+                'description': 'SCHEDULED',
+                'event': event,
+                'organization': org
+            }
+
+            data.append(event_dict)
+
+        sorted_data = sorted(data, key=lambda x:(x['date'],\
```
hancush, line 215: Can you leave a brief comment about this sort, namely what the desired outcome is and what's going on with the `upper()` and `lower()`?

hancush (approved):
Nice! A couple of small things inline. Please add the requested comment, then you can go ahead and merge this in so we can test it out on staging.

Issue #487, opened 2019-09-04
Board Reports - Identify searches by File ID and show the matching report first in results


shrayshray:
To address searches by File ID not returning the matching report first in the results list (e.g., someone types [2018-0101](https://boardagendas.metro.net/search/?q=2018-0101&search-all=on) in the search bar, and that report shows up 2nd): 
When someone searches using the File ID, identify this structure (20##-####) and assume they're looking for the matching report - have the matching report show up first in search results.

hancush:
@shrayshray Bonus! I've implemented and deployed a fix for this, as well. The upshot is that I doubled the relevance score of the result with an identifier matching the query, so it's ordered first. This should work if the query contains multiple identifiers, too. We've done some successful testing on our end. Want to give it a go on yours?

shrayshray:
@hancush wow - this is so exciting! Single identifier searches are working consistently. Multiple identifier searches aren't exactly what I'd expect, but they're good enough I'm not super concerned. One example is a query for "2020-0101 2020-0202" returns:
1. 2020-0201
2. 2018-0201
3. 2017-0201
4. 2020-0101
Since the single identifier result is what we were aiming for, I don't consider the results for multiple returning in not quite the right order an issue.

hancush:
I will try one small change to get multiple IDs working. If that doesn't work, I'll go ahead and take the single ID search lvie and break ou t a separate issue for multi ID search for the ice b ox.

shrayshray:
I received a user report of this happening again - searching for a report by file ID doesn't always display the exact match first. I tested and also found a few cases. Some examples:
2022-0037
2021-0222
2020-0220
2019-0089
Could you look into what might be causing this?

hancush:
Sure, @shrayshray!

PR #667, opened 2020-10-29
Boost search results with identifier/s matching query

hancush:
## Overview

This PR subclasses `SolrSearchQuery` and extends the `run()` method to add additional search kwargs if the query contains an identifier. The effect of these extra kwargs doubles the score of results with an identifier matching the query, hopefully ensuring that they are returned first. 

N.b., Haystack `SearchQueryset` objects contain methods to boost terms in general, but I wanted a more targeted matching strategy (boost results where the `identifier` field, specifically, matches the query) because full text can contain several identifiers, e.g., agendas, which could yield unexpected results.

#### Further reading

- `SearchQuery` API: https://django-haystack.readthedocs.io/en/master/searchquery_api.html
- Blog on how to order matching results first: https://medium.com/@pablocastelnovo/if-they-match-i-want-them-to-be-always-first-boosting-documents-in-apache-solr-with-the-boost-362abd36476c

### Checklist

## Testing Instructions

To be done on the staging site:

 * Search `2018-0101` and confirm that the corresponding board report appears first in the results.

Handles #487

### DIFF
--- a/lametro/views.py
+++ b/lametro/views.py
@@ -617,7 +617,7 @@ class IdentifierBoostSearchQuery(SolrSearchQuery):
         if identifiers:
             kwargs.update({
                 'defType': 'edismax',
-                'bq': '+'.join('identifier:{}^2.0'.format(i) for i in identifiers),
+                'bq': '+'.join('identifier:"{}"^2.0'.format(i) for i in identifiers),
             })
 
         return super().run(spelling_query, **kwargs)
--- a/lametro/views.py
+++ b/lametro/views.py
@@ -13,6 +13,7 @@ from collections import namedtuple
 import os
 
 from haystack.backends import SQ
+from haystack.backends.solr_backend import SolrSearchQuery
 from haystack.inputs import Exact, Raw
 from haystack.query import SearchQuerySet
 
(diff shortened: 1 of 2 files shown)


lametro/views.py
```diff
@@ -635,7 +657,7 @@ def _topic_search(self, sqs):
         topic_filter = Q()
 
         for term in terms:
-            topic_filter |= Q(**{'subject__icontains': term})
+            topic_filter |= Q(subject__icontains=term)
```
hancush, line 660: This is just a syntax change – some functionality!

lametro/views.py
```diff
@@ -680,19 +699,21 @@ def build_form(self, form_kwargs={}):
...
-                              .facet('topics')\
-                              .facet('lines_and_ways')\
-                              .facet('phase')\
-                              .facet('project')\
-                              .facet('metro_location')\
-                              .facet('geo_admin_location')\
-                              .facet('motion_by')\
-                              .facet('significant_date')\
-                              .highlight(**{'hl.fl': 'text,attachment_text'})
+        sqs = SearchQuerySet(
```
hancush, line 702: I don't really like this indentation. Does it also bug you, @fatima3558?

fatima3558: I definitely like the new indentation better!

hancush:
Hmm, this isn't working as I want it to. Will investigate.

hancush:
Fixed in https://github.com/datamade/la-metro-councilmatic/pull/667/commits/8bcdef032a5f852f9f753b1de096efb1ee72860f! Needed to boost on exact phrase match.

fatima3558 (approved):
the search for an agenda's identifier does return that agenda as the first result 👍

Issue #650, opened 2020-09-14
Board Report Search: Some terms missing from filters


shrayshray:
Omar noticed some tags missing from the "Subjects" section of the filters on the [search page](https://boardagendas.metro.net/search/).
To recreate, search "Customer Experience Plan" by typing in the search bar and clicking on the term in auto-suggest, then execute search.
If you look at the first report, 2018-0668, it's tagged with "NextGen Bus Study" and "Metro Vision 2028 Plan". If you click to expand "Filter by Subject", these terms do not appear in the list. 
In what we've seen so far, all the missing terms have had Class="Plan", "Program" or "Policy" (Subclasses of "Initiatives" ... but there are some other terms from those 3 classes which do appear, such as "Long Range Transportation Plan" and "Metro Equity Platform".
Could you look into why some terms are left out of the filter list?

hancush:
@shrayshray How would you like us to prioritize this issue?

shrayshray:
@hancush do you think there's time this month to at least look into it?

hancush:
@shrayshray Sure!

hancush:
Hey, @shrayshray! This was a fun question. The answer is that our search engine is configured to return at most 100 tags, in order of most hits to least, in each category. For the example search, if you click on the NextGen Bus Study tag under the initial result, you can see there are only 30 results from the original query with that tag. The smallest number of hits on the top 100 tags for the original query is 41, so NextGen Bus Study didn't make the cutoff. 

It'd be negligible to change the max number of tag facets that come back if you'd like. I think whether we should do so really depends on what purpose you want the facets to serve. Right now, they highlight topics that most often appear with the user's search, i.e., topics they're most likely to care about. We could broaden that view, though I'd caution against showing too many, as they decrease in relevance the fewer hits they're associated with, and especially because there are other entry points to unlisted topics, such as with the search result, as you discovered.

What do you think?

hancush:
Some ideas:

- Add a new facet for Metro plans, policies, and programs
- Add this particular relationship to related terms (HCG will update this issue with SES call / response for how related terms are populated for the test query)

shrayshray:
Yes, let's add a new facet. Terms in the facet Plans, Programs and Policies will have a class of either:
Plan
Program
Policy

hancush:
@fatima3558 I'd like you to give this issue a shot! 

[...]

PR #668, opened 2020-11-09
Update solr index for plan_program_policy searching

fatima3558:
## Overview
This PR adds a `plan_program_policy` facet to the solr schema and brings that filter to the search results page.

### Checklist

### Notes

Noticed a small quirk with the way the text for the filter accordions fits in its space. Didn't notice this issue on the staging site, but making a note here to be safe.

Screen Shot 2020-11-13 at 10 16 41 AM

## Testing Instructions

* make sure your app is completely down with `docker-compose down`
* remove your solr volume with `docker volume rm la-metro-councilmatic_lametro-solr-data`
* update the solr index with `docker-compose run --rm app python manage.py refresh_guid`, then repopulate it with  `docker-compose run --rm app python manage.py update_index`
* run the app, and search something like `bus`
* verify that there is a new filter at the bottom of the filters list, that it expands into a list of relevant terms, and that clicking on one of those terms successfully filters the search results

Handles #650


hancush (changes requested):
Looking good, @fatima3558. Two things:

1. You've successfully added and populated the facet, but you haven't rendered the filter in the search template! Here's how to do that: https://github.com/datamade/la-metro-councilmatic/blob/c08d1508cb655a12e733586563bb6eb693beb1f4/lametro/templates/search/search.html#L95-L100
2. Can you fill out the PR description and include testing instructions? Many thanks! 🙏

hancush (approved):
It works!!!!!!!!!!! 🎊 Once this is deployed, would you like to pair on applying the change in schema to the Solr instance on staging? Know you're pretty busy today, so we can schedule a time during your next Metro block, if you like. (In any case, let's wait to merge until we're ready to make and test the change on staging.)

fatima3558:
@hancush awesome! 🎉 yes, it would be helpful to pair on changing the solr schema on the staging site during my next metro block.

Will hold off on merging until next week.

Issue #671, opened 2020-11-17
Update Travis CI build to align with server OS and Python version


hancush:
Our Travis builds are taking forever. I suspect this has something to do with using an old version of Python / other config issues mentioned here: 

```
root: deprecated key sudo (The key `sudo` has no effect anymore.)
root: missing dist, using the default xenial
root: missing os, using the default linux
```

Our development and deployment environments use Python 3.6. Let's square this dependency, and potentially look into other measures for speeding up the builds.

This may also be a good opportunity to transition this project over to GitHub Actions, which has replaced Travis [as our preferred CI tooling](https://github.com/datamade/how-to/blob/master/ci/github-actions.md). I'd do this using investment time, i.e., DataMade time.

hancush:
Done (on DataMade investment time) in #678.

hancush:
Let's double check that our Docker stuff uses the same version of things as the server.

PR #678, opened 2020-12-03
Replace Travis CI with GitHub Actions

hancush:
## Overview

This PR adds a GitHub Action with two jobs: one to use the bundled application image and Docker Compose setup to build the app and run tests, and another to deploy the app to CodeDeploy. 

Commits to `master` will deploy to staging, and commits to `deploy` will deploy to production. This represents a change in pattern from our current practice, which is to deploy to production on tags, but the CodeDeploy integration [does not currently support builds from tags](https://github.com/webfactory/create-aws-codedeploy-deployment/blob/0d7a684950dae16883a140dd950257958730e0b5/create-deployment.js#L10). 

I think it's a worthwhile tradeoff because it otherwise makes deployment super easy, and it would have the added benefit of bringing deployment patterns for CodeDeploy in line with deployment patterns for Heroku.

Handles #671, connects https://github.com/datamade/how-to/issues/36.

## Testing instructions

- Tested staging deployment from this branch in [this run](https://github.com/datamade/la-metro-councilmatic/runs/1494821427).


.github/workflows/main.yml
```diff
@@ -0,0 +1,35 @@
...
+  test:
+    name: Run tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: satackey/action-docker-layer-caching@v0.0.8
+      - run: docker-compose -f docker-compose.yml -f tests/docker-compose.yml run --rm app
+
+  deploy:
+    needs: test
```
fgregg, line 22: you could deploy on tags by adding this line:

`
    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
`

fgregg: got it, but there's not a mapping from tags to deployment groups

fgregg:
if you wanted faster builds, you could probably get that by not building the entire app in docker-compose, but just setting up postgis as "service" 

i did this with django-councilmatic. https://github.com/datamade/django-councilmatic/blob/2.5/.github/workflows/main.yml#L13-L25

otoh, i see some benefits from having tests run the same here as local

fgregg (approved):

PR #682, opened 2020-12-11
Use service approach rather than containerized tests

hancush:
## Description

After some trouble in the Docker layer caching arena, I decided to adopt the services approach @fgregg suggested. This PR implements that pattern for the testing action. Bonus: It reduced build speed from > 5 minutes to < 2 minutes.

Connects https://github.com/datamade/how-to/issues/36, #671.

## Testing instructions

- View the latest build and observe that it has succeeded.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       postgres:
-        image: postgis/postgis:10-2.4
+        image: postgis/postgis:10-2.5
         env:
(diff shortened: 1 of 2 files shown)


hancush:
@fgregg Requesting review as an FYI. This directly mirrors your approach in https://github.com/datamade/django-councilmatic/runs/1482966504.

fgregg:
was it faster?

hancush:
@fgregg Yes, down to 1m45s with service from 5m some odd seconds with containers.

Issue #537, opened 2020-01-16
Board Reports: Operations, Safety, and Customer Experience Committee missing from sidebar filter


shrayshray:
On the Board Reports search page, under "Filter by Meeting" in the left sidebar, Operations, Safety, and Customer Experience Committee is not listed:
![image](https://user-images.githubusercontent.com/28063988/72562475-61613d80-3860-11ea-93f0-db902349ba40.png)

shrayshray:
I also noticed the other meeting filters are not picking up reports for recent meetings. With nothing in the search bar and no previous query or filter applied, clicking any of the meeting names filters reports from that meeting body, 1/25/2018 and prior. 
Operations, Safety, and Customer Experience Committee has only been meeting since 7/19/2018.

hancush:
Hello again! 🙃 We populate the meetings facet by looking at matter histories, so if there are no histories associated with a board report, then it won't come back when you filter by meeting. I queried the database: It doesn't look like we have any histories associated with the Operations, Safety, and Customer Experience Committee, which would explain why it's not an available filter.

I'm seeing more recent reports [when I filter by Board of Directors](https://boardagendas.metro.net/search/?selected_facets=sponsorships_exact%3ABoard+of+Directors&sort_by=date&order_by=desc), so I'm not convinced there's a temporal element at play.

Does this explanation track with your expertise on the data?

Thinking more about this, since we know not every report has a history (#340), it doesn't seem appropriate to populate this facet using histories. Instead, I'd think something like sponsorships or committees associated with agendas that the report appears on would be more accurate. Thoughts?

shrayshray:
Yes, thank you! This all make sense, thanks for putting the pieces together. And yes, we should definitely be doing this based on the name of the committee the report is associated with at agenda posting time, rather than the matter history entry. 
Sponsors is not a required field and not always entered, but "In Control"/Controlling Body is - Would that work?

hancush:
Possibly! Is that the information in the `MatterBodyName` in the API?

shrayshray:
Yes, it is in MatterBodyName in the API.

hancush:
Update: Looks like this filter isn't appearing at all! I can look into this.

hancush:
[...]

PR #689, opened 2021-01-12
re-add meetings facet

fatima3558:
## Overview

This PR adds back the `Meetings` facet that got lost. 

### Checklist

### Notes

Based on the searches I tried, I didn't ever see any duplicate organizations, but I still went ahead and added logic to prevent duplication.

## Testing Instructions

 * Update haystack index using `docker-compose run --rm app python manage.py update_index`
 * Search something, like `red line`, and verify that the `Meetings` facet appears on the left side of the page with the other filters.
 * Expand the `Meetings` filter and make sure there are no duplicate organizations.
 * run tests using `docker-compose -f docker-compose.yml -f tests/docker-compose.yml run --rm app` to check new test, `test_sponsorships`, and to make sure change to organization factory in test fixtures doesn't break other tests

Handles #537

### DIFF
--- a/lametro/search_indexes.py
+++ b/lametro/search_indexes.py
@@ -29,7 +29,7 @@ class LAMetroBillIndex(BillIndex, indexes.Indexable):
         return None
 
     def prepare_sponsorships(self, obj):
-        return [action.organization for action in obj.actions.all()]
+        return [action.organization for action in obj.actions_and_agendas.all()]
 
     def prepare_sort_name(self, obj):
         full_text = obj.extras.get('plain_text')
--- a/lametro/search_indexes.py
+++ b/lametro/search_indexes.py
@@ -29,7 +29,8 @@ class LAMetroBillIndex(BillIndex, indexes.Indexable):
         return None
 
     def prepare_sponsorships(self, obj):
-        return [action.organization for action in obj.actions_and_agendas.all()]
+        orgs_list = [action.organization for action in obj.actions_and_agendas.all()]
+        return set(orgs_list)
 
     def prepare_sort_name(self, obj):
         full_text = obj.extras.get('plain_text')
--- a/lametro/search_indexes.py
+++ b/lametro/search_indexes.py
@@ -29,7 +29,7 @@ class LAMetroBillIndex(BillIndex, indexes.Indexable):
         return None
 
     def prepare_sponsorships(self, obj):
-        orgs_list = [action.organization for action in obj.actions_and_agendas]
+        orgs_list = [action['organization'].name for action in obj.actions_and_agendas]
         return set(orgs_list)
 
     def prepare_sort_name(self, obj):
--- a/lametro/search_indexes.py
+++ b/lametro/search_indexes.py
@@ -29,7 +29,7 @@ class LAMetroBillIndex(BillIndex, indexes.Indexable):
         return None
 
     def prepare_sponsorships(self, obj):
-        orgs_list = [action.organization for action in obj.actions_and_agendas.all()]
+        orgs_list = [action.organization for action in obj.actions_and_agendas]
         return set(orgs_list)
 
     def prepare_sort_name(self, obj):
--- a/tests/test_solr_prep.py
+++ b/tests/test_solr_prep.py
@@ -31,37 +31,16 @@ def test_legislative_session(bill,
 
     assert indexed_data['legislative_session'] == prepared_session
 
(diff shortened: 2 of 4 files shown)


lametro/search_indexes.py
```diff
@@ -29,7 +29,8 @@ def prepare_controlling_body(self, obj):
         return None
 
     def prepare_sponsorships(self, obj):
-        return [action.organization for action in obj.actions.all()]
+        orgs_list = [action.organization for action in obj.actions_and_agendas]
```
hancush, line 32: Getting `AttributeError: 'dict' object has no attribute 'organization'` when I try to update the index. Isn't `actions_and_agendas` a list of dictionaries?

hancush (changes requested):
This looks close!

hancush:
P.s., The testing instructions should include a step to rebuild the index: `docker-compose run --rm app python manage.py update_index` to confirm your Haystack index change works.

lametro/search_indexes.py
fatima3558: woops, yup, it is! notation error on my part.

tests/test_solr_prep.py
```diff
@@ -29,3 +30,63 @@ def test_legislative_session(bill,
...
 
     assert indexed_data['legislative_session'] == prepared_session
+
+@pytest.mark.parametrize('session_identifier,prepared_session', [
+    ('2014', '7/1/2014 to 6/30/2015'),
+    ('2015', '7/1/2015 to 6/30/2016'),
+    ('2016', '7/1/2016 to 6/30/2017'),
+    ('2017', '7/1/2017 to 6/30/2018'),
+    ('2018', '7/1/2018 to 6/30/2019'),
+])
```
hancush, line 40: What role does this play in your test?

tests/test_solr_prep.py
```diff
@@ -29,3 +30,63 @@ def test_legislative_session(bill,
...
+                      event,
+                      mocker):
+    '''
+    This test instantiates LAMetroBillIndex – a subclass of SearchIndex from
+    Haystack, used for building the Solr index.
+
+    The test, then, calls the SearchIndex `prepare` function, 
+    which returns a dict of prepped data.
+    https://github.com/django-haystack/django-haystack/blob/4910ccb01c31d12bf22dcb000894eece6c26f74b/haystack/indexes.py#L198
+    '''
```
hancush, line 55: Don't think we need to repeat this.

```suggestion
```

tests/test_solr_prep.py
```diff
@@ -29,3 +30,63 @@ def test_legislative_session(bill,
...
+    legislative_session.identifier = session_identifier
+    legislative_session.save()
+    bill = bill.build(legislative_session=legislative_session)
```
hancush, line 58: Do you need to manually construct the legislative session to build a test bill?

hancush (changes requested):
This is looking good, @fatima3558 – thanks for writing a test! Your test contains some elements from the example test that aren't strictly necessary. I've left some prompts inline. Can you answer my questions, and then take a pass at removing the unnecessary elements?

tests/test_solr_prep.py
fatima3558: ah, didn't realize there was a legislative session fixture until just now. thanks for flagging this!

tests/test_solr_prep.py
fatima3558: After doing a bit of reading, my understanding is that this decorator takes the tuples and turns them into parameters that can be used in the test. This decorator would take the tuple on line 35 and make `'2014'` be available in the test as `session_identifier`, and `'7/1/2014 to 6/30/2015'` as `prepared_session`, for example.

Before reading this, I thought that this was creating the values needed to create a the legislative session for a bill, but even then I wouldn't have needed all of them. After checking for and finding a `legislative_session` fixture, I understand that I don't need this decorator.

tests/test_solr_prep.py
```diff
@@ -29,3 +30,42 @@ def test_legislative_session(bill,
     indexed_data = index.prepare(bill)
 
     assert indexed_data['legislative_session'] == prepared_session
+
+def test_sponsorships(bill, 
+                      legislative_session,
```
hancush, line 35: 👋

```suggestion
```

hancush (approved):
Works like a charm! Spotted one more fixture that can go, then this is ready to stage! 💪 Thank you, @fatima3558.

Issue #692, opened 2021-01-28
Add delete button to event detail pages


shrayshray:
There are two Finance & Budget meetings showing on the site. The one occurring at 1:30pm is the correct one. Please remove the one scheduled for 1:00pm. It has been removed in Legistar.
![image](https://user-images.githubusercontent.com/28063988/106212542-b1672580-617f-11eb-8ded-a8227baed3a5.png)

hancush:
Done. This has come up several times. I wonder if we should consider adding a delete button to event pages that is only visible to admin users, the same way you can upload an agenda if necessary. Thoughts, @shrayshray?

shrayshray:
@hancush yes! Having a delete button in admin view would be amazing!

hancush:
Task: Add a delete button visible only to admin users to the event detail page.

Context:

- There is some existing logic to show a document upload for admin users on the event page. This could be a helpful starting point / pattern (but also happy for revisions, as it's quite old).
- In addition to an admin user, you may want to consider a check against the Legistar API and only show the button if the event no longer exists. I think, but am not sure, that the document upload does a check (though I don't remember what it is) as well.

@fatima3558 With this in mind, want to do some code spelunking and propose how you'll implement this feature? Bullet points are fine!

fatima3558:
I'm taking a look at the code that currently exists on the event detail page, and it looks to me like there's [already logic to allow an admin to delete an agenda if the `agenda_url` is not present](https://github.com/datamade/la-metro-councilmatic/blob/master/lametro/templates/lametro/event.html#L184). 

The `agenda_url` is available in the template's context if the document note reads "Agenda", which seems to only happen if the agenda was uploaded automatically (based on the document notes in the other branches of the conditional). I'm not sure where this gets set.

Is my understanding of this right, @hancush? If there's already logic in place to do this, then there seems to be a bug in place that is preventing this button from showing up.

hancush:
@fatima3558 No, at a high level, the logic is, if there is not a scraped agenda, show the button. Granted, it's implicit and can certainly be improved (though that is outside the scope of this PR).

The block you linked to is conditional on `has_agenda` being `True`, and `agenda_url` being False-y.

Here's how `has_agenda` is set:

https://github.com/datamade/la-metro-councilmatic/blob/1137e4b3e4cae435d8a81bc14342cfd54f543bbb/lametro/views.py#L204-L206

And here's how the other agenda URL context variables are set:

[...]

Issue #635, opened 2020-07-16
Unify endpoints into coherent API


hancush:
We have a few endpoints for various actions:

- `/refresh-guid/` triggers syncs SES topics
- `/flush-cache/` flushes the cache
- `/public-comment/` redirects to the public comment page for the live meeting

I'm about to add another, `/object-counts/`, to report bill and event counts to scraping dashboard.

It would be ideal to have a unified API framework for these endpoints.

fatima3558:
Django's [`login_required` documentation](https://docs.djangoproject.com/en/3.1/topics/auth/default/#the-login-required-decorator) states that the decorator redirects users to a `LOGIN_URL`, which, if unspecified, is `accounts/login?next=some/redirect/url`. 

We should find a way of removing this redirect since we don't use Django accounts.

hancush:
@fatima3558 We do use Django accounts, but we hit these URLs programmatically, so the redirect when a user is not authenticated is a little awkward.

PR #697, opened 2021-02-16
Event delete button for duplicated events not in Legistar

fatima3558:
## Overview

This PR adds a button to an event page's header to delete the event from the database if it does not exist in Legistar. This happens when a duplicated version of an event has been deleted from Legistar but the object still exists in the database.

### Checklist

## Testing Instructions

 * run `docker-compose -f docker-compose.yml -f tests/docker-compose.yml run --rm app` and make sure all tests pass

Handles #692

### DIFF
--- a/lametro/views.py
+++ b/lametro/views.py
@@ -23,6 +23,7 @@ from django.db import transaction, connection, connections
 from django.conf import settings
 from django.contrib.auth import authenticate, login, logout
 from django.contrib.auth.forms import AuthenticationForm
+from django.contrib.auth.decorators import login_required
 from django.core.exceptions import MultipleObjectsReturned, ObjectDoesNotExist
 from django.shortcuts import render
 from django.db.models.functions import Lower, Now, Cast
@@ -238,6 +239,7 @@ def handle_uploaded_agenda(agenda, event):
     management.call_command('collectstatic', '--noinput')
 
 
+@login_required
 def delete_submission(request, event_slug):
     event = LAMetroEvent.objects.get(slug=event_slug)
     event_doc = EventDocument.objects.filter(event_id=event.id, note__icontains='Manual upload')
@@ -254,6 +256,7 @@ def delete_submission(request, event_slug):
     return HttpResponseRedirect('/event/%s' % event_slug)
 
 
+@login_required
 def delete_event(request, event_slug):
     event = LAMetroEvent.objects.get(slug=event_slug)
     event.delete()
--- a/lametro/views.py
+++ b/lametro/views.py
@@ -255,18 +255,8 @@ def delete_submission(request, event_slug):
 
 
 def delete_event(request, event_slug):
-    # event = LAMetroEvent.objects.get(slug=event_slug)
-    # event_doc = EventDocument.objects.filter(event_id=event.id, note__icontains='Manual upload')
-
-    # for e in event_doc:
-    #     # Remove stored PDF from Metro app.
-    #     if 'Manual upload PDF' in e.note:
-    #         try:
-    #             os.remove('lametro/static/%s' % e.links.get().url )
-    #         except OSError:
-    #             pass
-    #     e.delete()
-
+    event = LAMetroEvent.objects.get(slug=event_slug)
+    event.delete()
     return HttpResponseRedirect('/events/')
 
 
--- a/lametro/views.py
+++ b/lametro/views.py
@@ -160,6 +160,12 @@ class LAMetroEventDetail(EventDetailView):
             r = requests.get('https://metro.legistar.com/calendar.aspx')
             context['legistar_ok'] = r.ok
 
+        event_gid = event.api_source.url.split('/')[-1]
+        response = requests.get('https://metro.legistar.com/MeetingDetail.aspx?LEGID={}&GID=557&G=A5FAA737-A54D-4A6C-B1E8-FF70F765FA94'.format(event_gid))
+        # response = requests.get('https://metro.legistar.com/MeetingDetail.aspx?LEGID=3&GID=557&G=A5FAA737-A54D-4A6C-B1E8-FF70F765FA94')
+        string_in_content = response.content.find(b'This record no longer exists. It might have been deleted.')
+        context['event_in_legistar'] = not bool(string_in_content)
+
         try:
             context['minutes'] = event.documents.get(note__icontains='minutes')
         except EventDocument.DoesNotExist:
@@ -214,6 +220,9 @@ class LAMetroEventDetail(EventDetailView):
 
         context['USING_ECOMMENT'] = settings.USING_ECOMMENT
 
+        import pdb
+        pdb.set_trace()
+
         return context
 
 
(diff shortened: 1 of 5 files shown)


tests/test_events.py
```diff
@@ -330,17 +333,22 @@ def test_event_is_upcoming(event, mocker):
...
 
 
-def test_delete_duplicate_event(event, requests_mock):
+def test_delete_duplicate_event(event, client):
     # create 1 event with a fake api_source and use requests_mock to come up with 2 different responses (an ok and a 404)
     e = event.build()
-    event_url = 'http://webapi.legistar.com/v1/metro/events/0000'
+    slug = e.slug.int
+    # event_template = reverse('events', args=[slug]) # this is throwing a NoReverseMatch error despite my double checking the url pattern name
+    event_url = 'http://webapi.legistar.com/v1/metro/events/{slug}'
```
hancush, line 341: This was tricky. This was failing with the exception:

```
(Pdb) reverse('events', args=[e.slug])
*** django.urls.exceptions.NoReverseMatch: Reverse for 'events' with arguments '(UUID('2b2e72d4-64a2-40a9-945d-0ee7a63e07a8'),)' not found. 1 pattern(s) tried: ['events/$']
```

The thing that stood out to me is that it was trying `events/$` (no args) instead of `event/<slug>/` as defined in `lametro/urls.py`. 

Django resolves URLs by [looking through the URLs declared in each installed app and stopping at the first match](https://docs.djangoproject.com/en/2.2/topics/http/urls/#how-django-processes-a-request). Sure enough, `councilmatic_core` [declares a URL with the name `events`](https://github.com/datamade/django-councilmatic/blob/66e95bae035a635a65d39d74df3be35315c14d83/councilmatic_core/urls.py#L40) that does not accept a `slug` argument, and `lametro` [overrides it](https://github.com/datamade/la-metro-councilmatic/blob/23e1dccf8c60f7cf2c3f5624c05b28b233b7e25c/councilmatic/urls.py#L35). Django's resolver is encountering the `councilmatic_core` URL first, so it stops there and fails.

Luckily, Django allows for [namespacing URL lookups](https://docs.djangoproject.com/en/2.2/topics/http/urls/#reversing-namespaced-urls). In other words, we can tell it to look specifically for the URL named `events` in the `lametro` app. 

Adding the namespace to the URL resolves this issue.

```suggestion
    event_url = reverse('lametro:events', args=[e.slug]) 
```

fatima3558: thanks! I'll implement this when I pick this back up next week.

lametro/views.py
```diff
@@ -251,6 +254,12 @@ def delete_submission(request, event_slug):
     return HttpResponseRedirect('/event/%s' % event_slug)
 
 
+def delete_event(request, event_slug):
```
hancush, line 257: You definitely want to make sure this view cannot be accessed by non-admin users. Can you propose a solution?

tests/test_events.py
```diff
@@ -327,3 +332,46 @@ def test_event_is_upcoming(event, mocker):
...
+    api_source = f'http://webapi.legistar.com/v1/metro/events/{e.slug}'
+    
+    mock_source = mocker.MagicMock()
+    mock_source.url = api_source
+    mock_api_source = mocker.patch('lametro.models.LAMetroEvent.api_source', new_callable=mocker.PropertyMock, return_value=mock_source)
+
+    username = 'admin'
+    password = 'administrativepw'
+    user = django_user_model.objects.create_user(username=username, password=password)
+    client.force_login(user)
```
hancush, line 350: Good thought! `pytest-django` provides an `admin_client` fixture that will do this setup for you: https://pytest-django.readthedocs.io/en/latest/helpers.html#admin-client-django-test-client-logged-in-as-admin

What do you think about switching?

tests/test_events.py
```diff
@@ -327,3 +332,46 @@ def test_event_is_upcoming(event, mocker):
...
+
+        failure = m.get(source_matcher, status_code=404)
+        failure_response = client.get(event_template)
+
+        assert 'This event does not exist in Legistar. It may have been deleted from Legistar due to being a duplicate. To delete this event, click the button below.' not in success_response.content.decode('utf-8')
+        assert 'This event does not exist in Legistar. It may have been deleted from Legistar due to being a duplicate. To delete this event, click the button below.' in failure_response.content.decode('utf-8')
+
+
+@pytest.mark.django_db
+def test_delete_event(event, client):
```
hancush, line 368: You'll need to update this test once you only allow authorized requests to the delete endpoint.

lametro/views.py
```diff
@@ -159,6 +159,9 @@ def get_context_data(self, **kwargs):
         if self.request.user.is_authenticated:
             r = requests.get('https://metro.legistar.com/calendar.aspx')
             context['legistar_ok'] = r.ok
+            # GET the event URL; allow admin to delete event if 404
+            response = requests.get(event.api_source.url)
+            context['event_ok'] = True if response.status_code == 200 else False
```
hancush, line 164: Can you use `response.ok` here (as in L161) instead?

tests/test_events.py
```diff
@@ -327,3 +332,46 @@ def test_event_is_upcoming(event, mocker):
...
+        m.get(cal_matcher, status_code=200)
+
+        source_matcher = re.compile(api_source)
+        success = m.get(source_matcher, status_code=200)
+        success_response = client.get(event_template)
+
+        failure = m.get(source_matcher, status_code=404)
+        failure_response = client.get(event_template)
+
+        assert 'This event does not exist in Legistar. It may have been deleted from Legistar due to being a duplicate. To delete this event, click the button below.' not in success_response.content.decode('utf-8')
```
hancush, line 363: Could you pull this text out into a variable so you only need to update it in one place if it changes?

hancush (changes requested):
Looking good, @fatima3558, thank you! Left some comments inline that need to be addressed before this comes in.

hancush:
P.s., can you remove `[WIP]` from the PR title?

fatima3558:
look up django `login_required` & `admin_client` for this issue

tests/test_events.py
fatima3558: definitely a lot more concise to use that fixture!

tests/test_events.py
```diff
@@ -327,3 +332,49 @@ def test_event_is_upcoming(event, mocker):
...
+    e.save()
+    event_in_db = LAMetroEvent.objects.filter(id=e.id)
+    assert event_in_db.exists()
+
+    delete_event = reverse('delete_event', args=[e.slug])
+    admin_redirect_url = reverse('lametro:event')
+
+    user_response = client.get(delete_event)
+    assert user_response.url != admin_redirect_url
+
```
hancush, line 375: How about asserting the event is still in the database here?

tests/test_events.py
```diff
@@ -327,3 +332,49 @@ def test_event_is_upcoming(event, mocker):
...
+    admin_response = admin_client.get(delete_event)
+    assert admin_response.url == admin_redirect_url
```
hancush, line 377: What if you also checked the status code of these responses? I'd expect the unauthenticated request to return a 403, and the authenticated request to return a 200 (or a 301 or 302 redirect).

hancush (commented):
Two very small edits to your test, and then this looks good to me. Great revision, @fatima3558!

tests/test_events.py
fatima3558: I did check for this, and they both result in a 302 redirect. This is the url that the unauthenticated request attempted to redirect to when I ran my tests: `/accounts/login/?next=/delete-event/8159e708-f892-49cf-84b3-175dad36a7f5/`

I did briefly look in the [django docs](https://docs.djangoproject.com/en/3.1/topics/auth/default/#django.contrib.auth.decorators.login_required) to see if there was a way to avoid that redirect but didn't come across anything that would allow me to throw a 403 error, I only found a way to change the `REDIRECT_URL`. Let me know what you think as far as whether I should make any other changes

hancush (approved):
This is awesome, @fatima3558, thank you! Would you mind adding a comment to #635 about figuring out a more graceful path forward with `login_required`?

Issue #694, opened 2021-02-03
Show today's meetings on the homepage on meeting days


hancush:
In place of the About section to the right of the upcoming / current meeting block, show the day's meetings and their status (passed, live, scheduled or something to that effect).

hancush:
Part 1: Show the About section or a new "Today's Meetings" section, conditional on whether there are any meetings scheduled today.

Part 2: Render the day's meetings with the correct status. Leverage [existing code](https://github.com/datamade/la-metro-councilmatic/blob/1137e4b3e4cae435d8a81bc14342cfd54f543bbb/lametro/models.py#L634-L644) to determine whether a meeting is passed, ongoing, or scheduled.

@fatima3558 How about you submit a draft PR for review after you finish Part 1?

fatima3558:
@hancush was able to get part 1 done without too much trouble. Here's the first part of the PR: https://github.com/datamade/la-metro-councilmatic/pull/698

fatima3558:
@shrayshray The changes for this issue are now on the staging site! It looks like the soonest there will be meetings is Thursday, May 27. Could you give the staging site a look sometime then and let us know what you think?

shrayshray:
@fatima3558 Wow, this looks great!!
![image](https://user-images.githubusercontent.com/28063988/119873482-7250a300-bed9-11eb-92a1-93d87143e02a.png)

fatima3558:
@shrayshray great! are there any tweaks we should make before deploying to the live site?

shrayshray:
@fatima3558 no tweaks at the moment, though I do want to keep an eye on how it displays on days with multiple meetings. I'm comfortable with going live though - thanks, Fatima!

shrayshray:
@fatima3558 i noticed on little issue with display - the "upcoming" buble on the first meeting looks like it's cut-off on the line break: 
![image](https://user-images.githubusercontent.com/28063988/120525291-d3babb00-c38c-11eb-937a-3425a6962241.png)

hancush:
@shrayshray Ah! What browser are you using?

shrayshray:
@hancush Firefox. I just checked and it's happening on "Concluded", too - but it looks great in Chrome!

hancush:
Thanks, @shrayshray! I just put a patch in and forced meetings to display to test in Firefox. See https://github.com/datamade/la-metro-councilmatic/pull/729#issue-660962057.

Ok to go ahead and deploy?

[...]

PR #698, opened 2021-02-23
Show today's meetings in place of about text

fatima3558:
## Overview

This PR checks whether there are meetings today and replaces the about text on the index page with a list of today's meetings if there are meetings today.

### Checklist

### Demo
This is what the relevant portion of the index page looks like currently:
Screen Shot 2021-02-23 at 4 21 33 PM

This is what the page currently looks like if there are meetings scheduled for today:
Screen Shot 2021-05-19 at 12 23 23 PM

## Testing Instructions

 * on your local app, do a windowed scrape for a few events using `docker-compose run --rm scrapers pupa update lametro events --window=0.05 rpm=0`
 * bring the app up locally and go to the site homepage. Make sure you see a list of today's meetings replace the description text to the right of the "Next Board Meeting" section

Handles #694


lametro/views.py
```diff
@@ -76,6 +76,10 @@ def extra_context(self):
         extra['bilingual'] = bool([e for e in extra['current_meeting'] if e.bilingual])
         extra['USING_ECOMMENT'] = settings.USING_ECOMMENT
 
+        upcoming = extra['upcoming_board_meetings']
```
hancush, line 79: A couple of questions:

- What is returned from `LAMetroEvent.upcoming_board_meetings()`?
- We truncate whatever is returned to two items. Will this result in an incomplete list of upcoming meetings? (Are there ever more than two events on one day?)

hancush (changes requested):
Thanks for getting started on this, @fatima3558. I left a couple of questions inline. Let's table this until you land #697.

lametro/views.py
fatima3558: @hancush here are the answers to your questions:

* `LAMetroEvent.upcoming_board_meetings()` returns `LAMetroEvent` objects that contain the term 'Board Meeting' starting now or later.
* Yes, there are usually more than two events in one day. Based on a quick glance at Metro Legistar, it doesn't look like there are ever more than 2 events that contain the term 'Board Meeting' in a day, but I don't know of any reason why there couldn't be

hancush: @fatima3558 Both of these answers suggest that `upcoming_board_meetings()` will omit meetings, i.e., it isn't a good solution to populate the new section. You want a list of all meetings with a start date of today. Can you propose a different solution?

fatima3558: @hancush looks like the `is_upcoming` property is closer to what I need. It would include any meetings scheduled to start day or tomorrow (if the local time is after 5pm) up until the meeting starts. I would use this instead of `extra['upcoming_board_meetings']`

hancush: @fatima3558 Does `is_upcoming` return a queryset of events meeting those criteria?

hancush: I don't think that any of the existing properties are going to generate the list of events that you need. I'd like you to write a new property on the `LAMetroEvent` class that returns all meetings with a start date of the current date (being sure to calculate the current date in LA time, not UTC). Can you draft that, @fatima3558?

fatima3558: ah, to answer your first question -- it returns a boolean, not a queryset

I can go ahead and draft a new property.

hancush: Thanks, @fatima3558!

lametro/models.py
```diff
@@ -664,6 +664,14 @@ def ecomment_message(self):
...
             return self.UPCOMING_ECOMMENT_MESSAGE
 
+    @property
+    def todays_meetings(cls):
+        now = app_timezone.localize(datetime.now())
+
+        end_of_day = app.timezone.localize(datetime(now.year, now.month, now.day, 17, 0))
+
+        return cls.objects.filter(start_time__gte=now, start_time__lte=end_of_day)
+
```
fatima3558, line 674: @hancush here's where I wrote up my first go at this property.

lametro/models.py
```diff
@@ -664,6 +664,14 @@ def ecomment_message(self):
         else:
             return self.UPCOMING_ECOMMENT_MESSAGE
 
+    @property
+    def todays_meetings(cls):
+        now = app_timezone.localize(datetime.now())
```
hancush, line 669: Academic Q: Django has [a `timezone` util](https://docs.djangoproject.com/en/3.2/topics/i18n/timezones/). Does `timezone.now()` yield the same result as `app_timezone.localize(datetime.now())`?

lametro/models.py
```diff
@@ -664,6 +664,14 @@ def ecomment_message(self):
...
+
+        end_of_day = app.timezone.localize(datetime(now.year, now.month, now.day, 17, 0))
```
hancush, line 671: Why 5:00?

lametro/models.py
```diff
@@ -664,6 +664,14 @@ def ecomment_message(self):
...
+
+        return cls.objects.filter(start_time__gte=now, start_time__lte=end_of_day)
```
hancush, line 673: Per the issue description, the ultimate goal is to "show the day's meetings and their status (passed, live, scheduled or something to that effect)". Will this query return all meetings on a given date?

lametro/views.py
```diff
@@ -76,6 +76,10 @@ def extra_context(self):
...
+        today = timezone.now().date()
+        extra['todays_meetings'] = [e for e in upcoming if e.start_date == today]
```
hancush, line 81: Can you update this to populate `extra['todays_meetings']` from your new property?

hancush (changes requested):
Moving in the right direction! Left a few things inline. Let me know if you want to chat through anything out loud.

lametro/models.py
fatima3558: Before reading the docs, I thought about this and was thinking that it would be a big waste for Django to have a `timezone` util if it didn't make use of the ability to set the timezone in the project's `settings.py` file. If it does make use of this setting (which I was assuming the util did), then the two functions should return the same value.

After taking a glance at the docs you linked to and checking `councilmatic/settings.py`, I was able to verify that we do set the timezone to LA time, so they are equivalent here. It's also easier to read, so I think I'll make use of `timezone.now()`.

lametro/models.py
fatima3558: ah, thanks for bringing my attention back to that part of the issue description. This would remove meetings from the list if they have already started (if they are live or passed).

hancush:
Sounds great, @fatima3558. Proceed with adding that property and adding the events with their display status to the template. 😎

lametro/templates/lametro/index.html
```diff
@@ -66,7 +66,37 @@ <h2>
                         {% include "partials/meeting_details_next.html" %}
                     </div>
                     <div class='col-md-5'>
-                        {% include "partials/index_metro_description.html" %}
+                        {% if todays_meetings %}
```
hancush, line 69: Nesting the display logic like this will result in today's meetings not being shown when there is a current meeting. Can you revise the conditional logic in the index template so today's meetings are always shown if they exist?

hancush (changes requested):
Your screenshots look super good, @fatima3558! I left one comment about the template conditional, but this is very close. 👏

hancush (approved):
This looks great, @fatima3558! One thing inline, then want to merge to deploy to staging, then update the issue thread and prompt Shelly to review on Thursday, which is the next day meetings are scheduled?

lametro/views.py
```diff
@@ -76,6 +76,9 @@ def extra_context(self):
         extra['bilingual'] = bool([e for e in extra['current_meeting'] if e.bilingual])
         extra['USING_ECOMMENT'] = settings.USING_ECOMMENT
 
+        upcoming = extra['upcoming_board_meetings']
```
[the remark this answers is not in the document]
fatima3558: good call, I'll remove that 👍🏼

Issue #623, opened 2020-06-01
Bills that are deleted or made private are not always removed from the search index


hancush:
This can be easily fixed by shelling into the server and rebuilding the index (6-8 minute operation), but it would be great if we didn't have to do this.

The `update_index` command from Haystack [includes a `--delete` option](https://django-haystack.readthedocs.io/en/master/management_commands.html#update-index) that will remove items from the index that are no longer in the database, however last time I tried to use it, the command failed. Might be worth investing the time (1-2 hours, I'd venture) to debug and patch this.

Alternatively, we could rebuild the index at some interval, say very early every morning, when it's unlikely users are on the site.

hancush:
Update: I thought Haystack had removed the option to delete items in the search index if they are no longer in the database, but it's just been renamed to `--remove`! I think the work here is to update the indexing commands [in the `hourly_processing` DAG](https://github.com/datamade/la-metro-dashboard/blob/29daf5cfb3d5e24abf4c49d864f92fa592658fd0/dags/hourly_processing.py#L50-L53) to use the `--remove` option.

hancush:
Done and deployed!

PR #706, opened 2021-03-09
Override haystack.utils.default_get_identifier

fatima3558:
## Overview

This PR compliments [la-metro-dashboard/pull/72](https://github.com/datamade/la-metro-dashboard/pull/72), which adds the `--remove` option to the `update_index` command.

This PR overrides haystack's `default_get_identifier` method, which provides the first step in deleting records from the haystack index. We override haystack's `id` property, so this code returns the identifier formats we use rather than the default haystack format.

### Checklist

### Notes
The `update_index` command cleared the index out before recreating it when I ran it locally. I spent a few minutes trying to see if I could find this management command, but I couldn't find it in `la-metro-councilmatic` or `django-councilmatic`. I was hoping to identify where this command lives to double-check that the logic is set up to consistently clear the index before recreating it & include that code here.

Also, once this PR is brought in, I can go ahead and test the related [la-metro-dashboard/pull/72](https://github.com/datamade/la-metro-dashboard/pull/72) PR.

## Testing Instructions

 * run `docker-compose run --rm app python manage.py update_index --remove` and make sure the command runs successfully

Handles #623

### DIFF
--- a/councilmatic/settings.py
+++ b/councilmatic/settings.py
@@ -118,5 +118,6 @@ PIC_BASE_URL = 'https://pic.datamade.us/lametro/document/'
 # PIC_BASE_URL = 'http://127.0.0.1:5000/lametro/document/'
 
 HAYSTACK_SIGNAL_PROCESSOR = 'haystack.signals.RealtimeSignalProcessor'
+HAYSTACK_IDENTIFIER_METHOD = 'lametro.utils.get_identifier'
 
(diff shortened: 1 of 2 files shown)


fatima3558:
No, I didn't, which is odd. I ran `update_index` and know that the util function I wrote got run because the command failed and the traceback pointed me to the code I wrote. I did run the command you suggested, though, and ran without issue

Issue #674, opened 2020-11-18
Home: Text overlapping on mobile site


shrayshray:
Omar and I are both seeing overlapping text on the Home page of the mobile site (Android/Chrome and iOS/Safari). 
"Current Meeting" or "Next Meeting" header overlaps the red text below the search bar "Search the Metro Board Archive from 1952-2015 (Board Boxes 1994-Present)"
![image](https://user-images.githubusercontent.com/28063988/99593453-89508e80-29a6-11eb-8094-3becc7a046da.png)

fatima3558:
@shrayshray we've implemented a fix that's available for you to take a look at on the staging site. Let us know how it looks to you!

shrayshray:
@fatima3558 looks great on staging, thank you!

PR #719, opened 2021-04-14
fix text overlap on mobile issue

fatima3558:
## Overview

This PR addresses an issue on mobile devices where the text under the home page search bar overlaps with the heading for the next section.

### Checklist

### Demo
Screen Shot 2021-04-20 at 1 30 10 PM

### Notes
In resolving the text overlap issue, I introduced an issue where the top margin of the `h1` that reads "Los Angeles County Metropolitan Transportation Authority" It looks like we're using Bootstrap 2 in this project, and the classes I would usually use to remove that margin are unavailable in this version of Bootstrap. I'd be interested in any ideas around how to remove this whitespace.

This is no longer an issue as of second review request.

## Testing Instructions

 * run the app locally and use your browser's device toolbar to toggle mobile view
 * make sure none of the text on the home page overlaps, and in particular that the text on the background image at the very top of the page doesn't overflow its bounds

Handles #674

### DIFF
--- a/lametro/templates/lametro/index.html
+++ b/lametro/templates/lametro/index.html
@@ -7,7 +7,7 @@
 {% endblock %}
 
 {% block full_content %}
-    <div id="section-photo">
+    <div id="section-photo" class="container-fluid">
         <div class="row">
             <div class='col-sm-10 col-sm-offset-1'>
                 <span class="hidden-xs hidden-sm"><br/><br/></span>
--- a/lametro/templates/lametro/index.html
+++ b/lametro/templates/lametro/index.html
@@ -7,7 +7,8 @@
 {% endblock %}
 
 {% block full_content %}
-    <div class="container-fluid" id="section-photo">
+<div class="contianer">
+    <div id="section-photo">
         <div class="row-fluid">
             <div class='col-sm-10 col-sm-offset-1'>
                 <span class="hidden-xs hidden-sm"><br/><br/></span>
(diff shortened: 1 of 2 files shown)


hancush (changes requested):
Hm, I'm still seeing the overlap on sufficiently small screens. 

Screen Shot 2021-04-14 at 3 44 20 PM

Is there a fixed height on the `section-photo` class, by chance?

hancush:
P.s., I think the contemporary version of this issue is #674.

fatima3558:
when you say you still see the overlap, is it because part of the text is on a white background?

And I did look at the `section-photo` class. There is not a fixed height, but there is a minimum height. I can play around with the image size and position, see if I can get closer to what we need.

lametro/templates/lametro/index.html
```diff
@@ -7,8 +7,8 @@
 {% endblock %}
 
 {% block full_content %}
-    <div class="container-fluid" id="section-photo">
-        <div class="row-fluid">
+    <div id="section-photo">
```
hancush, line 10: How about retaining `.container-fluid` to preserve padding on smaller screens?

```suggestion
    <div id="section-photo" class="container-fluid">
```

Without:

Screen Shot 2021-04-20 at 1 46 21 PM

With:

Screen Shot 2021-04-20 at 1 45 13 PM

hancush (changes requested):
Looking good! One small tweak suggested inline.

hancush (approved):

PR #725, opened 2021-04-19
Patch actions_and_agendas to handle events without related organizations

hancush:
## Overview

Some events are not associated with an organization. This PR updates `actions_and_agendas` to try to return an `LAMetroOrganization` object, falling back to the event related entity if one does not exist.

### Checklist

### Notes

I maintained the event participant fallback because we still want to be able to filter by meetings that aren't associated with a body, such as public hearings, in the search. If we didn't fall back, those meetings would not be available as facets.

## Testing Instructions

 * Visit https://lametro-upgrade.datamade.us/board-report/2020-0522/ and confirm the page loads, and the actions and agendas are rendered as expected.
 * Confirm CI passes.
 * Review the code.
 
Connects #712

### DIFF
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -235,7 +235,7 @@ class LAMetroBill(Bill, SourcesMixin):
                 # If a corresponding org does not exist, e.g., in the case of
                 # appearing on the agenda of a public hearing, do not return an
                 # organization.
-                org = None
+                org = event.participants.first()
 
             event_dict = {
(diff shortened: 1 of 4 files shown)


fatima3558 (approved):
Looks good to me! 

Still working through the code for my own understanding (esp around EventParticipant vs EventRelatedEntity), but I'll approve this now so it's in before the end of your work day.

Issue #716, opened 2021-04-07
Unpopulated "Significant Date" facet


fatima3558:
Hi @shrayshray! In addressing the facet ordering issue, I noticed the "Significant Date" facet isn't populating. Do you have any info about whether you expect this facet to be populated or not?

shrayshray:
Hi, @fatima3558! "Significant Date" is the label we want to use for the facet, but the Class is actually "Date". I think we changed the class name at some point because it caused confusion on [this issue](https://github.com/datamade/la-metro-councilmatic/issues/463#issuecomment-648350338), too.

hancush:
Looks like we have Dates plural! 

https://github.com/datamade/la-metro-councilmatic/blob/8a83113363e586e6853532a326dbda0239f5a892/lametro/management/commands/refresh_guid.py#L70-L72

Hopefully changing to singular will resolve issue.

hancush:
@shrayshray This is also done and deployed to staging. Care to double check the significant date facet?

Issue #370, opened 2018-11-02
Board Reports: Filter by Fiscal Year error


shrayshray:
"Filter by Fiscal Year" displays results outside of the selected fiscal year. ![image](https://user-images.githubusercontent.com/28063988/47942904-fc1a6a00-deb0-11e8-9c1d-ec4ce848a73a.png)

hancush:
hey, @shrayshray. this looks to be the same issue as https://github.com/datamade/la-metro-councilmatic/issues/181#issuecomment-328914222. 

namely, the filters are working as expected. the legislative session filter applies to date the item was introduced, whereas the highlighted dates refer to the last action taken or, absent actions, the last agenda the item appeared on. 

we added the most recent action label to clarify this in #181, however it seems it may still be confusing. would you like us to implement one of the other solutions @reginafcompton proposed? i.e.,

- add a line for legislative session to each search result. (small change)
- add the introduction date to each search result. (large change, involving edits to the scraper and data import)

shrayshray:
@hancush I reviewed this thread and #181, and I think crossed our wires somewhere a long the line, so I'll try to clarify. Let me know whether this makes sense:

1. The filter on the sidebar should definitely match the date on the report as displayed in the search results.
2. If possible, we should use the Agenda date as the criteria for filter by Legislative Session.

Context:
A. Right now, the "Legislative Session" is determined by the 'MatterIntroDate' given in the Legistar API. The 'MatterIntroDate' comes from the date the **file was created.** Regina uses File ID 2016-0499 as an example in #181 "For example, [Board Report 2016-0499](https://boardagendas.metro.net/board-report/2016-0499/) has a MatterIntroDate of "2016-06-06": http://webapi.legistar.com/v1/metro/matters/3016. Thus, this board report appears in the 7/1/2015 to 6/30/2016 fiscal year (i.e., legislative session) when filtering.
![image](https://user-images.githubusercontent.com/28063988/52678231-bfb1cd80-2ee5-11e9-89b0-86464e3d43b7.png)
It usually takes at least a few months from the time a report was created until it ends up on an Agenda and in front of the Board. Until it's on an Agenda, it's not public; the date created isn't entirely helpful, from a user standpoint. Most people would be be interested in the Agenda date and Action date.

B. We're not getting entries in Minutes History consistently enough to rely on "Most Recent Action".
Thus, (2.) above -- can we use the Agenda date as the filter criteria?

hancush:
[...]

PR #727, opened 2021-06-01
fix `Significant Date` facet and add `Fiscal Year` facet

fatima3558:
## Overview

See title

### Checklist

## Testing Instructions

[for @fatima3558 to do]
 * Deploy to staging by shelling into the lametro server
 * run the `refresh_guid` command on the staging server
 * visit the staging site and check that the `Significant Date` facet is properly populated

Handles #716 
Handles #370


lametro/management/commands/refresh_guid.py
```diff
@@ -78,6 +78,9 @@ class ClassificationMixin:
             'Program',
             'Policy'
         ),
+        'legislative_session': (
+            'Legislative Session'
+        )
```
hancush, line 83: We aren't pulling this from SES, so we don't need to include it in this script.

lametro/models.py
```diff
@@ -911,7 +911,8 @@ class LAMetroSubject(models.Model):
         ('significant_date_exact', 'Significant Date'),
         ('motion_by_exact', 'Motion By'),
         ('topics_exact', 'Subject'),
-        ('plan_program_policy_exact', 'Plan, Program, or Policy')
+        ('plan_program_policy_exact', 'Plan, Program, or Policy'),
+        ('legislative_session', 'Legislative Session')
```
hancush, line 915: Ditto here – legislative session is not a classification we pull from SES, so we don't need this.

lametro/search_indexes.py
```diff
@@ -21,6 +21,7 @@ class LAMetroBillIndex(BillIndex, indexes.Indexable):
     significant_date = indexes.MultiValueField(faceted=True)
     motion_by = indexes.MultiValueField(faceted=True)
     plan_program_policy = indexes.MultiValueField(faceted=True)
+    legislative_session = indexes.MultiValueField(faceted=True)
```
hancush, line 24: This is already defined [on the base index in `django-councilmatic`](https://github.com/datamade/django-councilmatic/blob/5b074f376667766e4b6dbf093871d294bb35fc51/councilmatic_core/haystack_indexes.py#L30) – why redefine here?

lametro/search_indexes.py
```diff
@@ -50,8 +51,18 @@ def prepare_attachment_text(self, obj):
         )
 
     def prepare_legislative_session(self, obj):
-        start_year = obj.legislative_session.identifier
-        end_year = int(start_year) + 1
+        if len(obj.actions_and_agendas) > 0:
+            most_recent = sorted(obj.actions_and_agendas, key=lambda i: i['date'],reverse=True)[0]
+            action_date = most_recent['event'].start_time
+        else:
+            action_date = obj.get_last_action_date()
```
hancush, line 58: `obj.actions_and_agendas` should include all actions, so this method will return `None`, which will cause the subsequent logic which assumes a date to fail. If there are no actions or agendas, we should return `None` here.

```suggestion
        else:
            return None
```

lametro/search_indexes.py
```diff
@@ -50,8 +51,18 @@ def prepare_attachment_text(self, obj):
         )
 
     def prepare_legislative_session(self, obj):
-        start_year = obj.legislative_session.identifier
-        end_year = int(start_year) + 1
+        if len(obj.actions_and_agendas) > 0:
```
hancush, line 54: Let's store the result of `actions_and_agendas` under a variable so it is only evaluated once.

lametro/search_indexes.py
```diff
@@ -50,8 +51,18 @@ def prepare_attachment_text(self, obj):
...
+            most_recent = sorted(obj.actions_and_agendas, key=lambda i: i['date'],reverse=True)[0]
+            action_date = most_recent['event'].start_time
```
hancush, line 56: This isn't quite right. The desired logic is to return the session associated with the most recent agenda date. If there are no agenda dates, return the session associated with the most recent action date. If there is neither an agenda date or an action date, return None.

hancush: Can you propose how you'd go about getting the desired date, using the output of `obj.actions_and_agendas`?

lametro/search_indexes.py
```diff
@@ -50,8 +51,18 @@ def prepare_attachment_text(self, obj):
...
+        else:
+            action_date = obj.get_last_action_date()
+
+        if action_date.month <= 6:
+            start_year = action_date.year - 1
+            end_year = action_date.year
+        else:
+            start_year = action_date.year
+            end_year = action_date.year + 1
```
hancush, line 65: Good!

lametro/models.py
hancush: You might be able to remove the new migration when you remove this choice.

hancush (changes requested):
Thanks for your tenacity on this, @fatima3558. The new code is in the right place – the logic just needs a little revision!

Let's chat at the start of your next Metro block. I want to sync up on some of the finer (and undocumented 🙀) parts of search indexing before you bring this over the line.

lametro/search_indexes.py
fatima3558: As I'm thinking about it, I would have to add additional logic to check if the `most_recent` event is an agenda (`description` being `SCHEDULED`) or an action. I would use the agenda if possible, otherwise using an action as the `most_recent`.

Either way, I can use `most_recent['date']`, unless a `datetime` instance is preferable to a `date` instance for this?

fatima3558: @hancush am I on the right track with this?

hancush: I'd recommend something like:

```python
actions_and_agendas = sorted(obj.actions_and_agendas, key=lambda i: i['date'], reverse=True)
agendas = [item for item in actions_and_agendas if it is an agenda]  # will be sorted bc list order is preserved in python
if agendas:
    # return the most recent date from your agendas list
else:
    # there are no agendas, i.e., everything in actions_and_agendas is an action
    # return the most recent date from your actions_and_agendas list

Issue #738, opened 2021-06-18
Metro PDF Merger worker failed silently


fgregg:
The metro pdf merger worker failed because of an out of memory condition due to transient memory pressure on the server.

Because of that failure, board packets have not been building for a few days, which was finally noticed by @shrayshray today.

Suggested mitigation:

1. Update the supervisor settings to automatically attempt to restart the worker
2. Make sure that this type of failure alerts DataMade

fgregg:
It may also be worth considering a different data flow, wherein after a scraping job task finishes, airflow then picks up a compile_pdfs tasks that runs a new `lametro` management command that actually runs the compiling code that currently lives in `metro-pdf-merger`, and [saves the files in a S3 File store](https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html). That management command would have to run in a docker instance that had the system dependencies.

This would remove metro-pdf-merger completely, both as the place where pdf merges happen but also as the server for serving files. 

We could do something similar for the files we are currently mirroring on pic.datamade.us, and have an airflow job and management command take reponsibility for putting those files in s3 storage.

fgregg:
This consolidation would reduce the footprint of the application considerably, and bring all the data tasks into the airflow framework.

fgregg:
the worker failed again. Here's the traceback:

```bash
[June 18, 2021 - 19:52:31][ERROR] tasks.py:line:120 | <class 'SystemExit'> cause
d the failure of writing regular-board-meeting-c62a8c6bcd17 as a PDF, and we cou
ld not merge this file collection: 
[...]

PR #766, opened 2021-10-25
Revise packet creation to integrate with dashboard-based merger

hancush:
## Overview

This PR:

- Updates `_merge_docs` to trigger a `compile_packet` DAG with the relevant values for identifier and attachment links.
- Revises requests for merged documents to hit S3 directly, instead of our proxy file server.
  - Adds a data migration to re-save all packets to update their URLs accordingly.
- Replaces `MERGER_BASE_URL` with two settings: `MERGE_HOST` and `MERGE_ENDPOINT`.

### Checklist

## Testing Instructions

Performed by @hancush:

 * Manually deploy the app and dashboard to staging, and build/push a staging tag for the app.
 * Shell into the server and save the 50 most recent event packets.
   * 42 of the 50 packets built successfully, with most of the errors being transient 404s we can address by telling scrapelib to retry 404s. Outstanding issues:
     *  https://github.com/datamade/councilmatic-document-merger/issues/1
     * https://github.com/datamade/councilmatic-document-merger/issues/2
     * https://github.com/datamade/councilmatic-document-merger/issues/3
 * Download the generated packets and confirm they look ok.

Handles #738


lametro/models.py
```diff
@@ -825,8 +828,23 @@ def is_ready(self):
         return self.ready
 
     def _merge_docs(self):
-        merge_url = settings.MERGER_BASE_URL + '/merge_pdfs/' + self.related_entity.slug
-        requests.post(merge_url, json=self.related_files)
+        headers = {
+            'cache-control': 'no-cache',
+            'content-type': 'application/json',
```
fgregg, line 833: if you use the `json` argument in `request.post` you should not need to specify the content-type

why do we need cache-control here?

fgregg (changes requested):

lametro/models.py
hancush: Not sure. `cache-control` was in the example request in the Airflow docs: https://airflow.apache.org/docs/apache-airflow/stable/deprecated-rest-api-ref.html#endpoints

fgregg: by default, the requests methods don't respect anything related to caching headers, so we can omit that too.

hancush: Got it, removed the headers entirely in https://github.com/datamade/la-metro-councilmatic/pull/766/commits/13c5726ef17ef7777c46a70d65626738309b3fd3.

hancush:
This has been running on staging for ~10 days. Looks like there have been ~75 merges with only 4 errors: 3 instances of "could not merge encrypted file" (https://github.com/datamade/councilmatic-document-merger/issues/2) and one with an invalid filename (https://github.com/datamade/councilmatic-document-merger/issues/3). Want to get a sense of whether encrypted filenames can be handled in the doc creation workflow before going down that rabbit hole, plan to bring up at next priorities.

Issue #123, opened 2017-04-03
Add definitions of Board Report "Status" on About page


reginafcompton:
Omar and Shelly will provide definitions for Approved, Active, Withdrawn, and Adopted.

shrayshray:
Here are the report actions definitions! [Report Action Definitions.xlsx](https://github.com/datamade/la-metro-councilmatic/files/3556750/Report.Action.Definitions.xlsx)

hancush:
Also update Board Box to Board Correspondence.

hancush:
Start with the content we have (https://github.com/datamade/la-metro-councilmatic/issues/123#issuecomment-526289354) and then add board correspondence when it arrives.

fatima3558:
@shrayshray this has been merged into staging! could you confirm that this looks good on your end? 

https://lametro-upgrade.datamade.us/about/

shrayshray:
@fatima3558 this looks great!

PR #767, opened 2021-10-25
Add Board Report Status definitions to About page

fatima3558:
## Overview

See title.

### Checklist

### Demo
Screen Shot 2021-10-25 at 2 54 07 PM

## Testing Instructions

 * Check out this PR locally, bring up the app, and visit the About page
 * Make sure the status names and definitions make sense with each other & the table is responsive.

Handles #123


fatima3558:
```python
if description.upper() in ['APPROVED', 'APPROVED AS AMENDED', 'APPROVED THE CONSENT CALENDAR']:
                return 'Approved'
            elif description.upper() in ['ADOPTED', 'ADOPTED AS AMENDED']:
                return 'Adopted'
            elif description.upper() in ['RECOMMENDED FOR APPROVAL', 'RECOMMENDED FOR APPROVAL AS AMENDED', 'REFERRED', 'FORWARDED DUE TO ABSENCES AND CONFLICTS', 'FORWARDED WITHOUT RECOMMENDATION', 'NO ACTION TAKEN', 'NOT DISCUSSED']:
                return 'Active'
            elif description.upper() in ['RECEIVED', 'RECEIVED AND FILED']:
                return 'Received'
            elif description.upper() == 'FAILED':
                return 'Failed'
            elif description.upper() == 'WITHDRAWN':
              return 'Withdrawn'```

**Comment by fatima3558** [comment]
@hancush Here's the [PR for django-councilmatic](https://github.com/datamade/django-councilmatic/pull/281). I took a quick look at the build failure but wasn't able to get very far before lunch. I can take a closer look later this week.

File: councilmatic/settings_jurisdiction.py [hunk 1]
```diff
@@ -233,6 +233,119 @@
     },
 ]
 
+BILL_STATUS_DESCRIPTIONS = [
+    {
+        'name': 'ADOPTED',
```
hancush, line 238: With this structure, we duplicate the mapping of action to status: It is now here and in the `_status` method on the `LAMetroBill`. It seems right to store this mapping as a configuration variable. Can you propose a different structure for this dictionary that you could use to populate the About page and also to determine the corresponding status to a given action in the `_status` method?

lametro/templates/partials/board_report_statuses.html
```diff
@@ -0,0 +1,29 @@
+<table class="table">
+    <thead>
+        <tr>
+            <th>Status</th>
+            <th class="desktop-only">Description</th>
+        </tr>
+    </thead>
+    <tbody>
+    {% for status in BILL_STATUS_DESCRIPTIONS %}
```
hancush, line 9: I know I proposed adding this to `city_context` but it occurs to me that we won't have this info for the other instances. What do you think about scratching the change to `django-councilmatic` and passing this variable directly into the Metro about page context?

councilmatic/settings_jurisdiction.py
```diff
@@ -233,6 +233,119 @@
...
+        'definition': 'Item was approved as amended by a majority vote of the Board.',
+        'search_term': 'approved'
+    },
+    {
+        'name': 'APPROVED ON CONSENT CALENDAR',
+        'definition': 'Item was approved on Consent Calendar by a majority vote of the Board.',
+        'search_term': 'approved'
+    },
+    {
+        'name': 'CARRIED OVER --> (Sent To)',
```
hancush, line 263: Does this action actually include `--> (Sent To)`?

hancush (changes requested):

councilmatic/settings_jurisdiction.py
fatima3558: This is how it was in the spreadsheet Shelly provided. Do you think this is something worth clarifying? There's at least one other place where the board report status name from the spreadsheet doesn't match the code.

councilmatic/settings_jurisdiction.py
fatima3558: I'm not sure if this is what you mean, but I could remove the `search_term` key from each dictionary & add the `search_term` to the dictionary using the `_status` method in the view?

My only concern with this is that there are a few of these bill status definitions that aren't accounted for in the `_status` method. Those include all of the `under reconsideration` statuses:
* APPROVED UNDER RECONSIDERATION
* FAILED UNDER RECONSIDERATION
* ADOPTED UNDER RECONSIDERATION
* APPROVED AS AMENDED UNDER RECONSIDERATION
* ADOPTED AS AMENDED UNDER RECONSIDERATION

I thought these fit pretty clearly within the status facet buckets we currently have, so I added those `search_term`s.

councilmatic/settings_jurisdiction.py
hancush: These should match the statuses as they appear on the site (in our code), IMO.

councilmatic/settings_jurisdiction.py
hancush: Let's talk out loud about this. Can you put 5 minutes on the calendar when you next plan to work on Metro?

fatima3558: I will likely start working on metro after our checkin next week. can we talk during our meeting?

hancush: Let's do it! I updated the event name to Sprint planning/pairing as a reminder.

fatima3558:
Confirmed that delete button doesn't show up for every board meeting on staging. @hancush this is good to go!

Screen Shot 2021-11-10 at 9 11 31 AM

fatima3558:
I'll reorganize the `BILL_STATUS_DESCRIPTIONS` to look like this:

```python
BILL_STATUS_DESCRIPTIONS = {
    'ADOPTED': {
        'definition': "definition",
        'search_term': 'Adopted'
    },
    ...
}
```

lametro/templates/partials/board_report_statuses.html
```diff
@@ -0,0 +1,33 @@
...
+            <th class="desktop-only">Description</th>
+        </tr>
+    </thead>
+    <tbody>
+    {% for status, obj in BILL_STATUS_DESCRIPTIONS.items %}
+        <tr>
+            <td>
+                <p class='nowrap'>
+                    {% if obj.search_term != None %}
+                    <a href='/search/?selected_facets=bill_type_exact%3A{{obj.search_term}}'>{{status}}</a>
```
hancush, line 14: `bill_type_exact` will look in the Report Type facet for the specified value. You want the Status facet. Head over to the search on the live site and click one of the options under the Status facet to get the name of that facet from the URL.

lametro/templates/partials/board_report_statuses.html
```diff
@@ -0,0 +1,33 @@
...
+                    {% else %}
+                    <p>{{status}}</p>
```
hancush, line 16: What do you think about title casing this for display? I believe there's a Django template filter for that.

lametro/models.py
```diff
@@ -141,21 +142,9 @@ def inferred_status(self):
...
-                return 'Active'
-            elif description.upper() in ['RECEIVED', 'RECEIVED AND FILED']:
-                return 'Received'
-            elif description.upper() == 'FAILED':
-                return 'Failed'
-            elif description.upper() == 'WITHDRAWN':
-              return 'Withdrawn'
-        else:
-            return None
+        if description and BILL_STATUS_DESCRIPTIONS[description.upper()]:
```
hancush, line 145: What happens if the specified description is not in the descriptions dictionary?

councilmatic/settings_jurisdiction.py
```diff
@@ -233,6 +233,97 @@
...
+        'definition': 'No action was taken on the item.',
+        'search_term': 'Active'
+    },
+    'NOT DISCUSSED': {
+        'definition': 'Item was not discussed.',
+        'search_term': 'Active'
+    },
+    'APPROVED UNDER RECONSIDERATION': {
+        'definition': 'Item was approved when considered a 2nd time.',
+        'search_term': None
```
hancush, line 307: Wow this format (and your table in the About page) makes it so much easier to parse! I think we can map each of these "X under reconsideration" actions to the status X, e.g., "Approved under reconsideration" is "Approved", "Failed under reconsideration" is "Failed", and so forth. Make sense?

lametro/templates/partials/board_report_statuses.html
```diff
@@ -0,0 +1,33 @@
...
+            <th>Status</th>
+            <th class="desktop-only">Description</th>
+        </tr>
+    </thead>
+    <tbody>
+    {% for status, obj in BILL_STATUS_DESCRIPTIONS.items %}
+        <tr>
+            <td>
+                <p class='nowrap'>
+                    {% if obj.search_term != None %}
```
hancush, line 13: [PEP8](https://www.python.org/dev/peps/pep-0008) (the Python style guide) offers this guidance on comparisons to None:

> Comparisons to singletons like None should always be done with is or is not, never the equality operators.

> Also, beware of writing if x when you really mean if x is not None -- e.g. when testing whether a variable or argument that defaults to None was set to some other value. The other value might have a type (such as a container) that could be false in a boolean context!

tl;dr -

1. Use `is` or `is not` when comparing to None
2. You really only need to check for `None` explicitly if it means something different than another possible [Falsy](https://www.freecodecamp.org/news/truthy-and-falsy-values-in-python/) value, like False or an empty string. 

Here, any Falsy value means we don't have a corresponding search term, so I would recommend simplifying like this:

```suggestion
                    {% if obj.search_term %}
```

hancush (changes requested):
Looking good, @fatima3558! That dictionary really helps to clean up that `_status` method. A few things inline. Please prioritize investment time today!

lametro/models.py
fatima3558: my understanding, based on the logic of the previous _status method, was that the status should be None (line 158 in the previous version, line 147 in the updated code). did I miss something?

hancush: More precisely, what happens if you try to access a key that does not exist in a dictionary in this way?

fatima3558: oh, gotcha, it'll throw an error rather than considering the condition false. I'll revise the logic. thanks!

lametro/models.py
```diff
@@ -141,21 +142,9 @@ def inferred_status(self):
...
-                return 'Active'
-            elif description.upper() in ['RECEIVED', 'RECEIVED AND FILED']:
-                return 'Received'
-            elif description.upper() == 'FAILED':
-                return 'Failed'
-            elif description.upper() == 'WITHDRAWN':
-              return 'Withdrawn'
-        else:
-            return None
+        if description and description.upper() in BILL_STATUS_DESCRIPTIONS.items():
```
hancush, line 145: `dict.items()` is an iterable of the key-value pairs in the dictionary, so I'm not sure this will work as you expect. You can do `key in dict.keys()` or `dict.get(key)`. TIL `key in dict.keys()` is faster!

hancush (changes requested):
One small thing inline, then this should be good to go.

lametro/models.py
fatima3558: I did test this in a python shell with a simplified dictionary, but using `keys()` is more readable, I think. made the change, thanks!

hancush (approved):

Issue #495, opened 2019-09-14
Only suggest topics associated with at least one board report


hancush:
Regardless of whether the query is the preferred or alternative term, we should only suggest topics if they're associated with board reports.

hancush:
Update: I tried to accomplish this using the `minDocs` filter, which I took to be an indication of how many documents were tagged with a term in SmartLogic. Per Steve, that's not quite right:

> Mindocs is a filter where we could indicate how many resources in a system like Legistar had that concept attached. This value had to be periodically calculated from the source system and set for each concept. For Metro they’re all zero so not helpful.

He suggested that we might perform and store this count on our end, but in my view, it makes a lot more sense for it to be performed at the SmartLogic level, since their system is already set up to ingest this data. 

@shrayshray, Steve mentioned that he'd need to extend the SmartLogic-Legistar bridge for this count to be accurate, and that he'd talk to you all about it. Once it's in place, this will be a trivial change on our end, but until then, I'd unfortunately consider us blocked.

shrayshray:
@hancush would it be possible to do something like query the Legistar API each night to get a list of all terms used in MatterIndexes (exclude duplicates) and use that list to check and exclude those which are not used?

hancush:
@shrayshray It would be possible for us to do that, however if the SmartLogic-Legistar bridge is already set up to store that information and already does a nightly sync with Legistar, it would make more sense for that operation to happen at the SmartLogic level. Then, we could make a tiny change on our end to exclude unused indexes from suggestions. 

With all that said, I think it would good idea to have a call with you and Steve to discuss options and figure out what makes most sense and will be most cost effective for you all.

hancush:
Look into adding indication of how frequently a suggested term is used (or even ordering suggestions by frequency). 

Goal: Minimize dead end suggestions, without preventing legitimate searching.

hancush:
@shrayshray @camachoo I've staged this improvement. You can compare suggestions (from the search box) and related terms (on the search results page) between the staging site and the live site.

Here's an example using my go-to term, "red".

Original (live site):

![Screen Shot 2022-02-07 at 2 55 29 PM](https://user-images.githubusercontent.com/12176173/152870519-d58fda3c-ffb7-41e0-a574-2bb70d698f4b.png)
![Screen Shot 2022-02-07 at 2 56 30 PM](https://user-images.githubusercontent.com/12176173/152870597-899805f1-5a70-4606-878d-34fc670ece80.png)

[...]

PR #771, opened 2021-11-12
Improve search suggestions

hancush:
## Overview

This PR improves search suggestions in two ways:

1. Only suggests topics associated with at least one board report
2. Suggests the most frequently used topics first

To achieve this, I made a number of changes:

1. Added an explicit relationship between board reports and `LAMetroSubject` objects associated with them, as well as a cached count of board reports associated with each subject
2. Updated `refresh_guid` to create these relationships and update the cached document count
3. Refactored the SmartLogic integration into a separate app, allowing access to raw SmartLogic responses for the endpoints we use in the Metro app
4. Extended the views from the SmartLogic app to provide filtered and ordered results in a consistent format for use in the Metro app
5. Refactored `autocomplete.js`, moving complicated result processing out of JavaScript and into the extended views from point 4 and simplifying the rendering of suggestions.

### Checklist

### Demo

tk

## Testing Instructions

 * tk

Handles #495


hancush:
Stale.

Issue #399, opened 2019-02-12
Hidden bills and events raise Internal Server Error, rather than Page Not Found


reginafcompton:
The Metro Councilmatic database includes private bills (marked with "restrict_view=True") and test events. Trying to access one of these entries raises an Internal Server Error. 

[Example of bill](https://sentry.io/datamade/lametro-councilmatic/issues/883897462/?referrer=webhooks_plugin)
[Example of event](https://sentry.io/organizations/datamade/issues/914360896/?project=137567&referrer=alert_email&statsPeriod=14d)

It would be better to handle this more gracefully, i.e., with a Page Not Found (404). 

_I'd classify this as a very low priority issue._

hancush:
I had to fix this issue on another app this morning! The problem was that I put my 404.html and 500.html templates in a directory like templates/app_name/404.html, but I needed to tell Django to look for them there by adding templates/app_name to my TEMPLATES > DIRS setting: https://github.com/confluency/stormstore-app/blob/master/stormstore_app/settings.py#L94

Maybe that's at play here?

hancush:
@fatima3558 Update: I don't think my original comment is the problem, as our 404 page _does_ work sometimes, e.g., #747. Let's deprioritize this, for now, and if it stays in the backlog after our meetings next week, let's pair on getting the app running with DEBUG = False and louder logging so you can test locally!

PR #776, opened 2021-11-29
Hidden bills display 404 instead of 500

fatima3558:
## Overview
This PR defaults to the DetailView `get_object` method, which uses a 404 if an object does not exist, and overwrites the `get_queryset` method to include media links in the queryset.

[WIP] It also will implement tests for the event and bill models to make sure private bills and events display 404 pages.

### Checklist

## Testing Instructions

 * Make sure CI passes

Handles #399

### DIFF
--- a/tests/test_bills.py
+++ b/tests/test_bills.py
@@ -21,7 +21,7 @@ def test_bill_url(client, bill):
     This test checks that the bill detail view returns a successful response.
     '''
     bill = bill.build()
-    url = reverse('bill_detail', kwargs={'slug': bill.slug})
+    url = reverse('lametro:bill_detail', kwargs={'slug': bill.slug})
     response = client.get(url)
 
     assert response.status_code == 200
--- a/tests/test_bills.py
+++ b/tests/test_bills.py
@@ -15,6 +15,7 @@ from lametro.models import LAMetroBill
 from lametro.utils import format_full_text, parse_subject
 
 # This collection of tests checks the functionality of Bill-specific views, helper functions, and relations.
+@pytest.mark.django_db
 def test_bill_url(client, bill):
     '''
     This test checks that the bill detail view returns a successful response.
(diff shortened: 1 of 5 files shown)


tests/test_bills.py
```diff
@@ -14,14 +14,13 @@
...
 
-
 # This collection of tests checks the functionality of Bill-specific views, helper functions, and relations.
 def test_bill_url(client, bill):
     '''
     This test checks that the bill detail view returns a successful response.
     '''
     bill = bill.build()
-    url = reverse('bill_detail', kwargs={'slug': bill.slug})
+    url = reverse('lametro:bill_detail', kwargs={'slug': bill.slug})
```
fatima3558, line 23: This test is now failing; I took a look at the test fixture, but the `restrict_view` property is set to `False`, so I'm not sure why we're getting a 404 here. Haven't dug deep into this yet, though.

tests/test_events.py
```diff
@@ -415,3 +415,18 @@ def test_delete_event(event, client, admin_client):
...
+
+
+# test event (event location name = TEST; get 404)
+def test_private_event(client, event, mocker):
+    private_event = event.build()
+
+    mock_location = mocker.MagicMock(spec=EventLocation)
+    mock_location.name = 'TEST'
+
+    mocker.patch('lametro.models.LAMetroEvent.location', return_value=mock_location)
```
fatima3558, line 427: I tried mocking the location here, and it seems to be working, but in remembering the rule to patch where the property is called, there didn't seem to be a good place in the code to patch this single event so that it has the mocked properties. The best place I found to patch would be in the EventManager's `get_queryset` method, but patching over a whole queryset doesn't seem right. Need to rethink my approach here.
https://github.com/datamade/la-metro-councilmatic/blob/24cb5635b6958ade2bd9ded2c041dc2b64c1e4ee/lametro/models.py#L393-L396

fatima3558: I'm going to try to create an `EventLocation` testing fixture that I could use to then give `private_event` a location property.

tests/test_bills.py
hancush: I'd recommend checking out [the filter logic in the Metro bill manager](https://github.com/datamade/la-metro-councilmatic/blob/24cb5635b6958ade2bd9ded2c041dc2b64c1e4ee/lametro/models.py#L61). You'll have to make sure the bill meets at least one of the criteria for being shown. I'd pick the easiest one. 😇

hancush (approved):
Heck yes, way to close a years-old issue, @fatima3558!

Issue #699, opened 2021-02-24
Handle external request failures more gracefully


hancush:
Infrequently, the current meetings request to Legistar is reset or otherwise fails, causing the homepage to return a server error. 

```
ConnectionResetError: [Errno 104] Connection reset by peer
  File "urllib3/connectionpool.py", line 600, in urlopen
    chunked=chunked)
  File "urllib3/connectionpool.py", line 384, in _make_request
    six.raise_from(e, None)
  File "<string>", line 2, in raise_from
    
  File "urllib3/connectionpool.py", line 380, in _make_request
    httplib_response = conn.getresponse()
  File "http/client.py", line 1373, in getresponse
    response.begin()
  File "http/client.py", line 311, in begin
    version, status, reason = self._read_status()
  File "http/client.py", line 272, in _read_status
    line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
  File "socket.py", line 586, in readinto
    return self._sock.recv_into(b)
ProtocolError: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))
  File "requests/adapters.py", line 449, in send
    timeout=timeout
  File "urllib3/connectionpool.py", line 638, in urlopen
    _stacktrace=sys.exc_info()[2])
  File "urllib3/util/retry.py", line 368, in increment
    raise six.reraise(type(error), error, _stacktrace)
  File "urllib3/packages/six.py", line 685, in reraise
    raise value.with_traceback(tb)
  File "urllib3/connectionpool.py", line 600, in urlopen
    chunked=chunked)
  File "urllib3/connectionpool.py", line 384, in _make_request
    six.raise_from(e, None)
  File "<string>", line 2, in raise_from
    
  File "urllib3/connectionpool.py", line 380, in _make_request
    httplib_response = conn.getresponse()
  File "http/client.py", line 1373, in getresponse
    response.begin()
  File "http/client.py", line 311, in begin
    version, status, reason = self._read_status()
  File "http/client.py", line 272, in _read_status
    line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
  File "socket.py", line 586, in readinto
    return self._sock.recv_into(b)
ConnectionError: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))
  File "django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "django/core/handlers/base.py", line 115, in _get_response
[...]

PR #784, opened 2022-01-06
External request handling

fatima3558:
## Overview

This PR updates code to handle requests from external sites more gracefully and sets timeout limits for these requests.

### Checklist

## Testing Instructions

 * verify logic added makes sense
 * verify tests pass

Handles #699

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '2.4'
 services:
   app:
-    image: lametro:upgrade
+    image: ghcr.io/datamade/la-metro-councilmatic:master
     container_name: lametro
     restart: always
     build: .
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -552,21 +552,26 @@ class LAMetroEvent(Event, LiveMediaMixin, SourcesMixin):
         Hit the endpoint, and return the corresponding meeting, or an empty
         queryset.
         '''
-        running_events = requests.get('http://metro.granicus.com/running_events.php')
 
-        for guid in running_events.json():
-            # We get back two GUIDs, but we won't know which is the English
-            # audio GUID stored in the 'guid' field of the extras dict. Thus,
-            # we iterate.
-            #
-            # Note that our stored GUIDs are all uppercase, because they come
-            # that way from the Legistar API. The running events endpoint
-            # returns all-lowercase GUIDs, so we need to uppercase them for
-            # comparison.
-            meeting = cls.objects.filter(extras__guid=guid.upper())
+        try:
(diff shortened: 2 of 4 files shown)


fatima3558:
the test that's failing is expecting a mocked response to have a `status_code` property. I'll update the test to give the mock request that property

docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: '2.4'
 services:
   app:
-    image: lametro:upgrade
+    image: ghcr.io/datamade/la-metro-councilmatic:master
```
hancush, line 4: You love to see it!

lametro/models.py
```diff
@@ -552,21 +552,26 @@ def _streaming_meeting(cls):
...
-            # comparison.
-            meeting = cls.objects.filter(extras__guid=guid.upper())
-
-            if meeting:
-                return meeting
+
+        try:
+            running_events = requests.get('http://metro.granicus.com/running_events.php', timeout=5)
+        except requests.exceptions.ConnectTimeout:
+            return cls.objects.none()
```
hancush, line 559: Yes, love the early return!

hancush (commented):
Looking good. Can you propose a new test confirming that `current_meeting` returns the expected value when the live meeting endpoint resolves with a non-200 status code?

fatima3558:
I can model it off of this test:

https://github.com/datamade/la-metro-councilmatic/blob/57b1dbc45e7b78f7aa272fd5b7d4bd022850beb9/tests/test_events.py#L124-L142

I would set the mock_request.status_code to be 404, keep the patch the same without the return_value (since this is what the assertion would test), remove the `current_meeting` references, and then assert that what is returned is an empty queryset.

fatima3558:
@hancush I was referring to lines 136 and 139. The pseudocode would look something like this:

 def test_meeting_stream_link_unavailable(mocker): 
     mock_response = mocker.MagicMock(spec=requests.Response) 
     mock_response.status_code = 404
  
     mocker.patch('lametro.models.requests.get')
  
     assert empty queryset returned

fatima3558:
as I'm working through this, I'm actually not sure what the assertion should be; I can't assert that I'm getting an empty queryset back because this would be part of the response object, which I'm currently mocking; based on what I should be asserting, the structure of the test will change. Do you have suggestions about what should I base my assertion/s off of, @hancush ?

hancush:
@fatima3558 In general, that looks good. For your commented line, we do want to test that, if the streaming meetings request fails, the `current_meeting()` method returns an empty queryset. Does that make sense?

hancush (approved):
Nice job, @fatima3558, thank you!

Issue #770, opened 2021-11-12
New page: Minutes


shrayshray:
Because Minutes are the official record of Board meetings, we'd like to have a page dedicated to them so it's easier for users to get to them will minimal scrolling. Could you let us know an estimate of hours to create a new page for Minutes? Here is some context for this request:
1. We would like the new page to be included in the navigation at the top of the website, after "Board Reports" and before "Archive Search" 
![image](https://user-images.githubusercontent.com/28063988/141213136-fbfddf7d-7f81-4b5d-a968-7b40f061e10f.png)

2. Once we get it from our people, we will provide a spreadsheet with dates and links to Minutes from prior meetings.
3. Logic for adding links to the page for future meeting Minutes would be required.

shrayshray:
[https://app.zenhub.com/files/70518566/98125240-70d8-4e17-b593-04b1d4dbeb76/download](https://app.zenhub.com/files/70518566/98125240-70d8-4e17-b593-04b1d4dbeb76/download)

fatima3558:
proposed steps:
- [ ] identify any duplicate historical events that are currently searchable on the metro site
- [ ] import csv data based on risk and reach [utility method](https://github.com/datamade/risk-and-reach/blob/b5306f49a88804f55d70ee4bad5ac05e964d4b32/risk_and_reach/utils.py#L5) and [view](https://github.com/datamade/risk-and-reach/blob/b5306f49a88804f55d70ee4bad5ac05e964d4b32/risk_and_reach/views.py#L47)
- [ ] create a new view and list template that looks similar to the [events page on the metro site](https://boardagendas.metro.net/events/)

fatima3558:
@shrayshray it looks like there are several events in the spreadsheet that are already in our database. From March 30, 2015 through the end of the spreadsheet, all events are currently searchable on the metro site except for the Special Board Meeting on May 21, 2015.

Wanted to check in about this information -- is the information in this spreadsheet different than the information we scrape from legistar?

hancush:
@fatima3558 looking good! may i suggest splitting the last step into two? 

step 3a. created a view that lists all events and their minutes, both from the csv and the database
step 3b. implement the event filter ui (whether that's recycling the date picker from the event page or doing a simpler year picker)

thoughts?

shrayshray:
[...]

PR #787, opened 2022-01-25
Minutes Page

fatima3558:
## Overview

This PR adds a Minutes page to the navigation. It incorporates some historical events from a csv file in the Minutes page table and includes a UI to filter the events by date.

### Checklist

## Testing Instructions

 * Visit the staging site and click on the new "Minutes" link in the nav bar
 * Make sure the new Minutes page looks good. Try using the date filter and all the buttons associated with that filter

Handles #770


fatima3558:
@hancush quick update: made some final updates to the template (collapsing the list & fixing links), have the date filter ui in place but working on finalizing functionality. In a previous commit, I added logic to group events by date, and I want to try to reimplement that with this updated code.

lametro/views.py
```diff
@@ -734,6 +736,100 @@ class LAMetroArchiveSearch(TemplateView):
...
+                minutes_link = event.documents.get(note__icontains='minutes').links.first().url
+                stored_events_dict['minutes_link'].append(minutes_link)
+            except EventDocument.DoesNotExist:
+                pass
+
+            try:
+                agenda_link = event.documents.get(note__icontains='agenda').links.first().url
+                stored_events_dict['agenda_link'].append(agenda_link)
+            except EventDocument.DoesNotExist:
+                pass
```
hancush, line 790: This loop is going to execute a query for every event in the database, which will add up really quickly! I think this would be a good use case for [the raw method](https://docs.djangoproject.com/en/4.0/topics/db/sql/#adding-annotations):

```python
        # Write a query to retrieve the objects you want, plus any annotations (in this case, minutes and agenda links)
        query = '''
            with agenda_links as (
                select
                    ed.event_id,
                    edl.url
                from opencivicdata_eventdocument ed
                join opencivicdata_eventdocumentlink edl
                on ed.id = edl.document_id
                where ed.note ilike '%%agenda%%'
            ), minutes_links as (
                select
                    ed.event_id,
                    edl.url
                from opencivicdata_eventdocument ed
                join opencivicdata_eventdocumentlink edl
                on ed.id = edl.document_id
                where ed.note ilike '%%minutes%%'
            )
            select
                ce.event_id,
                e.name,
                e.start_date,
                al.url as agenda_link,
                ml.url as minutes_link
            from opencivicdata_event e
            join councilmatic_core_event ce
            on e.id = ce.event_id
            left join agenda_links al
            on e.id = al.event_id
            left join minutes_links ml
            on e.id = ml.event_id
            where ce.event_id in %s
        '''

        # Execute the query, passing in a tuple of event IDs within the specified time frame as a value for the filter parameter
        stored_events = LAMetroEvent.objects.raw(
            query, [tuple(stored_events.values_list('event_id', flat=True))]
        )
```

This approach will reduce the number of queries from the number of events in the database to 2: one to get the IDs of events within the time range, and another to get the document links for those events.

lametro/views.py
```diff
@@ -734,6 +736,100 @@ class LAMetroArchiveSearch(TemplateView):
...
+
+            filtered_db_events.append(stored_events_dict)
+
+        # filter events from csv based on start and end dates
+        csv_events = get_list_from_csv('historical_events.csv')
+        events_dicts = [dict(e) for e in csv_events]
```
hancush, line 796: What is `e` before you cast it to a `dict`?

hancush (changes requested):
Thanks for putting this together, @fatima3558! I see the most room for improvement in the view code. See my comments inline, and let me know if you'd like to pair on anything Monday during our sprint planning meeting!

lametro/views.py
fatima3558: try a prefetch method before raw method:

https://github.com/datamade/la-metro-councilmatic/pull/406

lametro/views.py
fatima3558: it's an ordered dict, so it looks like a list of 2 tuples with what end up becoming key/value pairs

hancush: Got it. While its representation in the console looks different, an OrderedDict [behaves just like a regular dictionary](https://docs.python.org/3/library/collections.html#collections.OrderedDict), only its order is guaranteed (vs. a standard dict, which is unordered) and it has a few additional methods to manipulate the order. So you don't need to cast `e` to dict here!

lametro/views.py
```diff
@@ -734,6 +736,115 @@ class LAMetroArchiveSearch(TemplateView):
...
+        return filtered_historical_events
+
+    def _get_stored_events(self, start_datetime=None, end_datetime=None):
+        minutes = EventDocument.objects.filter(note__icontains='minutes').prefetch_related(Prefetch('links', to_attr='prefetched_links'))
+        agenda = EventDocument.objects.filter(note__icontains='agenda').prefetch_related(Prefetch('links', to_attr='prefetched_links'))
+        all_events = LAMetroEvent.objects.prefetch_related(Prefetch('documents', queryset=minutes, to_attr='minutes_document'))\
+                                         .prefetch_related(Prefetch('documents', queryset=agenda, to_attr='agenda_document'))
+        if start_datetime:
+            stored_events = all_events.filter(start_time__gt=start_datetime)\
+                                      .order_by('start_time')
```
hancush, line 777: Do we need to order this queryset if we're sorting the events together?

lametro/views.py
```diff
@@ -734,6 +736,115 @@ class LAMetroArchiveSearch(TemplateView):
...
+            start_date_str = self.request.GET['minutes-from'] or None
+            if start_date_str:
+                start_datetime = datetime.datetime.strptime(start_date_str, '%m/%d/%Y')
+        if 'minutes-to' in self.request.GET:
+            end_date_str = self.request.GET['minutes-to'] or None
+            if end_date_str:
+                end_datetime = datetime.datetime.strptime(end_date_str, '%m/%d/%Y')
+
+        context['start_date'] = start_date_str
+        context['end_date'] = end_date_str
```
hancush, line 828: If these are None, they show None in the date pickers.

Screen Shot 2022-02-23 at 3 11 27 PM

 How about an empty string if they aren't set?

lametro/templates/lametro/minutes.html
```diff
@@ -0,0 +1,104 @@
...
+              <input type="text" id="minutes-from" name="minutes-from" class="form-control date-filter" placeholder="Select start date..." value='{{ start_date }}'>
+              <input type="text" id="minutes-to" name="minutes-to" class="form-control date-filter" placeholder="Select end date..." value='{{ end_date }}'>
+              <span class="input-group-btn">
+                <button class="btn btn-default btn-date" id="minutes-search" type="submit"><i class="fa fa-search" aria-hidden="true"></i> <span class='hidden-sm hidden-xs'>Search</span></button>
+              </span>
+            </div>
+          </form>
+        </div>
+        <div class="col-xs-2">
+          <a href="{% url 'lametro:minutes' %}?show=all" class="btn btn-salmon d-inline-block"><i class="fa fa-list" aria-hidden="true"></i><span class="hidden-xs"> All minutes</span></a>
```
hancush, line 30: Hm, clicking this doesn't seem to show all events. Can you take a look? 
Screen Shot 2022-02-23 at 3 11 57 PM

lametro/views.py
hancush: Ah, the None also breaks the date search! 

```
Environment:


Request Method: GET
Request URL: http://localhost:8001/minutes/?minutes-from=None&minutes-to=02%2F01%2F2022

Django Version: 2.2.25
Python Version: 3.6.14
Installed Applications:
('django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django.contrib.gis',
 'haystack',
 'opencivicdata.core',
 'opencivicdata.legislative',
 'lametro',
 'councilmatic_core',
 'adv_cache_tag',
 'debug_toolbar',
 'captcha')
Installed Middleware:
('debug_toolbar.middleware.DebugToolbarMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'django.middleware.security.SecurityMiddleware')



Traceback:

File "/usr/local/lib/python3.6/site-packages/django/core/handlers/exception.py" in inner
  34.             response = get_response(request)

File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py" in _get_response
  115.                 response = self.process_exception_by_middleware(e, request)

File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py" in _get_response
  113.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "/usr/local/lib/python3.6/site-packages/django/views/generic/base.py" in view
  71.             return self.dispatch(request, *args, **kwargs)

File "/usr/local/lib/python3.6/site-packages/django/views/generic/base.py" in dispatch
  97.         return handler(request, *args, **kwargs)

File "/usr/local/lib/python3.6/site-packages/django/views/generic/list.py" in get
  157.         context = self.get_context_data()

File "/app/lametro/views.py" in get_context_data
  821.                 start_datetime = datetime.datetime.strptime(start_date_str, '%m/%d/%Y')

File "/usr/local/lib/python3.6/_strptime.py" in _strptime_datetime
  565.     tt, fraction = _strptime(data_string, format)

File "/usr/local/lib/python3.6/_strptime.py" in _strptime
  362.                          (data_string, format))

Exception Type: ValueError at /minutes/
Exception Value: time data 'None' does not match format '%m/%d/%Y'
```

I think an empty string should resolve this. 👍

lametro/views.py
```diff
@@ -734,6 +736,115 @@ class LAMetroArchiveSearch(TemplateView):
...
     template_name = 'lametro/contact.html'
 
+
+class MinutesView(EventsView):
+    template_name = 'lametro/minutes.html'
+
+    def _get_historical_events(self, start_datetime=None, end_datetime=None):
+        csv_events = get_list_from_csv('historical_events.csv')
+
+        events_dicts = [dict(e) for e in csv_events]
```
hancush, line 746: You don't need to cast to a dict here, since `get_list_from_csv` returns a list of ordered dicts. You can just work with `csv_events` directly.

hancush (changes requested):
Big improvement, @fatima3558! That Prefetch, tho!!! 🔥 A few more things for your consideration inline.

lametro/templates/lametro/minutes.html
fatima3558: oh shoot, missed this piece, will address & test and let you know when it's ready for your eyes

fatima3558: I'm addressing this by removing the query params, but then this link is the same as the reset button's link. I'm not sure whether keeping this interface the same as the events page is more important than avoiding the redundancy of having the same link twice. What do you think @hancush ?

hancush: You know, that's a great point! Why don't we just remove the All minutes link from the top? We're already loading everything in, and it doesn't make a lot of sense from a UI perspective to have a button at the top of the page expand an element at the bottom of the page, which isn't visible.

hancush (approved):
Thanks so much for your persistence here, @fatima3558! This is ready to come in.

Issue #798, opened 2022-02-18
BOARD MEMBERS: Add note on pending redistricting


hancush:
@shrayshray Can you provide some text about ongoing redistricting that we can put on the board members page? 🙏

shrayshray:
@hancush On December 15, 2021, the County of Los Angles Citizens Redistricting Commission passed resolutions to adopt a final redistricting [plan](https://redistricting.lacounty.gov/wp-content/uploads/2021/12/LA-County-CRC-Resolution-Adopting-Redistricting-Plan.pdf) and [report](https://redistricting.lacounty.gov/wp-content/uploads/2021/12/LA-County-CRC-Resolution-Adopting-Redistricting-Report.pdf). Click [here](https://redistricting-lacounty.hub.arcgis.com/) to view the final redistricting plan and all other submitted maps in the Redistricting Hub. The map on this page will be updated once the County provides finalized maps.

hancush:
@shrayshray We've added a note to the staging site: https://lametro-upgrade.datamade.us/board-members/

Look ok?

shrayshray:
@hancush looks great, thank you!

hancush:
Done and deployed!

PR #802, opened 2022-02-23
Added redistricting alert to Board of Directors Page

antidipyramid:
## Overview

Adds an alert to the top of the Board of Directors page notifying users of changes due to redistricting.

### Checklist

### Demo

Screen Shot 2022-02-23 at 11 37 27 AM

### Notes

`django-debug-toolbar` seemed to have been causing some performance issues during development so we also pared down its displayed panels to only SQL and Cache.

## Testing Instructions

 * Visit the Board of Directors page and verify that the alert is displayed, the links are valid, and the alert is responsive.

Handles #798

### DIFF
--- a/lametro/templates/lametro/board_members.html
+++ b/lametro/templates/lametro/board_members.html
@@ -27,7 +27,6 @@
 </div>
 
 <div class="container-fluid">
-
     {% if MAP_CONFIG %}
     <div class="row">
         <div class="col-sm-6">
--- a/lametro/templates/lametro/board_members.html
+++ b/lametro/templates/lametro/board_members.html
@@ -20,6 +20,14 @@
 {% cache 86400 members_wrapper 'members' %}
 
 <div class="container-fluid">
+    <br>
+    <div class="alert alert-danger mt-3" role="alert">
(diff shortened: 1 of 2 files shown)


lametro/templates/lametro/board_members.html
```diff
@@ -19,6 +19,13 @@
 
 {% cache 86400 members_wrapper 'members' %}
 
+<div class="container-fluid">
+    <br>
+    <div class="alert alert-danger mt-3" role="alert">
+        On December 15, 2021, the County of Los Angles Citizens Redistricting Commission passed resolutions to adopt a final redistricting <a href='https://redistricting.lacounty.gov/wp-content/uploads/2021/12/LA-County-CRC-Resolution-Adopting-Redistricting-Plan.pdf'>plan</a> and <a href='https://redistricting.lacounty.gov/wp-content/uploads/2021/12/LA-County-CRC-Resolution-Adopting-Redistricting-Report.pdf'>report</a>. Click <a href='https://redistricting-lacounty.hub.arcgis.com/'>here</a> to view the final redistricting plan and all other submitted maps in the Redistricting Hub. The map on this page will be updated once the County provides finalized maps.
```
antidipyramid, line 25: @hancush Are there any particularly standards we want to adhere to when it comes to line length in HTML?

lametro/templates/lametro/board_members.html
```diff
@@ -19,6 +19,13 @@
 
 {% cache 86400 members_wrapper 'members' %}
 
+<div class="container-fluid">
+    <br>
+    <div class="alert alert-danger mt-3" role="alert">
```
hancush, line 24: What do you think about using a less severe color scheme here. Maybe `alert-info`?

lametro/templates/lametro/board_members.html
hancush, line 25: These links are a bit hard to read in the red. If they're still a bit subdued with the change in color scheme, how about bolding them?

hancush: Good question! I try to break up lines at the 80-char mark.

hancush (changes requested):
Looking good, @antidipyramid! Very small items inline.

lametro/templates/lametro/board_members.html
antidipyramid: Here's what `alert-info` looks like:
Screen Shot 2022-02-23 at 2 47 13 PM

I think it looks good. The links also are obvious, color-wise. Do we want to underline them also?

hancush (approved):
This looks good to me!

Issue #785, opened 2022-01-11
Reorganize nav


fatima3558:
Due to the addition of the 'Contact Us' page in the navigation, the navigation bar is wrapping and taking up more vertical space than intended on certain window widths:
<img width="1202" alt="Screen Shot 2022-01-11 at 12 36 58 PM" src="https://user-images.githubusercontent.com/48227264/149001599-6a392801-dc9d-4ea8-a79c-c2a22868561b.png">

@shrayshray We can group some of these nav items into sub-menus or potentially move some of these items so that they only live in the footer, for example. Would either of these, or another, approach work for you?

hancush:
Ideas from 1/18 priorities:

About
  - Board of Directors
  - Committees
  - Contact Us
Search
  - Board Reports
  - Archive Search
Meetings and Agendas
  - Minutes

Shelly and Omar to get internal feedback and let us know.

hancush:
Shelly to follow up on this in Feb. Let us know when you hear back, @shrayshray!

shrayshray:
@hancush feedback received! They like this grouping:

- About 
Board of Directors
Committees
Contact Us
About us [Change name of current About page]

- Search 
Board Reports
Archive Search

- Meetings 
Meetings and Agendas 
Meeting Minutes

antidipyramid:
@shrayshray New navbar is up on the [staging site](https://lametro-upgrade.datamade.us/)! How does it look?

PR #804, opened 2022-03-02
Reorganize navbar

antidipyramid:
## Overview

Condenses navbar items using dropdowns and fixes responsiveness of the bar.

### Checklist

### Demo

navbar

### Notes

A section of `city_custom.css` was removed to change the navbar's breakpoint.

## Testing Instructions

 * Navigate to the site and verify that the navbar has been condensed and the dropdowns render correctly.
 * Verify that the navbar is responsive.

Handles #785

### DIFF
--- a/lametro/static/css/city_custom.css
+++ b/lametro/static/css/city_custom.css
@@ -331,21 +331,6 @@ hr .events-line {
     }
 }
 
-@media (min-width: 767px) and (max-width: 1180px) {
-    .navbar-collapse.collapse {
-        display: none !important;
-    }
-    .navbar-collapse.collapse.in {
-        display: block !important;
-    }
-    .navbar-header .collapse, .navbar-toggle {
-        display:block !important;
-    }
-    .navbar-header {
-        float:none;
-    }
-}
-
 .badge-highlight {
     background-color: #d9edf7;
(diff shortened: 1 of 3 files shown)


lametro/static/css/city_custom.css
```diff
@@ -331,21 +331,6 @@ hr .events-line {
...
-        display: block !important;
-    }
-    .navbar-header .collapse, .navbar-toggle {
-        display:block !important;
-    }
-    .navbar-header {
-        float:none;
-    }
-}
-
```
fatima3558, line 348: curious if you looked to see where else these classes were being used // if elements affected by these had their behavior changed by removing this styling?

antidipyramid: I believe this block was just used to set the breakpoint for the navbar at a larger window size than bootstrap's default. Most likely because the navbar had so much on it.

It's a media rule plus only the navbar is styled here so I don't think anything else is affected. But if you saw any weird behavior, let me know 🙏🏾

antidipyramid: @fatima3558 Anything you want me to modify before I get your approval?

hancush (approved):
@antidipyramid This LGTM!

Issue #790, opened 2022-02-02
HOMEPAGE: Add recently passed meetings to the homepage


hancush:
![Screen Shot 2022-02-02 at 3.37.31 PM.png](https://images.zenhubusercontent.com/5b0dc8d3348b7966b4340de0/b3e9e9f4-8de6-41df-8118-4f50e744072b)

Add a bit more visual distinction between blocks. @shrayshray, also let us know if you have a better title than "Recently Passed Meetings" and if you all want to add any explainer text to either section.

hancush:
Show meetings from the last two weeks, provided they're in the current calendar month. Hide section if no meetings within that window.

hancush:
Add a new section, "Recent Meetings", to the homepage, below the "Upcoming Meetings" section.

This section should contain links to meetings that occurred in the last two weeks of the current calendar month. (That means, on March 1, even though there were meetings the last two weeks of February, this section should be empty.)

Consistent with `todays_meetings` and `upcoming_committee_meetings`,  implement a classmethod, `most_recent_past_meetings`, on the `LAMetroEvent` model that returns meetings within the specified time frame, or an empty queryset. (FYI, the naming convention for the proposed classmethod comes from [a similar one upstream in `django-councilmatic` called `most_recent_past_city_council_meeting`](https://github.com/datamade/django-councilmatic/blob/5b074f376667766e4b6dbf093871d294bb35fc51/councilmatic_core/models.py#L373).)

This section should be styled similarly to "Upcoming Meetings", but with some visual distinction, e.g., a different icon and/or button color, a horizontal rule, etc. You have creative liberty with the distinction, so long as it's cohesive with the rest of the site.

Metro has an extensive test suite. Please also add a test for your new classmethod. @fatima3558 has some experience with this pattern from her work on `todays_meetings`! The event tests also contain several examples of creating test events with different dates and mocking different times with [`freezegun`](https://pypi.org/project/freezegun/0.1.11/).

hancush:
Stage for @shrayshray and Omar to meet with folks after this month's meetings to review.

hancush:
Use concluded logic to make sure events don't show up in the new section until they have been broadcast.

shrayshray:
@hancush The broadcast has concluded for Planning and Programming, but it's not showing in "Recent Meetings" yet. Is there a time frame in which we can expect to see it there?

shrayshray:
[...]

PR #805, opened 2022-03-02
Add recently passed meetings to homepage

antidipyramid:
## Overview

Adds a section to the homepage that displays meetings that have occurred in the past two weeks within the current month. Also adds two unit tests.

### Checklist

### Demo

Screen Shot 2022-03-02 at 3 58 57 PM

## Testing Instructions

 * Verify that the new section appears and works correctly.
 * Run the tests.

Handles #790


fatima3558 (approved):
this looks great to me! request hannah's review here for final approval next week

lametro/templates/lametro/index.html
```diff
@@ -86,6 +86,29 @@ <h2>
...
         </div>
+        <div class="row-fluid">
+            <div class='col-sm-10 col-sm-offset-1'>
+                <h2>
+                    <span class="non-mobile-only"><i class="fa fa-calendar-check-o" aria-hidden="true"></i>
+</span>
+                    Recent Meetings
+                </h2>
+                {% if most_recent_past_meetings %}
+                    {% if most_recent_past_meetings %}
```
fatima3558, line 97: why is this conditional duplicated?

lametro/models.py
```diff
@@ -487,6 +487,21 @@ class LAMetroEvent(Event, LiveMediaMixin, SourcesMixin):
...
+        '''
+        Returns meetings in the current month that have occured in the past
+        two weeks.
+        '''
+        yesterday = datetime.today() - timedelta(days=1)
+        current_month = datetime.today().month
+        two_weeks_ago = datetime.today() - timedelta(weeks=2)
+
+        past_meetings = cls.objects.filter(
+            Q(start_time__month=current_month), Q(start_time__range=(two_weeks_ago, yesterday))).order_by('start_time')
```
hancush, line 501: Can you say more about ending the range yesterday? What happens if a meeting happed at 10 a.m., and now it's 1 p.m.?

hancush: BTW, TIL you can pass a range! So much nicer than `__gt` and `__lt`. Are both ends of the range inclusive?

lametro/templates/lametro/index.html
hancush: Ditto this question.

tests/test_events.py
```diff
@@ -338,6 +338,57 @@ def test_event_is_upcoming(event, mocker):
...
+        days=21).strftime('%Y-%m-%d %H:%M')
+    one_week_from_now = LAMetroEvent._time_from_now(
+        days=7).strftime('%Y-%m-%d %H:%M')
+    one_minute_from_now = LAMetroEvent._time_from_now(
+        minutes=1).strftime('%Y-%m-%d %H:%M')
+    eight_days_ago_last_month = LAMetroEvent._time_ago(
+        days=8).strftime('%Y-%m-%d %H:%M')
+
+    four_days_ago = LAMetroEvent._time_ago(days=4).strftime('%Y-%m-%d %H:%M')
+    five_days_ago = LAMetroEvent._time_ago(days=5).strftime('%Y-%m-%d %H:%M')
```
hancush, line 362: Maybe this is an insane request, 😝, but can you put these in chronological order?

tests/test_events.py
```diff
@@ -338,6 +338,57 @@ def test_event_is_upcoming(event, mocker):
...
+    event_one_week_from_now = event.build(
+        name='Board Meeting', start_date=one_minute_from_now, id=get_event_id())
+    event_last_month = event.build(
+        name='Board Meeting', start_date=eight_days_ago_last_month, id=get_event_id())
+
+    # Build events that should be in list
+    event_four_days_ago = event.build(
+        name='Board Meeting', start_date=four_days_ago, id=get_event_id())
+    event_five_days_ago = event.build(
+        name='Board Meeting', start_date=five_days_ago, id=get_event_id())
```
hancush, line 378: These, too? Maybe with consistent comments, like:

```
# Events that shouldn't be returned
...

# Events that should be returned
...
```

hancush (changes requested):
This looks great, @antidipyramid! Nice and concise. A couple of relatively small things inline.

lametro/templates/lametro/index.html
```diff
@@ -86,6 +86,29 @@ <h2>
...
+                        {% for event in most_recent_past_meetings %}
+                            {% include "partials/event_item.html" %}
+                        {% endfor %}
+                    {% endif %}
+                {% else %}
+                    <p><em>No meetings held in the past two weeks of this month.</em></p>
```
hancush, line 103:
```suggestion
                    <p><em>No meetings in the past two weeks of this month.</em></p>
```

lametro/models.py
antidipyramid: When I wrote this, I was assuming that we wouldn't want to display events from earlier in the current day. But IIRC we received clarification that we do want to include these events. I can modify this.

And yes, both ends of `range` are inclusive! From the Django docs:
Screen Shot 2022-03-09 at 3 33 14 PM

tests/test_events.py
antidipyramid: Of course!

lametro/templates/lametro/index.html
antidipyramid: It shouldn't be. Pretty sure it's just a mistake on my end 😅

lametro/models.py
```diff
@@ -487,6 +487,25 @@ class LAMetroEvent(Event, LiveMediaMixin, SourcesMixin):
...
+    def most_recent_past_meetings(cls):
+        '''
+        Returns meetings in the current month that have occured in the past
+        two weeks.
+        '''
+        current_month = timezone.now().month
+        two_weeks_ago = timezone.now() - timedelta(weeks=2)
+
+        meetings_in_past_two_weeks = cls.objects.filter(
+            Q(start_time__month=current_month), Q(start_time__gte=(two_weeks_ago)), Q())
```
hancush, line 500: Since you're not doing any fancy logic here (using OR, or programmatically building filters in a loop), you don't need to use `Q()` instances here – you can pass your keyword arguments directly to `filter()`.

lametro/models.py
```diff
@@ -487,6 +487,25 @@ class LAMetroEvent(Event, LiveMediaMixin, SourcesMixin):
...
+
+        # since display_status is a property of LAMetroEvent rather than
+        # a model attribute, we have to make sure returned meetings 
+        # have concluded separately from the above Queryset filter
+        past_meetings = list(filter(lambda m: m.display_status == 'Concluded', meetings_in_past_two_weeks))
```
hancush, line 505: Thoughts on using `has_passed` directly, rather than `display_status`, which uses it to determine whether a meeting has concluded?

tests/test_events.py
```diff
@@ -1,6 +1,7 @@
 import pytest
 import pytz
 import re
+import pdb
```
hancush, line 4: You can strike this, since we aren't using it anymore. :-)

tests/test_events.py
```diff
@@ -338,6 +339,64 @@ def test_event_is_upcoming(event, mocker):
...
+def test_first_of_month_returns_no_recent_past_meetings(event):
+    yesterday = LAMetroEvent._time_ago(days=1).strftime('%Y-%m-%d %H:%M')
+    yesterdays_event = event.build(
+        name='Board Meeting', start_date=yesterday, id=get_event_id())
+
+    assert len(LAMetroEvent.most_recent_past_meetings()) == 0
+
+
+@freeze_time("2021-02-07 10:00:00")
+def test_most_recent_past_meetings(event):
```
hancush, line 352: Sick test! 🤙

hancush (changes requested):
Nice second pass, @antidipyramid! A couple of very small things inline, then this should be good to go. :-)

lametro/models.py
antidipyramid: Good call.

lametro/models.py
```diff
@@ -487,6 +487,26 @@ class LAMetroEvent(Event, LiveMediaMixin, SourcesMixin):
...
+        '''
+        Returns meetings in the current month that have occured in the past
+        two weeks.
+        '''
+        current_month = timezone.now().month
+        two_weeks_ago = timezone.now() - timedelta(weeks=2)
+
+        meetings_in_past_two_weeks = cls.objects.filter(
+            start_time__month=current_month).filter(
+            start_time__gte=two_weeks_ago)
```
hancush, line 501: You can combine these keyword arguments into a single `filter()` call, like this: `filter(kwarg_1=foo, kwarg_2=bar)`

lametro/models.py
```diff
@@ -487,6 +487,26 @@ class LAMetroEvent(Event, LiveMediaMixin, SourcesMixin):
...
+
+        # since display_status is a property of LAMetroEvent rather than
```
hancush, line 503: Want to update this to avoid future confusion? 

```suggestion
        # since has_passed is a property of LAMetroEvent rather than
```

hancush (commented):
Two more tiny things! Can you squeeze them in, in the morning?

lametro/models.py
antidipyramid: Sure!

hancush (approved):
Well done!

Issue #799, opened 2022-02-18
Update implicated geographies when redistricting is finalized


hancush:
@shrayshray Can you all provide or point us to updated GIS files for the supervisory districts as soon as they become available?

shrayshray:
Are you able to pull the files from here? https://lavote.gov/apps/precinctsmaps It looks like they have been updated.

![image](https://user-images.githubusercontent.com/28063988/155409656-9cbad85a-c8e7-4ac8-98d6-c46c1efabd34.png)

hancush:
Hi, @shrayshray – I can get all but District 2. For some reason, I'm getting this error when I try to export it. 

<img width="549" alt="Screen Shot 2022-02-23 at 3 46 23 PM" src="https://user-images.githubusercontent.com/12176173/155414550-ee86b277-2ac3-4173-8f85-d770a6777c8e.png">

Are you able to save District 2?

hancush:
Sent an email to the Clerk's web team!

hancush:
Clerk's team provided the missing district. New GeoJSON file is here: https://gist.github.com/hancush/3cce4e7c6908e1f1d2fc911b1c3da485

Update the Make pipeline [the `data/` directory](https://github.com/datamade/la-metro-councilmatic/tree/master/data) to use the updated GeoJSON file, instead of trying to retrieve supervisory districts from the legacy API, to create a refreshed `boundaries.geojson` file. 

N.b., it is fine to continue using the other boundaries from the legacy API, as written. Let's truncate `data/README.md` starting at "You shouldn't need to rerun the Makefile, but keep it for historical reference...", since we will in fact be reusing this pipeline.

Next, run the management command specified in `data/README.md` to load the new shapes and confirm they are rendered correctly on the board members map.

@shrayshray Do you happen to have any examples of addresses that were previously in one supervisory district previously but is now in a different one? That would be really helpful for testing!

shrayshray:
@hancush Previously in District 4, now in District 2:
303 N Pacific Coast Hwy, Redondo Beach, CA 90277
605 1/2, 601 Virginia St, El Segundo, CA 90245
1100 Pacific Coast Hwy, Hermosa Beach, CA 90254

hancush:
@shrayshray Thanks to @antidipyramid, the new districts are live on the staging site! I've confirmed the addresses you listed are now showing up in District 2. Want to take a look and confirm everything looks above board on your end?

shrayshray:
[...]

PR #806, opened 2022-03-03
Update metro districts

antidipyramid:
## Overview

Adds in the new LA Metro boundaries that changed after redistricting.

### Checklist

#### Issues

Initially, I tried to just swap in the URL for the new boundaries (remotely stored in a secret Gist by @hancush) in the Makefile. However, this caused two issues:

1. The properties in the new boundaries do not match the properties in the GeoJSON files that the Makefile currently fetches from the DataMade OCD endpoint. In particular, the new boundaries use `LABEL` while the old files use `name`. This caused `add_district.py` to fail while labeling the features.
2. The new boundaries are structured as a nested `FeatureCollection`. The actual boundaries are inside `FeatureCollection` that is essentially empty. This caused `combine.py` to add an empty `FeatureCollection` to the final boundaries instead of the new Metro boundaries.

#### Current Solutions

I manually modified the new boundaries (stored as `supervisory_districts.geojson`) to work with the current scripts. Because the data pipeline now depends on a local file, inline with DataMade's ETL guidelines,  I added the directory `data/raw/` to store both `supervisory_districts.geojson` and the files fetched from the OCD endpoint. I also added `data/final/` to store the final output boundaries, `boundary.geojson`, that are displayed on the site. 

The new directories and use of `supervisory_districts.geojson` necessitated minor changes in the Makefile. Additionally, I added a `clean` command to clear the `data/final/` if we want to re-run the Makefile and generate `data/final/boundary.geojson`.

Both `supervisory_districts.geojson` and the data fetched from the OCD endpoint are stored in `data/raw/` but if it'd be more appropriate to separate those (and perhaps add a command to clean the fetched files), please let me know.

#### Docker

Currently, Docker isn't able to run `make`. For some reason, the container run by `docker-compose` doesn't have `make` installed. My suspicion is that this could be due some aspect of the Dockerfile. I attempted to modify it to get access to make but didn't have any luck. 

We can discuss whether or not we want Docker to be able to run `make` and how to get that done.

## Testing Instructions

 * Run `make clean` and then `make all` from the `data/` directory.
 * Import the new shapes with `docker-compose run --rm app python manage.py import_shapes data/final/boundary.geojson`
 * Verify that the new boundaries are live (might need to clear your cache)

Handles #799


fatima3558 (commented):
works as described! thanks for all the context in the PR description, @antidipyramid!

as far as your question about the `data/raw/` directory, in my experience we keep all initial versions of data in that directory, so I don't think we need to separate out the files gotten through `wget` from the files from the gist. If files need to go through a processing step between the initial file and the final version, we will add a `data/processed` directory, like in the cps repo.

@hancush I spent some time on Monday also trying to figure out the issue with running Make in the Docker container, but I also didn't get very far.

antidipyramid:
@hancush The separate container to run `make` was a great idea and worked perfectly. The Makefile can now be run with `docker-compose run --rm app make all`. I also designated the remote files as intermediate targets so they're downloaded, used, and removed each time we run `make`.

data/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM python:3.8-alpine
+
+LABEL maintainer "DataMade <info@datamade.us>"
+
+# gdal-tools contains ogr2ogr (needed by our scripts)
```
hancush, line 5: Great thought to include this comment.

hancush (approved):

Issue #811, opened 2022-03-16
Write tag analytics management command


hancush:
Write a new Django management command that creates a CSV file structured like this:

- File ID
- Title
- Tag (flat, one tag per line, a record for every tag associated with a board report)

PR #815, opened 2022-03-17
Add tag analytics management command

antidipyramid:
## Overview

Adds a django-admin command to produce a CSV file that links each board report to its tags. The CSV's fields are:

- File ID
- Identifier
- Title
- Last Action Date
- Tag Classification
- Tag

The file is saved to the root directory.

### Checklist

## Testing Instructions

 * Run `docker-compose run --rm app python manage.py generate_tag_analytics`
 * Examine the generated CSV file

Handles #811

### DIFF
--- a/lametro/management/commands/generate_tag_analytics.py
+++ b/lametro/management/commands/generate_tag_analytics.py
@@ -1,4 +1,5 @@
 import csv
+from tqdm import tqdm
 from django.core.management.base import BaseCommand
 
 from lametro.models import LAMetroBill
@@ -12,13 +13,19 @@ class Command(BaseCommand):
 
         with open('tag_analytics.csv', 'w', newline='') as f:
             writer = csv.writer(f, delimiter=',')
-            files = [(b.board_report.id,
-                      b.identifier,
-                      b.friendly_name,
-                      b.last_action_date,
-                      b.subject) for b in LAMetroBill.objects.all()]
+            files = []
+            for bill in LAMetroBill.all_objects.all():
+                if bill.board_report:
+                    files.append(
+                        (bill.board_report.id,
+                         bill.identifier,
+                         bill.friendly_name,
+                         bill.last_action_date,
+                         bill.subject)
+                    )
+
             writer.writerow(['File ID', 'Identifier', 'Title', 'Last Action Date', 'Tag'])
-            for report_id, identifier, title, last_action_date, tags in files:
+            for report_id, identifier, title, last_action_date, tags in tqdm(files):
                 for tag in tags:
                     writer.writerow([
                         str(report_id),
--- a/lametro/management/commands/generate_tag_analytics.py
+++ b/lametro/management/commands/generate_tag_analytics.py
@@ -1,0 +1,29 @@
+import csv
+from django.core.management.base import BaseCommand
+
+from lametro.models import LAMetroBill
+
+
+class Command(BaseCommand):
+
+    help = "This command produces a CSV file that lists each Board Report's tags."
+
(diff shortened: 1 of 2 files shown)


lametro/management/commands/generate_tag_analytics.py
```diff
@@ -0,0 +1,36 @@
...
+class Command(BaseCommand):
+
+    help = "This command produces a CSV file that lists each Board Report's tags."
+
+    def handle(self, *args, **options):
+
+        with open('tag_analytics.csv', 'w', newline='') as f:
+            writer = csv.writer(f, delimiter=',')
+            files = []
+            for bill in LAMetroBill.all_objects.all():
```
hancush, line 17: I see what you're doing here, but we only want this to return the board reports on the site, i.e., those returned by the LAMetroBillManager.

lametro/management/commands/generate_tag_analytics.py
```diff
@@ -0,0 +1,36 @@
...
+                if bill.board_report:
+                    files.append(
+                        (bill.board_report.id,
+                         bill.identifier,
+                         bill.friendly_name,
+                         bill.last_action_date,
+                         bill.subject)
+                    )
+
+            writer.writerow(['File ID', 'Identifier', 'Title', 'Last Action Date', 'Tag'])
```
hancush, line 27: Just had a thought: How about we also include the classification of each tag, so they can look at them in those buckets, as well.

lametro/management/commands/generate_tag_analytics.py
```diff
@@ -0,0 +1,36 @@
...
+
+class Command(BaseCommand):
+
+    help = "This command produces a CSV file that lists each Board Report's tags."
+
+    def handle(self, *args, **options):
+
+        with open('tag_analytics.csv', 'w', newline='') as f:
+            writer = csv.writer(f, delimiter=',')
+            files = []
```
hancush, line 16: This works but it isn't necessarily optimal, because it will build up this list in memory. Can you think of a way to write the file in this first loop, instead of building up the list and writing it in a second loop?

lametro/management/commands/generate_tag_analytics.py
hancush: (A hint: LAMetroBill instances have a `rich_topics` property that returns a queryset of LAMetroSubject instances corresponding to their subjects.)

hancush (changes requested):
Looking good, @antidipyramid! Command works and the output matches my expectations. Love that you used tqdm to show progress. 

Left a couple of comments inline.

lametro/management/commands/generate_tag_analytics.py
antidipyramid: Very true-- this is a very inefficient way to do this. Thanks for pointing this out.

hancush (approved):
So clean! Great refactor, @antidipyramid.

lametro/management/commands/generate_tag_analytics.py
```diff
@@ -0,0 +1,37 @@
...
+                 'Title',
+                 'Last Action Date',
+                 'Tag Classification',
+                 'Tag')
+            )
+
+            for bill in tqdm(LAMetroBill.objects.all()):
+                if not bill.board_report:
+                    continue
+
```
hancush, line 28: P.s., Do we need this check?

antidipyramid: Hm, I guess we don't since the instances returned by `LAMetroBillManager` should all have valid `board_report`s.

Issue #818, opened 2022-03-17
Home: Show "cancelled" not "concluded" for cancelled meetings in Today's Meetings


shrayshray:
The 3/17 Construction Committee meeting was cancelled. After the time the meeting was scheduled to have been held, the tag on it showed as "Concluded" in the "Today's Meetings" section. Since it was cancelled, it should show as cancelled instead.
![image](https://user-images.githubusercontent.com/28063988/158902213-d2df5918-e6b6-489d-a37a-21b845d93f28.png)

PR #822, opened 2022-03-21
Fix cancelled meetings label on Homepage

antidipyramid:
## Overview

In the "Today's Meetings" portion of the homepage, cancelled meetings earlier in the day were incorrectly displayed as "Concluded". This seems to be because the `LAMetroEvent.display_status()` checked whether or not an event has passed before checking if it had been cancelled.

### Checklist

### Notes

I did some refactoring in `todays_meetings.html`. While going through the code, I found it to be a little difficult to parse due to the many nested tags and conditionals. More code is repeated but I thought that what's gained in clarity was worth the increase in length.

But if we'd rather leave it in its previous format, I'm more than happy to change it back.

## Testing Instructions

 * Run the tests
 * Verify that cancelled meetings are labeled as 'Cancelled' rather than 'Concluded' on the homepage

Handles #818

### DIFF
--- a/tests/test_events.py
+++ b/tests/test_events.py
@@ -472,7 +472,6 @@ def test_display_status(event):
     assert cancelled_this_morning.display_status == 'Cancelled'
 
 
-
 def test_todays_meetings(event):
     # create event for some day
     event_time = app_timezone.localize(datetime(2020, 3, 15, 15, 0, 0, 0)) # March 15, 2020 at 3pm LA time
--- a/tests/test_events.py
+++ b/tests/test_events.py
@@ -458,15 +458,12 @@ def test_most_recent_past_meetings(event):
 def test_display_status(event):
     this_morning = LAMetroEvent._time_ago(minutes=120).strftime('%Y-%m-%d %H:%M')
 
-    e = event.build(
+    cancelled_this_morning = event.build(
         name='Board Meeting',
         start_date=this_morning,
         status='cancelled',
         id=get_event_id()
     )
-
-    # Get event from queryset so it has the start_time annotation from the manager
-    cancelled_this_morning = LAMetroEvent.objects.get(id=e.id)
 
     assert cancelled_this_morning.has_passed == True
     assert cancelled_this_morning.display_status == 'Cancelled'
--- a/tests/test_events.py
+++ b/tests/test_events.py
@@ -454,6 +454,25 @@ def test_most_recent_past_meetings(event):
     assert event_earlier_today in recent_past_meetings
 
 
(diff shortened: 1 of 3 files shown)


tests/test_events.py
```diff
@@ -395,6 +395,24 @@ def test_most_recent_past_meetings(event):
...
+
+    e = event.build(
+        name='Board Meeting',
+        start_date=this_morning,
+        status='cancelled',
+        id=get_event_id()
+    )
+
+    # Get event from queryset so it has the start_time annotation from the manager
+    cancelled_this_morning = LAMetroEvent.objects.get(id=e.id)
```
hancush, line 410: FYI the event fixture should return an LAMetroEvent instance. Can you confirm and if so, remove this bit?

hancush (approved):
Looking good, @antidipyramid. I agree that splitting out the code makes it a lot easier to parse! Left one small thing inline. Once you address, you can merge this in.

tests/test_events.py
antidipyramid: It does return an `LAMetroEvent` instance but without this line, the `LAMetroEvent` instance does not have a `start_time` property. This is the error I get without it:

Screen Shot 2022-03-23 at 10 21 28 AM

I added it after getting this error and noticing this in 'test_event_is_upcoming()`:

Screen Shot 2022-03-23 at 10 22 14 AM

It does seem like `start_time` is not technically an attribute of `LAMetroEvent` but an annotation added by `EventManager` in `councilmatic-core`. So the error does make sense.

Screen Shot 2022-03-23 at 10 28 56 AM

hancush: Oh, word, your interpretation of the code is correct!

Since the fixture patch is on master, try rebasing this branch (check out and pull from master, then check out this branch and `git rebase master`), then see if you can remove.

https://github.com/datamade/la-metro-councilmatic/blob/471fc28583bdbb1e19c386fcb26b6a39149a0de2/tests/conftest.py#L134-L135

antidipyramid: I've rebased but now I'm getting this error in `test_upcoming_meetings()`:

Screen Shot 2022-03-23 at 11 04 49 AM

I tried playing around with adding timezones but no luck.

But `test_display_status` does now pass without having to run it through `EventManager`.

hancush: Huh, let's look at this during Metro sprint planning.

hancush:
@antidipyramid Is the only thing remaining on this to push the rebased code?

antidipyramid:
@hancush Yes, sorry for the delay! I'll get this merged in after the checks pass.

Issue #820, opened 2022-03-18
Priority: Remove call-in info for 3/31 meeting


shrayshray:
Can this be prioritized? For the upcoming 3/31 Measure R meeting (https://boardagendas.metro.net/event/measure-r-independent-taxpayer-oversight-7159aebed1a8/), can you please delete the following text from the website that references submitting public comments by phone?:
![image](https://user-images.githubusercontent.com/28063988/159083114-0f0f47db-527c-43cf-b6b7-c171c045cc93.png)

 
There are no live public comments during the Measure R meetings (neither are there for Measure M and ICAOC).

hancush:
@shrayshray Should we also disable the live ecomment link that appears during the meeting?

hancush:
@shrayshray Ah, disregard, I forgot we hid ecomment links a while ago!

shrayshray:
@hancush, perfect! Looks great to me!

PR #823, opened 2022-03-21
Conditionally show call in comment instructions

hancush:
## Overview

This PR hides call in instructions for a subset of committees that do not accept live comment.

### Checklist

## Testing Instructions

 * Scrape agendas for upcoming events:
```bash
docker-compose run --rm scrapers pupa update lametro bills window=7 --rpm=0
docker-compose run --rm scrapers pupa update lametro events window=7 --rpm=0
```
 * View the 3/24 board meeting and confirm you see call in instructions
 * View the 3/31 Measure R meeting and confirm you do not see call in instructions

Handles #820


lametro/templates/partials/event_header_live_public_comment.html
```diff
@@ -0,0 +1,122 @@
+<div class="row">
```
hancush, line 1: This is copy/pasted exactly from `event.html`.

lametro/templates/partials/event_header_no_live_public_comment.html
```diff
@@ -0,0 +1,54 @@
+<div class="row">
```
hancush, line 1: This is almost identical to the public comment partial, but omits ecomment and call in instructions. Defining these separately makes the conditional logic easier to follow.

lametro/models.py
```diff
@@ -722,6 +722,17 @@ def ecomment_message(self):
             return self.UPCOMING_ECOMMENT_MESSAGE
 
 
+    @property
+    def accepts_live_comment(self):
+        meetings_without_live_comment = (
```
fgregg, line 727: not a requirement but making this a set with

{
           'Measure R Independent Taxpayer Oversight Committee',
            'Measure M Independent Taxpayer Oversight Committee',
            'Independent Citizen’s Advisory and Oversight Committee',
}

make intent slightly clearer.

tests/test_events.py
```diff
@@ -564,3 +564,40 @@ def test_meeting_stream_link_unavailable(mocker):
...
+    test_event = event.build(name=event_name, start_date=in_an_hour)
+    event_document.build(note='Agenda', event_id=test_event.id)
+
+    assert test_event.accepts_live_comment == expected_live_comment_value
+
+    url = reverse('lametro:events', args=[test_event.slug])
+    response = client.get(url)
+    response_content = response.content.decode('utf-8')
+
+    live_comment_signature = (
```
fgregg, line 596: you are testing both a model and a template rending here. please split into two tests. the test for the template just needs a true and false condition.

fgregg (changes requested):
looks good, would be good to split the test in two, since you might want to change template at some point.

tests/test_events.py
hancush: Great call, done in https://github.com/datamade/la-metro-councilmatic/pull/823/commits/8c8383b851a5df352d9c5d361edcc8a5168c9979.

lametro/models.py
hancush: Done in https://github.com/datamade/la-metro-councilmatic/pull/823/commits/8c8383b851a5df352d9c5d361edcc8a5168c9979.

hancush:
Thanks for keeping me honest, @fgregg! Made both requested changes, back over to you.

Issue #808, opened 2022-03-08
Omit meetings without Minutes from Minutes page


hancush:
Include all meetings of the board, Crenshaw and LA Safe committees, whether or not they have an available minutes file yet. Exclude committees.

hancush:
We're pulling the minutes page until this fix is in. To start this PR, revert the revert commit from #824.

```bash
git revert -m 1 <REVERT COMMIT HASH>
```

Then, add the fix to omit meetings without minutes.

[Here's a good overview of `git revert`](https://www.datree.io/resources/git-undo-merge), with a link to the git-scm docs for further reading.

hancush:
Hi, @shrayshray – @antidipyramid landed a change to show all board meetings and meetings of the LA Safe and Crenshaw committees, but omit all other committee meetings from the minutes page. New version is deployed to the staging site. What do you think?

shrayshray:
@antidipyramid @hancush i sent this out for review by the team and they all love it! We're good to go live.

hancush:
Deployed!

PR #825, opened 2022-03-25
Omit meetings without minutes from Minutes page

antidipyramid:
## Overview

This PR modifies the Minutes page to only list meetings that can have minutes, i.e. meetings from the board, LA SAFE, and Crenshaw.

### Checklist

## Testing Instructions

 * Navigate to the Minutes page
 * Verify that the page only displays meetings from the board, LA SAFE, and Crenshaw

Handles #808


hancush (approved):
Nice!

Issue #814, opened 2022-03-17
Store analytics remotely


antidipyramid:
After creating a management command to generate tag analytics (#811), we should store the output to a remote location and/or email it to someone

hancush:
Hey, @shrayshray! We're thinking through how these tag analytics reports should be delivered to you. I think I remember you saying Metro uses Dropbox – would it be convenient for us to post the reports to a shared Dropbox? Then, you all could manage access (and even your own email notifications) on your end, and have an archive of past reports handy. 

Let me know what you think!

hancush:
Steps to upload a file to Google Drive:

- Create a service account following the instructions in the first section of this blog post: https://www.twilio.com/blog/2017/02/an-easy-way-to-read-and-write-to-a-google-spreadsheet-in-python.html
  - Be sure to follow the project / service account naming conventions outlined in https://github.com/datamade/how-to/issues/178#issue-815551873
- Create a reports folder in Google Drive and share it with DataMade, Shelly, Omar, [boardreport@metro.net](mailto:boardreport@metro.net), and your service account
- Encrypt and add your client secret to version control
- Use the Google API Python library to upload generated reports to your reports folder
  - Authentication example:
    - Part 1: https://github.com/security-force-monitor/sfm-cms/blob/bdf72bee05fb9842353a563e110d9fb70aa92afe/sfm_pc/management/commands/import_google_doc.py#L9
    - Part 2: https://github.com/security-force-monitor/sfm-cms/blob/bdf72bee05fb9842353a563e110d9fb70aa92afe/sfm_pc/management/commands/import_google_doc.py#L181-L181
    - Part 3: https://github.com/security-force-monitor/sfm-cms/blob/master/sfm_pc/management/commands/import_google_doc.py#L321-L324
   - Uploading will probably look something like this: https://developers.google.com/drive/api/v3/manage-uploads#python

hancush:
@hancush to run May report and share access to Drive with Metro, then close.

shrayshray:
@hancush there are many, MANY, copies of the monthly report .csv in the [Google Drive folder](https://drive.google.com/drive/folders/1iC4IhGGlzMNGKtAWWgi6HOSZ7QYoUEiR?usp=sharing) - is this expected? 
![image](https://user-images.githubusercontent.com/28063988/172256146-b1a1a221-02d6-4cd8-8010-bacfb4116fe8.png)

hancush:
[...]

PR #828, opened 2022-03-30
Upload analytics report to Google Drive

antidipyramid:
## Overview

Generates an analytics CSV file and uses a Google Service Account to write it to a pre-defined folder in Google Drive. I held off on sharing the folder with LA Metro colleagues until this has been tested.

### Checklist

## Testing Instructions

 * Run `blackbox_postdeploy` to decrypt the config files.
 * Get the ID of a test Google Drive folder you'd like to use. See note below for instructions.
 * Add `REMOTE_ANALYTICS_FOLDER = 'id'` (where id is the ID of the folder) to the end of `councilmatic/settings_deployment.py`.
 * Run `docker-compose run --rm app python manage.py generate_tag_analytics`
 * Verify that there is a new file in your test folder.
 * Run `blackbox_shred_all_files` to remove decrypted files from your local repo.

### Notes

In a browser, you can find the ID of a Google Drive folder by right-clicking it and selecting 'Get link':

id

The ID is the highlighted portion of the URL in the picture above.

Handles #814


lametro/management/commands/generate_tag_analytics.py
```diff
@@ -1,34 +1,97 @@
...
+
+        if not drive_service:
+            return
+
+        media = MediaIoBaseUpload(csv_string, mimetype='text/csv')
+        drive_service.files().create(
+                body=file_metadata,
+                media_body=media,
+                fields='id'
+            ).execute()
```
hancush, line 43: Can you remove one level of indentation here for consistency?

lametro/management/commands/generate_tag_analytics.py
```diff
@@ -1,34 +1,97 @@
...
+        csv_string = StringIO()
+        writer = csv.writer(csv_string, delimiter=',')
+        writer.writerow(
+            ('File ID',
+             'Identifier'
+             'Title',
+             'File Name',
+             'Last Action Date',
+             'Tag Classification',
+             'Tag')
```
hancush, line 61: You have seven headings, but you're only writing 6 fields in your loop before. Strike one of "Title" or "File Name"?

Also, what do you think about ordering tag first, then tag classification?

lametro/management/commands/generate_tag_analytics.py
```diff
@@ -1,34 +1,97 @@
...
+        try:
+            credentials = service_account.Credentials.from_service_account_file(
+                    settings.SERVICE_ACCOUNT_KEY_PATH
+                ).with_scopes(SCOPES)
+        except FileNotFoundError:
+            self.stdout.write(self.style.ERROR('Service account key file not found.'))
+            return
+
+        if not credentials:
+            self.stdout.write(self.style.ERROR('Could not generate credentials!'))
```
hancush, line 94: Can you say more about logging rather than raising exceptions on these conditions?

hancush (changes requested):
Nice working figuring out a tricky pattern, @antidipyramid! Don't forget to create a folder for production and add its ID to the corresponding config. (We can always create the folder now and share it with Metro later, after we've QA'ed on staging.)

A few things for your attention inline.

lametro/management/commands/generate_tag_analytics.py
antidipyramid: I did think about raising exceptions instead but I thought that since I wasn't going to be including logic to handle the exception (and the script will just abort), it would suffice to just display the error. 

Is it preferable to raise exceptions in situations like this?

hancush: My rule of thumb is that if an exception prevents a program from completing its intended work, it should raise an exception. In this case, we'll be running this from Airflow, and we'll definitely want the indicator that this command failed so we know we have a problem to fix.

antidipyramid: Ok cool. I'll add in exceptions.

antidipyramid: @hancush And would it still be helpful to handle the exception in a try/except block by printing to stdout with the error info?

lametro/management/commands/generate_tag_analytics.py
antidipyramid: Yeah, I think I got confused while looking at the analytics file. For some reason I thought I was missing a heading for one of the columns.

Tag first works too!

lametro/management/commands/generate_tag_analytics.py
```diff
@@ -1,34 +1,97 @@
...
+        media = MediaIoBaseUpload(csv_string, mimetype='text/csv')
+        drive_service.files().create(
+                body=file_metadata,
+                media_body=media,
+                fields='id'
+            ).execute()
+
+        self.stdout.write(
+            self.style.SUCCESS(f'Successfully uploaded {output_file_name}!')
+        )
```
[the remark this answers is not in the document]
antidipyramid: True, that could also go into a separate method.

antidipyramid:
I added custom exceptions in an attempt to provide helpful error messages. The goal is to raise exceptions that reflect what part of the script failed (the HTTP connection, building the drive, or uploading the file). But if this adds unnecessary complexity, I'm happy to remove it. @hancush

hancush:
@antidipyramid Good thought! I tend to reach for custom exceptions when my program would otherwise fail with a cryptic error, i.e., if I want to unpack one item from a list but there are two, raise an exception saying why that's unexpected instead of the default ValueError.

What errors would be raised in these instances if you didn't add custom exceptions? How descriptive are they?

lametro/management/commands/generate_tag_analytics.py
```diff
@@ -1,34 +1,114 @@
...
+        except DriveBuildError as e:
+            self.stdout.write(
+                self.style.ERROR(f'Unable to connect to Google Drive!')
             )
+            raise
+        except HttpError as e:
+            self.stdout.write(
+                self.style.ERROR(f'HTTP Error!')
+            )
+            raise
```
hancush, line 49: This note seems redundant, since you go on to raise the HTTP error.

antidipyramid:
@hancush [From my understanding](https://googleapis.dev/python/google-auth/latest/reference/google.oauth2.service_account.html), during authentication `google.oauth2` might raise a `FileNotFoundError` or `NotImplementedError`. Since `FileNotFoundError` isn't actually listed in the docs, I imagine there are other exceptions it could raise.

I think the custom `UploadError` is the most useful because if there's a problem uploading the file (not an HTTP error), the Drive API will return one of [many error messages](https://developers.google.com/drive/api/guides/handle-errors) as the JSON response. So without `UploadError`, the upload might fail silently.

lametro/management/commands/generate_tag_analytics.py
antidipyramid: Good point.

hancush:
@antidipyramid Yes to UploadError. 

If you aren't sure what happens if `build()` fails, why don't you try running the code with 1. a missing key file (configure some other path in your settings) and 2. an invalid key file (create a copy of the valid key file and swap a phony value in for the key). My instinct is that we could fail to build a connection for a number of reasons and the distinct error messages would be more helpful than consolidating them into a blanket exception.

lametro/management/commands/generate_tag_analytics.py
```diff
@@ -1,34 +1,100 @@
...
+
+    def generate_tag_analytics(self):
+        """Returns a CSV-formatted byte stream of the tag analytics"""
+
+        csv_string = StringIO()
+        writer = csv.writer(csv_string, delimiter=',')
+        writer.writerow(
+            ('File ID',
+             'Identifier'
```
hancush, line 53: Looks like you're missing a comma here. 👀

```suggestion
             'Identifier',
```

hancush (approved):

PR #862, opened 2022-06-23
Use last setuptools_scm version with Python 3.6 support 

antidipyramid:
## Overview

This PR fixes an issue caused by `django-haystack` using a version of `setuptools_scm` that only supports Python 3.7+.


hancush:
@antidipyramid Ok, figured it out! Our version of Haystack declares `setuptools_scm` in the `setup_requires` section of `setup.py`. [That installs the package globally at build time](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html, if it isn't already installed. So, I added a line prior to installing requirements to install a compatible version of `setuptools_scm`, and that seems to have fixed the issue. I also updated the deployment script that installs our dependencies to install a compatible version of `setuptools_scm`. Assuming CI passes, this should come in, and then be merged into your PRs from yesterday, so CI is a reliable red/green indicator.

Issue #854, opened 2022-05-25
Minutes and Events: Add logic to only show Minutes link after they are approved


shrayshray:
On the Events and Minutes pages, the link to the Minutes shows as soon as the Minutes are available in Legistar. Instead, we'd like them to only show after they have been approved. However, the Minutes report should still show on the Agenda page of the meeting where the approval will take place. In this case (screenshot below) the link to the Minutes of the April Board meeting should not show on the Events page listing until after they're approved at this Thursday's meeting:
![image](https://user-images.githubusercontent.com/28063988/170312978-b4882db5-488d-4cb7-b32f-a68d86f85c80.png)

hancush:
@shrayshray Deployed to the live site

PR #865, opened 2022-06-24
Only show approved minutes on Meetings & Agendas page

antidipyramid:
## Overview

This PR ensures that only minutes that have been approved are displayed on the Meetings & Agendas page.

### Checklist

## Testing Instructions

 * Navigate to Meetings & Agendas page and make sure that only approved minutes are displayed

Handles #854

### DIFF
--- a/lametro/views.py
+++ b/lametro/views.py
@@ -283,9 +283,10 @@ class LAMetroEventsView(EventsView):
         day_grouper    = lambda x: (x.local_start_time.year, x.local_start_time.month, x.local_start_time.day)
 
         # We only want to display approved minutes
-        minutes_queryset = EventDocument.objects.select_related('event')\
-                                        .filter(note__icontains='minutes')\
-                                        .filter(event__extras__approved_minutes=True)
+        minutes_queryset = EventDocument.objects.filter(
+                event__extras__approved_minutes=True,
+                note__icontains='minutes'
+            )
 
         # If yes...
         if start_date_str and end_date_str:
--- a/lametro/views.py
+++ b/lametro/views.py
@@ -284,8 +284,8 @@ class LAMetroEventsView(EventsView):
 
         # We only want to display approved minutes
         all_event_docs = EventDocument.objects.select_related('event')
-        minutes = [doc.id for doc in all_event_docs if doc.event.extras.get('approved_minutes')]
-        minutes_queryset = all_event_docs.filter(id__in=minutes)
+        approved_minutes = [doc.id for doc in all_event_docs if doc.event.extras.get('approved_minutes')]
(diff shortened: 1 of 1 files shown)


lametro/views.py
```diff
@@ -282,7 +282,10 @@ def get_context_data(self, **kwargs):
         end_date_str   = self.request.GET.get('to')
         day_grouper    = lambda x: (x.local_start_time.year, x.local_start_time.month, x.local_start_time.day)
 
-        minutes_queryset = EventDocument.objects.filter(note__icontains='minutes')
+        # We only want to display approved minutes
+        all_event_docs = EventDocument.objects.select_related('event')
+        approved_minutes = [doc.id for doc in all_event_docs if doc.event.extras.get('approved_minutes')]
+        minutes_queryset = all_event_docs.filter(id__in=approved_minutes)
```
hancush, line 288: This will give you all documents related to events that have approved minutes. It's also possible to do all of this in a single ORM statement. Can you add back the original note condition, and refactor this into one query? LMK if you need any hints.

hancush (changes requested):

lametro/views.py
antidipyramid: Great points, thanks!

lametro/views.py
```diff
@@ -282,7 +282,10 @@ def get_context_data(self, **kwargs):
         end_date_str   = self.request.GET.get('to')
         day_grouper    = lambda x: (x.local_start_time.year, x.local_start_time.month, x.local_start_time.day)
 
-        minutes_queryset = EventDocument.objects.filter(note__icontains='minutes')
+        # We only want to display approved minutes
+        minutes_queryset = EventDocument.objects.select_related('event')\
```
hancush, line 286: `select_related` is helpful if you need to access the related objects outside of your query, e.g., if we were going to iterate over these documents, accessing the related event in each loop. Here, Django does the join for you when you add the filter on the event extras. So, you can remove this `select_related` – unless I'm missing something?

lametro/views.py
```diff
@@ -282,7 +282,10 @@ def get_context_data(self, **kwargs):
...
+                                        .filter(note__icontains='minutes')\
+                                        .filter(event__extras__approved_minutes=True)
```
hancush, line 288: It behaves the same way, IIRC, but I prefer combining filter conditions into a single filter call. Can you make that change?

hancush (changes requested):

lametro/views.py
antidipyramid: No, you're right!

lametro/views.py
antidipyramid: No problem.

hancush (approved):

Issue #864, opened 2022-06-23
Top Nav: Point Archives Search Links to FSS


shrayshray:
We're ready to replace the old Board Archives Search with our new Federated Search System! In the top navigation, please replace the the current URL on the link with this one: https://mtasearch01.metro.net:23352/apps/boardarchives/
![image](https://user-images.githubusercontent.com/28063988/175430653-eaa5e4a7-f3fb-46bc-b61c-7173a1abcb8f.png)

@camachoo Do you also want to re-add the link to Board Box Search, using this link: https://mtasearch01.metro.net:23352/apps/boardarchives/?tab=boardbox or do you view that direct link as having use just internally?

hancush:
@shrayshray Deployed to the live site

Issue #863, opened 2022-06-23
Home: Open PDF calendar in new tab


shrayshray:
in #853 we added a link to the PDF calendar. Please have this link open the PDF in a new tab.

hancush:
@shrayshray Deployed to the live site

PR #867, opened 2022-06-27
Link federated search and open calendar links in new tab

antidipyramid:
## Overview

This PR routes the "Archive Search" link in the navbar to Metro's federated search and opens PDF calendar links on the homepage in a new tab.

### Checklist

## Testing Instructions

 * Navigate to homepage and verify that calendar links open in a new tab
 * Verify that archive search link in navbar sends users to federated search

Handles #864 #863

### DIFF
--- a/lametro/templates/base.html
+++ b/lametro/templates/base.html
@@ -72,7 +72,7 @@
                             <a href="{% url 'lametro:search' %}">Board Reports</a>
                         </li>
                         <li>
-                            <a href="{% url 'lametro:archive-search' %}">Archive Search</a>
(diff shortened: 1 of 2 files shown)


lametro/templates/base.html
```diff
@@ -72,7 +72,7 @@
                             <a href="{% url 'lametro:search' %}">Board Reports</a>
                         </li>
                         <li>
-                            <a href="{% url 'lametro:archive-search' %}">Archive Search</a>
```
hancush, line 75: If we're removing this link, do we need [the archive search page](https://boardagendas.metro.net/archive-search) anymore, or can we remove it from the app?

hancush: Cc @shrayshray

Issue #816, opened 2022-03-17
Home: Current Meeting Display Watch buttons are misaligned


shrayshray:
Omar just noticed the display of the "Watch in English" and "Ver en Espanol" are misaligned. I could've sworn they were lined up at the start of the meeting, but I'm seeing it now, too. Here it is in Firefox:
![image](https://user-images.githubusercontent.com/28063988/158886688-1778dfb2-6b5c-4cf3-a227-41326b0ce9b9.png)

And in Chrome: 
![image](https://user-images.githubusercontent.com/28063988/158886732-03b5fe43-71e5-4096-8f77-63c868f4b732.png)

hancush:
@shrayshray We put the listen links below the main block a couple of years ago, because the addition of the eComment link crowded it too much. https://github.com/datamade/la-metro-councilmatic/pull/589#issuecomment-611584220

Do we want to revise this now?

hancush:
Let's try logic like, if there is public comment, show buttons below; otherwise, show them next to the photo, in the event deetz column.

antidipyramid:
@hancush It seems like when the event has an agenda, it'd still look pretty off. 

<img width="703" alt="Screen Shot 2022-07-25 at 11 38 42 AM" src="https://user-images.githubusercontent.com/36973363/180820640-7506a443-ef6e-4f6b-8318-da44e9b616c1.png">

What about placing all buttons in a row under the meeting info/photo if there's an agenda?

hancush:
@antidipyramid Can you mock that up and post here, tagging Shelly?

antidipyramid:
@shrayshray The number of buttons + different screen sizes can lead to an awkward layout.

What do we think of putting all of the buttons in a row? Here's what that could look like on a few different screen sizes:

(13" Macbook)
<img width="679" alt="Screen Shot 2022-07-26 at 12 18 40 PM" src="https://user-images.githubusercontent.com/36973363/181058797-8358d5af-38a6-4977-9454-884207efca2c.png">

(slightly wider screen)
<img width="627" alt="Screen Shot 2022-07-26 at 12 18 59 PM" src="https://user-images.githubusercontent.com/36973363/181058920-0f86a7d0-8573-4bfb-aa15-3d0b785f3916.png">

(very wide screen)
[...]

PR #878, opened 2022-08-03
Fix current meeting buttons

antidipyramid:
## Overview

This PR fixes the alignment of buttons that appear under ongoing meetings on the homepage. 

### Checklist

### Demo

Screen Shot 2022-07-29 at 3 30 59 PM
Screen Shot 2022-07-29 at 3 31 49 PM

## Testing Instructions

Your local app needs to think a meeting is ongoing to display the changes. One way to do this is to manually pass a meeting to the homepage's view. You can do that by shelling into your app with `docker-compose run --rm app python manage.py shell` and grabbing the `id` of an `LAMetroEvent`. Then, set `extra["current_meeting"]` in `LAMetroIndexView.extra_content` with `LAMetroEvent.objects.get(id=[id-of-your-chosen-event])`.

After, verify the changes on the homepage. Some buttons (e.g. spanish and agenda pdf) appear conditionally so you'll need to remove some of that logic in the template to view them.

Handles #816

### DIFF
--- a/lametro/templates/partials/meeting_details_current.html
+++ b/lametro/templates/partials/meeting_details_current.html
@@ -82,46 +82,32 @@
                         <i class="fa fa-fw fa-calendar-o"></i> {{ current_meeting.first.start_time | date:"D n/d/Y"}}<br/>
                         <i class="fa fa-fw fa-clock-o"></i> {{ current_meeting.first.start_time | date:"g:i a"}}<br/>
                         <i class="fa fa-fw fa-map-marker"></i> {{ current_meeting.first.location.name}}<br />
+					  <br />
                     </p>
 
-                    {% if current_meeting.first.documents.all %}
-                        <div class="text-center">
-                            <p>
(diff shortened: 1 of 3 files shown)


lametro/templates/partials/meeting_details_current.html
```diff
@@ -83,46 +83,43 @@ <h4>
...
+					Go to public comment
+				  </a>
+				  {% endif %}
+
+				  {% if current_meeting.first.documents.all %}
+					<a class="btn btn-sm btn-teal current-meeting-info" id="pdf-download-link" target='_blank' href='{{ current_meeting.first.documents.all|find_agenda_url }}'>
+					  <i class='fa fa-fw fa-download'></i>
+					  Get Agenda PDF
+					</a>
+				  {% endif %}
```
hancush, line 118: Can you fix this indentation so each level is nested one level. Re: 2 or 4 spaces, be consistent with the rest of this file.

hancush (commented):
One comment inline, then this can come in. Thanks, @antidipyramid!

Issue #886, opened 2022-08-12
About: Weird behavior on links to search from Report Type


shrayshray:
Two weird behaviors happening when people click on "Board Box / Board Correspondence" from the About Us page:
![image](https://user-images.githubusercontent.com/28063988/184274001-caea8e42-88ee-4a76-9cd5-fb775959aa4d.png)

1. Users in Chrome, FireFox and also Edge: Clicking "Board Box / Board Correspondence" on the page above gets zero results. (filtering by Board Report Type on an empty/fresh search gets 6 results)
2. When they try clicking "Search>Archive Search" from the 0 search results page, the drop down keeps disappearing before they can select "Archive Search" or any other option.
Let me know if you want to do a screenshare illustrate this.

I'd like you all to investigate what's causing this behavior. And some fixes in the meantime:
1. Please change the link on "Board Box / Board Correspondence" on the "About Us" page to point here instead of the boardagendas.metro.net search page: https://mtasearch01.metro.net:23352/apps/boardarchives/?tab=boardbox The six results we're getting with a successful filter are still not the full scope we want to make public - so this update is separate from a potential bug causing the nav bar to act weird - does that make sense?
2. Please update the hyperlink to "Archive Search" (highlighted in the screenshot above) to point here: https://mtasearch01.metro.net:23352/apps/boardarchives/?tab=boardbox
Thank you!

PR #887, opened 2022-08-12
Fix Board Report type search links on About page

antidipyramid:
## Overview

This PR fixes search links in the 'Types of Board Reports' table on the About page.

It also points the Board Box link in the table to the Archive Search for now as we determine which Board Box should be public.

### Checklist

#### URL Encoding

An interesting point here is that form data in query parameters in URLs use `+` to encode spaces. Django's `urlencode` uses standard percent encoding so using it to escape `bill_type` in search URLs would cause errors. 

I found that Firefox automatically escapes form data in URLs correctly but I thought it'd be good form to pass browsers properly escaped URLs so I made a template tag to use `urllib.quote_plus()`.

#### CAPTCHAs

The login in the CAPTCHA script in `base.html` was causing strange navbar behavior. Since navbar dropdowns are also links with `#` for their `href`s, the page was reloading every time a user clicked on them.

## Testing Instructions

 * Navigate to the About page and verify that links to board report types render search results correctly
 * Verify that the navbar behaves properly when you're on the search page
 * Verify that the Board Box link sends you to the archive search

Handles #886

### DIFF
--- a/lametro/templatetags/lametro_extras.py
+++ b/lametro/templatetags/lametro_extras.py
@@ -244,7 +244,7 @@ def get_list(querydict, key):
 @register.filter
 def get_bill_type_link(bill_type):
     for legislation_type in LEGISLATION_TYPE_DESCRIPTIONS:
-        if legislation_type['search_term'].title() == bill_type:
+        if legislation_type['search_term'].upper() == bill_type.upper():
             return f'/about#{legislation_type["html_id"]}'
 
     return ''
--- a/lametro/templatetags/lametro_extras.py
+++ b/lametro/templatetags/lametro_extras.py
@@ -244,7 +244,11 @@ def get_list(querydict, key):
 @register.filter
 def get_bill_type_link(bill_type):
     for legislation_type in LEGISLATION_TYPE_DESCRIPTIONS:
-        if legislation_type['search_term'].upper() == bill_type.upper():
+        if legislation_type['search_term'].title() == bill_type:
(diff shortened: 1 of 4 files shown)


lametro/templatetags/lametro_extras.py
```diff
@@ -244,7 +244,11 @@ def get_list(querydict, key):
 @register.filter
 def get_bill_type_link(bill_type):
     for legislation_type in LEGISLATION_TYPE_DESCRIPTIONS:
-        if legislation_type['search_term'].upper() == bill_type.upper():
+        if legislation_type['search_term'].title() == bill_type:
```
hancush, line 247: Can you say more about this change?

hancush (commented):
Thanks, @antidipyramid! One question inline.

lametro/templatetags/lametro_extras.py
antidipyramid: Looking back over this, I think it's probably an unnecessary change. I just used `title()` because the `bill_type` is usually in title case.

It might actually be better to use `upper()` to avoid any weird edge cases down the line.

hancush: Yeah, I like to superstitiously apply the same transformation on both sides so there are no capitalization mishaps during comparison.

hancush (changes requested):
Hey, @antidipyramid – can you revert the comparison change, and this then can come in? Much obliged! 🙏

PR #891, opened 2022-08-24
Define 'link' variable in find_agenda_url Django filter

antidipyramid:
## Overview

See title.

### Checklist

Handles #890


hancush (approved):
Thanks, @antidipyramid!

PR #892, opened 2022-08-24
Fix manual agenda upload

antidipyramid:
## Overview

This PR fixes an error raised while manually uploading agendas.

### Checklist

## Testing Instructions

 * Login to the Django admin on your local instance
 * Verify that you can upload an agenda PDF for any given event

Handles #889

### DIFF
--- a/tests/test_events.py
+++ b/tests/test_events.py
@@ -83,17 +83,16 @@ def test_can_manually_upload_agenda(event, admin_client, mocker):
     assert not event.documents.filter(note__icontains='Manual')
 
     with open('tests/test_agenda.pdf', 'rb') as agenda:
-        agenda_file = agenda.read()
+        agenda_file = SimpleUploadedFile(
+            "test_agenda.pdf",
+            agenda.read(),
+            content_type="application/pdf"
+        )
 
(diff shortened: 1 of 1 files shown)


hancush (commented):
Wow! Thanks for catching this @antidipyramid. What do you think of adding a test for this functionality? It's pretty crucial, and I'm surprised it doesn't exist already!

antidipyramid:
@hancush Great idea! I'll whip something up.

tests/test_events.py
```diff
@@ -83,17 +83,16 @@ def test_can_manually_upload_agenda(event, admin_client, mocker):
...
+        agenda_file = SimpleUploadedFile(
+            "test_agenda.pdf",
+            agenda.read(),
+            content_type="application/pdf"
+        )
 
-    event.refresh_from_db()
+        admin_client.post(
+            reverse("lametro:events", args=[event.slug]),
+            data={"pdf_form": "", "agenda": agenda_file},
```
hancush, line 94: Passing the file as an instance of `SimpleUploadedFile`, rather than passing `agenda.read()`, is what fixed this! Also, didn't need to instantiate the form in the test. When I tested the upload manually, the `pdf_form` value is just an empty string.

hancush:
@antidipyramid I think we've done it! Want to take a quick glance at the test?

antidipyramid:
Wow, thanks a lot for your help on this one. It passed for me now.

Issue #866, opened 2022-06-24
Home: Update FY23 PDF calendar


shrayshray:
The new Chair changed the time of the Construction Committee meeting. Please replace the PDF calendar on the home page with the version attached.
[FY23 Committee & Board Calendar.pdf](https://github.com/datamade/la-metro-councilmatic/files/8978960/FY23.Committee.Board.Calendar.pdf)

hancush:
@shrayshray How often do meeting times change? If this is something you expect might happen relatively frequently, we should consider taking the extra step to allow you all to manage the meeting PDF yourselves via the admin interface.

shrayshray:
@hancush I'm really praying it doesn't happen frequently (partly why we pushed back on putting it up in the first place). But technically, it can happen as frequently as any particular chair desires, so being able to manage it via the admin interface would be great.

hancush:
@shrayshray Got it. Before we write custom code on this end, I had a thought: Do you all have an easy way of hosting the PDF calendar on your side? That way, we could have one stable link for the file, you can update the file in place if something changes, and the update will show up on the site automatically.

No worries if not, just wanted to check before we invest the (modest amount of) time on our side.

hancush:
@shrayshray While we decide the best path forward, I went ahead and updated the calendar on the site so the information is up to date. 👍

shrayshray:
> @shrayshray Got it. Before we write custom code on this end, I had a thought: Do you all have an easy way of hosting the PDF calendar on your side? That way, we could have one stable link for the file, you can update the file in place if something changes, and the update will show up on the site automatically.
> 
> No worries if not, just wanted to check before we invest the (modest amount of) time on our side.

@camachoo What do you think about hosting this on libraryarchives.metro.net?

hancush:
Waiting for a link from @shrayshray!

shrayshray:
@hancush there was another update to the meeting calendar. Can you please replace the PDF on the site with the version attached?
[FY23 Committee  Board Calendar (002).pdf](https://github.com/datamade/la-metro-councilmatic/files/9647917/FY23.Committee.Board.Calendar.002.pdf)

PR #897, opened 2022-09-27
Update FY23 calendar

antidipyramid:
## Overview

See title.

### Checklist

## Testing Instructions

 * Verify that the calendar links on the homepage and Meetings & Agendas page have been updated

Handles #866


hancush (changes requested):
Thanks, @antidipyramid! Can you use the same filename, just in case someone has the old one bookmarked? Want to avoid link rot, if possible.

Issue #893, opened 2022-08-29
Removing selected facet doesn't work


hancush:
![Screen Shot 2022-08-29 at 9.06.36 AM.png](https://images.zenhubusercontent.com/5b0dc8d3348b7966b4340de0/42a276ff-d9bb-420d-a316-525ae4c6d6c0)

Clicking the X next to a selected facet should remove that facet from the query. Instead, nothing happens.

antidipyramid:
The fix for this bug is now live.

PR #898, opened 2022-10-05
Fix remove facet bug on search pages

antidipyramid:
## Overview

This PR fixes a bug that prevented users from removing a selected facet from their search filter.

### Checklist

## Testing Instructions

 * Navigate to the search page and execute a search
 * Verify that you can filter results by both adding and removing facets

Handles #893

### DIFF
--- a/lametro/templates/base.html
+++ b/lametro/templates/base.html
@@ -148,7 +148,7 @@
                     $target = $target.parent();
                 }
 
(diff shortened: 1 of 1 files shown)


hancush (approved):
Nice bug squash, @antidipyramid! Can you merge to deploy to staging, double check this works ok in deployment, then cut a release to the live site and update the issue thread? TIA! 🙏

PR #921, opened 2022-12-16
Add forms to update bio and headshot

### LINKED ISSUE #909: Allow Metro to manage board member headshots and bios

xmedr:
## Overview
As per issue #909 
> Add a form to the person detail page that allows logged in users to upload a headshot and/or input a bio to the person detail page.

### Checklist

### Demo
The user is logged in and has fields to edit the bio and headshot:
image
<br/><br/>
After separately submitting a new image url like [this stock photo](https://images.unsplash.com/photo-1573496358961-3c82861ab8f4?ixlib=rb-4.0.3&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=688&q=80) and some lorem ipsum text into the bio:
image

### Notes

- I added some form validation into forms.py and PersonHeadshotForm now raises a few different errors on the page for different exceptions.
- I tried doing something similar to PersonBioForm but it seemed like no matter what adjustments I made, the form wasn't running the clean_bio method for validation. I put in print statements and that didn't show up, then I tried having it return the text if there was actually nothing there, but that didn't have any effect either. What I had was:
```
def clean_bio(self):
        bio_text = self.cleaned_data["biography"]

        if bio_text:
            return bio_text
        else:
            raise forms.ValidationError("The new bio cannot be empty. Please enter a biography.")
```
- On the plus side it looks like if someone were to remove the required attribute from the bio form, something in the backend actually isn’t updating the bio if it’s empty or just whitespace, and doing that on its own. Not sure how though.

## Testing Instructions

 * navigate to the board of directors page and select a person
 * the forms should not be visible until logging in
 * log in via the admin or otherwise, and navigate back to a person
 * enter a url into the headshot form and submit
   * if the url was valid, the headshot should change
   * else, an error is displayed
 - enter text into the bio form and submit
   * if text was entered, the bio should change
   * if only whitespace was entered, the bio stays the same

Closes #909


lametro/forms.py
```diff
@@ -135,4 +135,68 @@ def clean_agenda(self):
           if isinstance(agenda_pdf, InMemoryUploadedFile) and agenda_pdf.name.lower().endswith('pdf'):
               return agenda_pdf
           else:
-              raise forms.ValidationError('File type not supported. Please submit a PDF.')
\ No newline at end of file
+              raise forms.ValidationError('File type not supported. Please submit a PDF.')
+
+
+class PersonHeadshotForm(forms.ModelForm):
```
hancush, line 141: Ah, I see how could interpret what I asked for to be the image URL directly, and this achieves that! A better UX would be to upload a file, have the app store that file in the configured storage backend, and save the URL to the file in our backend.

We have a pattern for storing user uploads directly on the server, but looking ahead, it would be more future proof to store user uploads in S3 so the app is portable. `django-storages` provides an easy interface for working with S3. 

What do you think about updating this form to accept a file, then uploading the file and storing the URL on the model in the `form_valid` method of the view? 

See [the docs](https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html) (specifically Usage for configuration, then down to "Or you may want to upload files to the bucket in some view that accepts file upload request...") for an example of how to upload to S3!

lametro/templates/lametro/person.html
```diff
@@ -72,7 +76,7 @@ <h4>
         <div class="col-sm-4 mobile-only">
             <div class="row">
                 <div class="col-xs-4">
-                    <img src='{{person.headshot_url}}' alt='{{person.name}}' title='{{person.name}}' class='img-responsive img-thumbnail' />
+                    <img src='{% if person.image %}{{person.image}}{% else %}{{person.headshot_url}}{% endif %}' alt='{{person.name}}' title='{{person.name}}' class='img-responsive img-thumbnail' />
```
hancush, line 79: How about updating `person.headshot_url` to do this check, so you don't have to do this conditional in the template?

lametro/templates/lametro/person.html
```diff
@@ -83,20 +87,33 @@ <h4>
...
+                    {% endif %}
                 </div>
             </div>
 
             <hr/>
         </div>
 
         <div class='col-sm-8 no-pad-mobile'>
-            {% if member_bio %}
+            {% if member_bio or person.biography %}
```
hancush, line 100: Similarly to `member.headshot_url`, how about adding a property to the LAMetroPerson model that conditionally retrieves `person.biography`, falling back to the baked in biographies from the settings?

lametro/views.py
```diff
@@ -554,11 +556,14 @@ def get_context_data(self, **kwargs):
         return context
 
 
-class LAPersonDetailView(PersonDetailView):
+class LAPersonDetailView(PersonDetailView, FormMixin):
```
hancush, line 559: With two forms, I'm not sure we're getting much from the `FormMixin`.

lametro/views.py
```diff
@@ -585,11 +590,32 @@ def dispatch(self, request, *args, **kwargs):
...
+            return self.render_to_response(self.get_context_data(form=form))
+
     def get_context_data(self, **kwargs):
 
         context = super().get_context_data(**kwargs)
         person = context['person']
 
+        # The first form is set with form_class, but we need to explicitly
+        # add the second, different form to the context.
+        context['form2'] = self.second_form_class
```
hancush, line 617: How about adding both forms to the context here with descriptive names, e.g., headshot_form and biography_form?

lametro/views.py
```diff
@@ -585,11 +590,32 @@ def dispatch(self, request, *args, **kwargs):
...
+    def post(self, request, *args, **kwargs):
+        slug = self.kwargs['slug']
+        person = self.model.objects.get(slug=slug)
+        self.object = self.get_object()
+
+        # The submitted hidden field determines which form was used
+        if 'bio_form' in request.POST:
+            form = PersonBioForm(request.POST, instance=person)
+        elif 'headshot_form' in request.POST:
+            form = PersonHeadshotForm(request.POST, instance=person)
```
hancush, line 602: Clever!

lametro/views.py
```diff
@@ -585,11 +590,32 @@ def dispatch(self, request, *args, **kwargs):
 
         return response
 
+    def post(self, request, *args, **kwargs):
+        slug = self.kwargs['slug']
+        person = self.model.objects.get(slug=slug)
+        self.object = self.get_object()
```
hancush, line 596: What's going on here?

lametro/templates/partials/person_bio_form.html
```diff
@@ -0,0 +1,16 @@
...
+    {% csrf_token %}
+
+    {% if form2.biography.errors %}
+        <br>
+        <p style="color: #eb6864;">{{form2.biography.errors.as_text}}</p>
+    {% endif %}
+    {{form2.bio_form}}
+    <label for="{{form2.biography.id_for_label}}">Update {{form2.biography.label}}:</label>
+    {{form2.biography}}
+    <input type="submit" value="Submit">
```
hancush, line 14: Can you use `button` elements for submit?

hancush (changes requested):
Hey, @xmedr! This is looking groovy. I left some comments inline. The big one is that we should refactor the headshot form so it accepts a file, processes it, and stores the URL on the model instance on the back end.

Let me know if you'd like to pair on anything!

lametro/forms.py
xmedr: Oh I getcha. This sounds good! In that case, I'm looking through the docs to get django-storages set up, and it looks like I'll probably need an account/bucket made for aws keys, is that right? Do we have anything I should use, or should I just make one?

lametro/views.py
xmedr: So for this, I was getting an error along the lines of `LAPersonDetailView object has no attribute 'object'` and after looking it up found that people were having success with adding `self.object = self.get_object()` to assign an object for `get_context_data()` to use. 

I later found that on line 119 someone had the same issue and [linked to this SO answer](https://stackoverflow.com/questions/34460708/checkoutview-object-has-no-attribute-object). So maybe this is just a quirk with django?

lametro/views.py
xmedr: Done! That makes a lot more sense, I was following online suggestions a little too literally there.

lametro/templates/partials/person_bio_form.html
xmedr: Done! Is using a button for submissions more of a better practice?

lametro/forms.py
xmedr: I hit a snag here, what do you think about this: 

I was trying to add a FileField for headshots to LAMetroPerson so I could get them into the bucket I made for this, but I found that you can’t add a field to a proxy model. I asked in chat and got a couple suggestions to potentially either 
- add this field onto django-councilmatic's Person model, or
- create a different model for the file field and refer to it via foreign key on the LAMetroPerson model

I figured the second option might be better, but Derek rightfully suggested that I consult with you after the break before going down either route.

xmedr:
Aright I've got what I have pushed, just a few more comments:

- now that I have the key and secret in the encrypted settings files, I imagine it's safe to remove those from the local settings file?
- I'm realizing that since I moved the form validation into the view, that I no longer have the form showing errors. I'm at the end of my day now, but I can probably figure that out in the future!
- also I'm just now noticing that the image works when it's uploaded, but after a few, it stops working, and when I open the link for it in a new tab, it says the request has expired. I'm not sure what's going on there
- lastly, I tried re-encrypting those settings files, and I got the same error for both of them after running `blackbox_end_edit configs/settings_deployment.staging[or production].py`
```
========== PLAINFILE "configs/settings_deployment.staging.py"
========== ENCRYPTED "configs/settings_deployment.staging.py.gpg"
========== Encrypting: configs/settings_deployment.staging.py
gpg: error retrieving 'hannah.cushman@gmail.com' via WKD: No data
gpg: hannah.cushman@gmail.com: skipped: No data
gpg: configs/settings_deployment.staging.py: encryption failed: No data
```

hancush:
@xmedr 

> now that I have the key and secret in the encrypted settings files, I imagine it's safe to remove those from the local settings file?

You'll want to leave these in, so we can set them in local development. For optional settings, like this, I usually do something [like this](https://github.com/confluency/stormstore-app/blob/b5277094fcf6e553f027bc26a9fc3e1c76fff9b9/stormstore_app/settings.py#L196-L208).

> I'm realizing that since I moved the form validation into the view, that I no longer have the form showing errors. I'm at the end of my day now, but I can probably figure that out in the future!

👍

> also I'm just now noticing that the image works when it's uploaded, but after a few, it stops working, and when I open the link for it in a new tab, it says the request has expired. I'm not sure what's going on there

By default, django-storages / boto3 will do querystring authentication, which creates expiring links. Since these are nonsensitive images, I think it makes sense to make our bucket public (you can copy the permissions from the `councilmatic-document-cache` bucket) and disable querystring authentication [with the `AWS_QUERYSTRING_AUTH` setting](https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html).

> lastly, I tried re-encrypting those settings files, and I got the same error for both of them after running blackbox_end_edit configs/settings_deployment.staging[or production].py

You need to import the keys of everyone listed in keyrings/live/blackbox-admins.txt. You can find them by email, like this: `gpg --search ${EMAIL_ADDRESS}`. The only one you won't find is the deploy key. I'll send that to you via Slack.

xmedr:
@hancush The bucket seems to allow access to new files without expiring which is great. And the settings files are reencrypted with their respective api keys.

Other than the form errors which I can create an issue for, the last thing I've got is that I might not have access to the Stormstore link you sent with your example of optional settings. It gives me a 404 error and going to Confluency's page says they have no public repo's.

hancush:
Whoops, always forget that repo is private! Just sent you an invite.

xmedr:
@hancush At long last! I ran into another issue with the s3 bucket objects not being public, but I paired with Sam and now they work just fine. I also have errors showing up on the headshot form:
image

hancush:
Woo! Ready for a full re-review, @xmedr?

xmedr:
@hancush I absolutely am! Should I do the pr template in another comment here, or should I open a new pr? Or something completely different?

hancush:
@xmedr Nothing else required! Just wanted to make sure you were done making changes. 👍

lametro/forms.py
```diff
@@ -135,4 +135,37 @@ def clean_agenda(self):
...
+
+    headshot_form = forms.BooleanField(widget=forms.HiddenInput, initial=True)
+
+    def __init__(self, *args, **kwargs):
+        super(PersonHeadshotForm, self).__init__(*args, **kwargs)
+        self.fields['image'].widget.attrs.update({
+                    'required': 'True',
+                    'placeholder': 'Enter URL to image...'
```
hancush, line 149: How about updating (or removing) this placeholder?

lametro/models.py
```diff
@@ -391,7 +391,12 @@ def headshot_url(self):
             filename
         )
 
-        if Path(manual_headshot).exists():
+        if self.image:
+            # Assigning this to image_url would make the return static() at the
+            # end of this add 'static' to the url. Returning here solves that.
+            return self.image
```
hancush, line 397: Nice! How about returning image, if it exists, at the very top of this property – that way we aren't doing any of the other work if we don't need to?

lametro/models.py
```diff
@@ -402,6 +407,16 @@ def headshot_url(self):
...
 
+    @property
+    def current_bio(self):
+        if self.biography:
+            bio = self.biography
+
+        elif self.slug_name in MEMBER_BIOS:
+            bio = MEMBER_BIOS[self.slug_name]
+
+        return bio
```
hancush, line 418: What happens if neither of these is True?

lametro/views.py
```diff
@@ -585,11 +587,74 @@ def dispatch(self, request, *args, **kwargs):
...
+            # create full file path
+            file_path_within_bucket = os.path.join(
+                file_dir_within_bucket,
+                file_obj.name
+            )
+
+            media_storage = MediaStorage()
+
+            media_storage.save(file_path_within_bucket, file_obj)
+            file_url = media_storage.url(file_path_within_bucket)
```
hancush, line 626: How about putting this into its own method that returns `file_url`?

hancush (changes requested):
Hey, @xmedr! This looks awesome. Can you do me a favor and add a section to the README about how to configure the AWS credentials for development? Left a couple of small things inline!

lametro/models.py
xmedr: Oh, yea this makes sense! We wouldn't need any of the file directory business if the image just already exists.

xmedr:
@hancush Alright, I've got your suggestions incorporated and the README is updated! Let me know if what I wrote in there could be more clear or if I missed some steps.

councilmatic/custom_storage.py
```diff
@@ -0,0 +1,4 @@
+from storages.backends.s3boto3 import S3Boto3Storage
+
+class MediaStorage(S3Boto3Storage):
+    bucket_name = 'la-metro-headshots-staging'
```
hancush, line 4: Will this need to vary based on environment?

hancush (commented):
Left one comment to be resolved, then this can come in! Thanks, @xmedr.

xmedr:
Hmm it looks like the tests are failing because it can't find `BUCKET_NAME`, what do you think I missed here @hancush?
`AttributeError: 'Settings' object has no attribute 'BUCKET_NAME'`

hancush:
@xmedr Is the variable in your `settings.py` file actually called `BUCKET_NAME`, or something else? To debug, why don't you write down how the bucket variable is being set, starting from how the user configures it, then tracing how it flows from settings deployment to the main settings file?

xmedr:
@hancush It's alive! Good call on that, ty. I was confused because I had the variable names right in the gpg files, and it was printing out the right value locally. But looking at the error again I saw the problem was I didn't put `BUCKET_NAME` into the `test_settings.deployment.py` file 🤦 So now we've got the tests passing.

hancush:
Thanks, @xmedr! I've merged the upgrade and screwed up your groove over here. Want to pair on rebasing Monday AM? I'll throw a time on the calendar.

hancush:
Superseded by #939.

Issue #923, opened 2023-01-04
Update tag analytics structure


hancush:
Per Omar:

---

Changes in the report:

Please have the column name changed to
File ID -> ??? (What is this?)
Identifier -> File ID

ADD: I would like to have the “MatterId” after the File ID (Identifier) column with the column title “Matter ID”.

ADD: I would like to have "api_metadata" after the Tag column with the column title “Tag GUID”.

CHANGE: I would like to not have the values of the Tag Classification all end with “_exact”.

hancush:
Done in #932

PR #932, opened 2023-01-11
Update tag analytics file structure

antidipyramid:
## Overview

This PR implements requested changes to the tag analytics file.

### Checklist

## Testing Instructions

 * Run `docker-compose run --rm app python manage.py generate_tag_analytics -l` to generate an analytics file in the project root
 * Note: since this might take a while to run, you might want to amend the `generate_tag_analytics` function to only loop over the first few bills in your database

Handles #923

### DIFF
--- a/lametro/management/commands/generate_tag_analytics.py
+++ b/lametro/management/commands/generate_tag_analytics.py
@@ -79,12 +79,11 @@ class Command(BaseCommand):
         self.stdout.write('\nGenerating tag analytics...')
 
         for bill in tqdm(LAMetroBill.objects.iterator(chunk_size=200)):
-            matter_id = bill.api_representation['MatterId']
             for tag in bill.rich_topics:
                 writer.writerow(
                     (bill.board_report.id,
                      bill.identifier,
-                     matter_id,
+                     self.get_matter_id(bill.api_source.url),
                      bill.friendly_name,
                      bill.last_action_date,
                      tag.name,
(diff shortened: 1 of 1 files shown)


lametro/management/commands/generate_tag_analytics.py
```diff
@@ -49,31 +66,48 @@ def generate_tag_analytics(self):
...
              'Last Action Date',
              'Tag',
+             'Tag GUID',
              'Tag Classification')
         )
 
         self.stdout.write('\nGenerating tag analytics...')
 
         for bill in tqdm(LAMetroBill.objects.iterator(chunk_size=200)):
+            matter_id = bill.api_representation['MatterId']
```
hancush, line 82: You can parse this out of the API source rather than hitting the API. :-) 

```
>>> b.api_source
<BillSource: https://webapi.legistar.com/v1/metro/matters/5627>
```

lametro/management/commands/generate_tag_analytics.py
```diff
@@ -20,12 +22,27 @@ class Command(BaseCommand):
...
 
         csv_string = self.generate_tag_analytics()
         date = datetime.today().strftime('%m_%d_%y')
         output_file_name = f'{date}_tag_analytics.csv'
 
+        if local:
+            with open(output_file_name, 'wb') as f:
+                copyfileobj(csv_string, f)
+
+            return
```
hancush, line 44: Can you say more about what's going on here?

hancush (changes requested):
Thanks, @antidipyramid! Couple of small things inline.

lametro/management/commands/generate_tag_analytics.py
antidipyramid: Sure, this is a way to write the StringIO object containing the CSV to a local file. It's already a file-like object so you have to use `shutil.copyfileobj`.

I included the `-l` flag to make local testing easier.

hancush (approved):

PR #934, opened 2023-01-18
Heroku Deployment

fgregg:
Adjusting the application so that it could be deployed on heroku.

Todo 
- [x] Make many settings set by the environment
- [x] Get some data loaded in.
- [x] Set up third-party elastic search
- [x] Clean up old settings files
- [x] Get tests passing


fgregg:
Made some good progress today!

The site is deploying to https://la-metro-councilmatic-staging.herokuapp.com/

The next step is to figure out why whitenoise is not serving the static files.

Dockerfile
```diff
@@ -26,13 +17,8 @@ RUN pip install --upgrade pip setuptools && \
...
 COPY . /app
-COPY councilmatic/settings_deployment.py.example /app/councilmatic/settings_deployment.py
 
-# Copy Blackbox executables from builder stage
-COPY --from=builder /usr/local/bin/blackbox* /usr/local/bin/
-COPY --from=builder /usr/local/bin/_blackbox* /usr/local/bin/
-COPY --from=builder /usr/local/bin/_stack_lib.sh /usr/local/bin/
-
-RUN DATABASE_URL='' python manage.py collectstatic --noinput
+RUN mkdir /app/static
```
hancush, line 21: @fgregg You need to bake static files into the image with collectstatic. Heroku dynos are ephemeral, so files collected on release won't persist.

I'm not sure, actually, why we have collect static in our templated release script.

fgregg: isn't this what whitenoise is supposed to address, or is that a terrible lie?

hancush: No, whitenoise is just a static file server: https://whitenoise.evans.io/en/latest/

There's nothing for it to serve if the files aren't there!

fgregg: got it.

Dockerfile
```diff
@@ -1,20 +1,11 @@
-FROM ubuntu:20.04 as builder
-
-# Clone and build Blackbox
-RUN apt-get update && \
-    apt-get install -y build-essential git-core && \
-    git clone https://github.com/StackExchange/blackbox.git && \
-    cd blackbox && \
-    make copy-install
```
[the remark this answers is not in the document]
fgregg: cool, i'll bring them back.

fgregg:
search is working now. and some data is loaded in.

fgregg:
Okay, every bit seems to be working now, and I think you can treat this as complete bit of work.

https://la-metro-councilmatic-staging.herokuapp.com/

I have set up this app with the settings from settings_deployment.staging.

I spiked a little GithubAction scraper to populate, https://github.com/datamade/la-metro-scrapers. If and when you bring it in, you can delete that repo.

PR #935, opened 2023-01-24
Log sources

xmedr:
## Overview

Print out the sources for each person, event, and bill on their respective detail page.

### Checklist

### Demo

image

## Testing Instructions

 * Navigate to the detail page for a person, event, or bill
 * Open the console to check if it's source has been printed
 * Follow link(s) present

Handles #910


xmedr:
At time of writing, I haven't been able to test bill detail pages. The thought is that I need the secrets file for my scraper.

lametro/views.py
```diff
@@ -90,6 +90,9 @@ def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         bill = self.get_object()
 
+        sources = list(bill.sources.all())
+        context['sources'] = sources
```
hancush, line 94: If you have access to `bill` in the template, do you need to add sources as an additional context variable, or can you access it via `bill`?

lametro/views.py
```diff
@@ -156,6 +159,9 @@ def get_context_data(self, **kwargs):
         context = super(EventDetailView, self).get_context_data(**kwargs)
         event = context['event']
 
+        sources = list(event.sources.all())
```
hancush, line 162: Ditto.

lametro/views.py
```diff
@@ -590,6 +596,9 @@ def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         person = context['person']
 
+        sources = list(person.sources.all())
```
hancush, line 599: Ditto.

hancush (changes requested):
Looking good, @xmedr! One comment, applicable to three spots, inline. Can you also go ahead and remove the OCD ID logs while you're in these files? We used to host this data in an API, but it's no longer active, so they aren't useful any more.

lametro/views.py
xmedr: Ah gotcha, we already have a way to access the sources with the variables we have, so it makes sense to just use what we've got instead of adding more vars. I've started this change for the bills and I'll hopefully have the rest done by Monday!

hancush:
Superseded by #940

Issue #482, opened 2019-08-29
Board of Directors: Change Map display from Districts, Sectors and City of LA upon member selection


shrayshray:
On the[ Board of Directors page](https://boardagendas.metro.net/board-members/), is it possible to change the map display from Districts, Sectors and City of LA depending on which Board Member is selected? 
The default on entering the Board of Directors' page is "Districts". When I select someone who represents a District, their district is highlighted; perfect. But when I select someone who is nominated by a Sector, nothing happens on the map. I have to switch the button above the map to "Sectors", then click on the Sector nominee, to see his/her nominating sector. It would be great  to have the map area display automatically switch from District, Sector or City when a member of whichever type is selected. Here's an example where James Butts is selected, but his area is not displaying because we're in the default "Districts" mode:
![image](https://user-images.githubusercontent.com/28063988/63905019-b989c700-c9c7-11e9-83d5-1176789fc563.png)

hancush:
DataMade to also propose a new, more prominent home for the address lookup.

hancush:
Update "Look up Metro Board Member" and/or add example text

xmedr:
Hi @shrayshray I made up a quick wireframe to rearrange some elements of the board members page based off of some comments from our last meeting, and wanted to know what you think!

There was mention of the search bar being slightly hidden on the page, and if this site is for finding your members and their activities, we may want to make this more prominent.

Then, we have some examples of searches for the user underneath the search bar for a clearer idea of how specific you can get.

Last, the "Learn More" tip was moved to the right of the labels since they directly describe them.

![board members page](https://user-images.githubusercontent.com/114717958/216664395-81a7f24a-fc50-49fb-91c3-08db1f7b1b49.png)

shrayshray:
@xmedr this looks great to me! @neilarellano do you have any feedback on the wireframe?

neilarellano:
this makes good use of the available space. I'm always a fan of seeing everything important in one page instead of having to scroll through it

neilarellano:
this new layout puts good emphasis on the search option which I think most people would appreciate

PR #944, opened 2023-02-20
Move search bar and popover

### LINKED ISSUE #943: Update search bar on directors' page
Create new, more prominent home for the address lookup.

Update "Look up Metro Board Member" and/or add example text.

xmedr:
## Overview

After getting approval on [this comment](https://github.com/datamade/la-metro-councilmatic/issues/482#issuecomment-1416169503), the search bar and _Learn More_ tooltip have been moved to improve UX and make the search bar more of a highlighted feature.

Closes #943 

### Demo

image
image

### Notes

I expected the popover to follow where the _Learn More_ link was, but it stayed in place at the top of the page. So I added a top margin to the css of popovers since, doing a search through the files showed that here was the only place a popover was used like this. Because of that I'm not expecting for that margin to mess with anything else, but if it does then I can absolutely figure something else out for the positioning!

## Testing Instructions

 * Navigate to board of directors page
 * Check the styling/positioning of the search bar and _Learn More_ tooltip


lametro/templates/board_members/board_members.html
```diff
@@ -23,16 +23,15 @@
...
-
         </div>
-        <div class="col-sm-6">
-            <h4>Look up a Metro Board Member</h4>
+    </div>
+</div>
+
+<div class="container-fluid d-flex justify-center align-center" id="hero">
+    <div class="row d-flex justify-center w-100">
+        <div class="col-sm-9 bg-white w-80">
```
hancush, line 33: This works but doesn't do well on smaller screens. We prefer the column classes for setting widths and aligning content horizontally because they're automatically responsive. In this case, you can do something like:

```
<div class="container-fluid" id="hero">
    <div class="row d-flex align-center h-100">
        <div class="col-sm-10 col-sm-offset-1 bg-white">
```

Then, you just need to define an `h-100` (height: 100%) class and give the `#hero` element a fixed height for the search bar to be vertically aligned center.

lametro/static/css/city_custom.css
```diff
@@ -387,4 +388,42 @@ hr .events-line {
...
+
+.w-80 {
+    width: 80%;
+}
+
+.w-100 {
+    width: 100%;
+}
+
+#hero {
```
hancush, line 417: How about a more descriptive id, like `#search-header`?

lametro/static/css/city_custom.css
```diff
@@ -387,4 +388,42 @@ hr .events-line {
...
+    background-image:
+        linear-gradient(
+          rgba(0, 0, 0, 0.9),
+          rgba(270, 270, 270, 0.6)
+        ),
+        url('../images/la-metro-regional-rail-map-copy.gif');
```
hancush, line 423: Thoughts on using a different image from the homepage, to differentiate?

lametro/templates/board_members/board_members.html
```diff
@@ -44,8 +43,12 @@ <h4>Look up a Metro Board Member</h4>
                     </a>
                 </div>
             </div>
+            <p>Search by zip (ex. 90012), city (ex. Acton, Los Angeles), or street address (ex. 200 N Spring St, Los Angeles)</p>
```
hancush, line 46: How about a little top margin between the search bar and this example text?

lametro/templates/board_members/board_members.html
```diff
@@ -23,16 +23,15 @@
...
-                <strong>(2) Sectors</strong> – As defined by the California League of Cities, all the incorporated and unincorporated areas in Los Angeles County, excluding the City of Los Angeles, are divided into four sectors. Each sector nominates one representative, which is approved and appointed by all members of the California League of Cities, to serve on the Metro Board.<br><br>
-                <strong>(3) City of Los Angeles</strong> – Residents within the City of Los Angeles are represented by the mayor and three city appointees on the Metro Board.<br><br>
-                Learn more about <a href='{% url 'about' %}#about-la-metro'>Metro Board appointments</a>.</p>"><h2><small> <i class="fa fa-info-circle" aria-hidden="true"></i> Learn more</small></h2><a>
             </h1>
-
         </div>
-        <div class="col-sm-6">
-            <h4>Look up a Metro Board Member</h4>
+    </div>
+</div>
```
hancush, line 29: Any reason to have multiple container elements, or can we consolidate into one?

lametro/templates/board_members/board_members.html
```diff
@@ -73,6 +76,11 @@ <h1>
                   <label id="cityToggle" class="btn btn-xs btn-city">
                     <input type="radio" name="options" autocomplete="off"> City of LA
                   </label>
+                  <a tabindex="0" class="map-info" data-container="body" data-toggle="popover" data-placement="right" title="What does this map represent?" data-content="<p style='font-family: &quot;Open Sans&quot;, sans-serif; font-size: 14px;'>The three layers of this map show areas from which members of the Metro Board of Directors are nominated:<br><br>
```
hancush, line 79:
```suggestion
                  <a tabindex="0" class="map-info" data-container="body" data-toggle="popover" data-placement="right" title="What do these map layers represent?" data-content="<p style='font-family: &quot;Open Sans&quot;, sans-serif; font-size: 14px;'>The three layers of this map show areas from which members of the Metro Board of Directors are nominated:<br><br>
```

hancush (changes requested):
Looking great, @xmedr! Some markup suggestions inline.

lametro/templates/board_members/board_members.html
xmedr: Gotcha, that makes sense. That offset seems super useful

lametro/templates/board_members/board_members.html
xmedr: We absolutely can. This was a leftover from when I moved the Learn More tooltip and I hadn't considered removing these, so that's for sure the move!

xmedr:
@hancush Awesome, I've got all these changes incorporated! I also changed the example search text to just be black so it's easier to read.

hancush (commented):
Nice! One last thing: The background image is kind of huge. I like to run any images for the web through tinypng.com to reduce their size. Can you do that, commit the smaller image, and then bring this in?

xmedr:
@hancush awesome that cut the size by a little more than half, makes sense to use that site

hancush (approved):

PR #949, opened 2023-03-03
Update map to change layers based off selected member

### LINKED ISSUE #482: Board of Directors: Change Map display from Districts, Sectors and City of LA upon member selection

xmedr:
## Overview

This PR allows users to change map layers based off clicking board members in the adjacent table.

Closes #482 

### Notes

At time of writing, the layers change and mouseover still works, with one caveat. When first clicking a member to change the map layers, that member's section does not highlight right away. You need to continue hovering over them to get the highlight to start working for that layer. 

I think this is because a `tableover` event that is responsible for highlighting that member's section doesn't go off until the appropriate layer is active. I'm not sure how to go about tackling this at the moment.

## Testing Instructions

 * In the board members page, click through different members in order to change map layers

### DIFF
--- a/lametro/templates/board_members/_map.html
+++ b/lametro/templates/board_members/_map.html
@@ -368,21 +368,6 @@
         });
     }
 
-    function activateButton(button) {
-        $('#' + button + 'Toggle').prepend('<i class="fa fa-check" aria-hidden="true"></i>');
-        $('#' + button + 'Toggle').addClass('active');
-
-        // Remove the 'active' class from unused layer buttons
-        let layersArr = ['district', 'section', 'city'];
-        layersArr = jQuery.grep(layersArr, function(value) {
-            return value != button;
-        });
-
-        layersArr.forEach(element => {
-            $('#' + element + 'Toggle').removeClass('active');
-        });
-    }
-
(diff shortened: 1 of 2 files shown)


hancush (changes requested):
Nice refactor, @xmedr! Now that I actually use this, I'm thinking hover might be the better choice. What do you think? Would that be tough to implement?

xmedr:
@hancush Done and done! Thankfully I had a commit where the mouseover was working well and just had to change that up a bit.

hancush (approved):
👏

PR #960, opened 2023-04-12
Live stream links

### LINKED ISSUE #925: Live stream links are visible after meeting conclusion
Per @neilarellano:

---

We were having issues earlier today where the view stream links for both English and Spanish streams are still showing even after the meeting ended. This was a short meeting but the view stream links showed up for another 20 minutes or so. Are we able to check the logs for this one to check?

xmedr:
## Overview

This prevents short meetings from being included in `potentially_current_meetings` if it is past their start time and the event has already been broadcast.

Closes #925 

### Notes

The `test_current_meeting_no_streaming_event_late_start` was failing after implementing the new filter in `potentially_current_meetings` because `event.build()` assumes that if the current time is past the event's start time, then it has been broadcast. This test exists to check meetings that are 15 minutes late and have not yet been broadcasted, so I believe that removing the broadcast here would more closely mimic what would actually happen, and get that test to pass.

Also something to note is that`test_private_event` was failing locally for me on this branch and on the main branch, so I don't immediately think it's related to these changes. But I see that all the tests are passing here. If necessary, I can try to figure out what's going on in a separate pr!

## Testing Instructions

 * Run pytest and confirm that all tests pass
   * (Again, `test_private_event` may fail, not sure why)

### DIFF
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -641,7 +641,7 @@ class LAMetroEvent(Event, LiveMediaMixin, SourcesMixin):
 
         return cls.objects.filter(
             start_time__gte=six_hours_ago, start_time__lte=five_minutes_from_now
-        ).exclude(status="cancelled")
+        ).exclude(status="cancelled", broadcast__observed=True)
 
     @classmethod
     def _streaming_meeting(cls):
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -641,7 +641,7 @@ class LAMetroEvent(Event, LiveMediaMixin, SourcesMixin):
 
         return cls.objects.filter(
             start_time__gte=six_hours_ago, start_time__lte=five_minutes_from_now
-        ).exclude(status="cancelled", broadcast__observed=True)
+        ).exclude(Q(status="cancelled") | Q(broadcast__observed=True))
 
     @classmethod
     def _streaming_meeting(cls):
--- a/tests/test_events.py
+++ b/tests/test_events.py
@@ -203,7 +203,7 @@ def test_current_meeting_no_streaming_event(concurrent_current_meetings, mocker)
 
 def test_current_meeting_no_streaming_event_late_start(event, mocker):
     """
-    Test that if an meeting is scheduled but not yet streaming, it is returned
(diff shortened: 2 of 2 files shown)


lametro/models.py
```diff
@@ -641,7 +641,7 @@ def _potentially_current_meetings(cls):
 
         return cls.objects.filter(
             start_time__gte=six_hours_ago, start_time__lte=five_minutes_from_now
-        ).exclude(status="cancelled")
+        ).exclude(Q(status="cancelled") | Q(broadcast__observed=True))
```
hancush, line 644: Nice!

tests/test_events.py
```diff
@@ -203,7 +203,7 @@ def test_current_meeting_no_streaming_event(concurrent_current_meetings, mocker)
 
 def test_current_meeting_no_streaming_event_late_start(event, mocker):
     """
-    Test that if an meeting is scheduled but not yet streaming, it is returned
+    Test that if a meeting is scheduled but not yet streaming, it is returned
```
hancush, line 206: "An meeting" 🧐🍷

tests/test_events.py
```diff
@@ -216,6 +216,10 @@ def test_current_meeting_no_streaming_event_late_start(event, mocker):
     }
     late_current_meeting = event.build(**crenshaw_meeting_info)
 
+    # Remove the broadcast since it hasn't been started. If it were kept,
+    # it would be filtered out by potentially_current_meetings
+    late_current_meeting.broadcast.get().delete()
```
hancush, line 221: Great catch! It'd be great if we could pass in a keyword argument to event.build() that allows us configure whether to [build the event broadcast](https://github.com/Metro-Records/la-metro-councilmatic/blob/5931c6debd49c8bd4c7286f20866f261c1520bc5/tests/conftest.py#L148) regardless of event time. Want to open an issue for the improvement?

hancush (approved):
👏

tests/test_events.py
xmedr: Sounds good! I've got that issue opened and self-assigned.

PR #962, opened 2023-04-19
Allow streaming meetings to be current

hancush:
## Overview

See title. Meeting broadcasts are created when a meeting goes live, so we were prematurely excluding those events from the current meeting query.

Cc @xmedr – I didn't catch this! 

Connects #925 

### Demo

Deployed to staging.

Screenshot 2023-04-19 at 1 22 33 PM

## Testing Instructions

 * Deployed to staging to test.

### DIFF
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -639,9 +639,14 @@ class LAMetroEvent(Event, LiveMediaMixin, SourcesMixin):
         six_hours_ago = cls._time_ago(hours=6)
         five_minutes_from_now = cls._time_from_now(minutes=5)
 
+        was_cancelled = Q(status="cancelled")
+        has_passed = Q(broadcast__observed=True) & ~Q(
+            pk__in=cls._streaming_meeting().values_list("pk")
+        )
+
         return cls.objects.filter(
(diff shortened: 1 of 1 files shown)


lametro/models.py
```diff
@@ -639,9 +639,14 @@ def _potentially_current_meetings(cls):
         six_hours_ago = cls._time_ago(hours=6)
         five_minutes_from_now = cls._time_from_now(minutes=5)
 
+        was_cancelled = Q(status="cancelled")
+        has_passed = Q(broadcast__observed=True) & ~Q(
+            pk__in=cls._streaming_meeting().values_list("pk")
```
xmedr, line 644: This is neat, first time I've seen `values_list()`. So from the docs, it looks like this is a way to list out all the values in a field. And you're using that to ensure that meetings that are currently streaming aren't being filtered out, cool!

xmedr (commented):
Nice, this is a cool solution to not filter out meetings that still in progress.

lametro/models.py
hancush: I love a values list. It's cuter (and more efficient) than making a list of values in a list comprehension over an entire queryset.

PR #963, opened 2023-04-21
Add kwarg for creating test broadcasts, adjust tests

### LINKED ISSUE #961: Have keyword argument decide whether to create broadcasts for test events
Right now, `event.build()` in `conftest.py` defaults to making a broadcast associated with the event [if the start time is in the past](https://github.com/Metro-Records/la-metro-councilmatic/blob/5931c6debd49c8bd4c7286f20866f261c1520bc5/tests/conftest.py#L148). Let's change that to instead create the broadcast based off of a keyword argument regardless of start time.

xmedr:
## Overview

`event.build()` would default to creating a broadcast for an event if the start time was set in the past. This caused issues with the way we're now determining potentially current meetings (see #960). This changes that behavior to instead be based off of a new keyword argument `has_broadcast` which defaults to false. Tests that involve broadcasts will now have to explicitly create them.

Closes #961 

## Testing Instructions

 * Run pytest and confirm that all tests pass

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -127,7 +127,7 @@ def legislative_session(db, jurisdiction):
 @pytest.mark.django_db
 def event(db, jurisdiction):
     class EventFactory:
-        def build(self, **kwargs):
+        def build(self, has_broadcast=False, **kwargs):
             event_info = {
                 "id": "ocd-event/17fdaaa3-0aba-4df0-9893-2c2e8e94d18d",
                 "created_at": "2017-05-27 11:10:46.574-05",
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -145,7 +145,7 @@ def build(self, **kwargs):
             # Get event from queryset so it has the start_time annotation from the manager
             metro_event = LAMetroEvent.objects.get(id=event.id)
 
-            if metro_event.start_time < datetime.now(timezone.utc):
+            if has_broadcast:
```
hancush, line 148: It'd be more intuitive for this fixture to create the broadcast for past events unless has broadcast is False. Want to try that revision?

hancush (changes requested):

hancush (approved):
This is great, @xmedr! Thank you!

PR #964, opened 2023-05-01
Fix small text for board reports on phones using firefox

### LINKED ISSUE #941: Some Agenda Item Text Appears Small
Hi Team!

We noticed that some agenda items appear smaller than the rest. We confirmed this to be the case on multiple devices when using Firefox mobile browser. Please see screenshot.

Thank you!
Screenshot_20230210-181456~2

xmedr:
## Overview

This fixes an issue where text for board report agendas were being made smaller on some items. It looks like this only happened in td's where the text **could** fit in one line, if it were made smaller. So the solution I found was to increase the size of that text slightly to make it take more than one line.

Closes #941

### Demo

Before
image

After
image

### Notes

The first thing I tried was simply setting a font-size for all those elements (16px), and that didn't do anything. Then I tried resetting firefox specific styles, but that also didn't work out. This is why I'm just increasing the size for those item blurbs.

I also found that this issue isn't present when the device is turned to landscape. So I'm thinking that with the `max-device-width` and `orientation: portrait` css rules, this will only apply to smartphones in that orientation. So far with the devices/browsers I've tried on browserstack, this has been successful.

## Testing Instructions

 * Using firefox on a mobile device, navigate to an event with an agenda.
 * Confirm that the agenda items' text are all the same size.
 * Confirm that this does not effect other browsers on mobile or desktop

### DIFF
--- a/lametro/static/css/city_custom.css
+++ b/lametro/static/css/city_custom.css
@@ -423,12 +423,12 @@ hr .events-line {
     background-position: center center;
     height: 200px;
 }
-
+/*
 body {
     /* text-size-adjust: none; */
-    -moz-text-size-adjust: none;
+    /* -moz-text-size-adjust: none; */
     /* font-size: 30px; */
-}
+} */
 
 
 /* @-moz-document url-prefix() {
--- a/lametro/static/css/city_custom.css
+++ b/lametro/static/css/city_custom.css
@@ -427,7 +427,7 @@ hr .events-line {
 @-moz-document url-prefix() {
     @media only screen and (max-device-width : 415px) and (orientation: portrait) {
         .blurb {
-            font-size: 2vh;
+            font-size: 35px;
         }
     }
 }
--- a/lametro/static/css/city_custom.css
+++ b/lametro/static/css/city_custom.css
@@ -424,10 +424,19 @@ hr .events-line {
     height: 200px;
 }
 
-@-moz-document url-prefix() {
+body {
+    /* text-size-adjust: none; */
+    -moz-text-size-adjust: none;
+    /* font-size: 30px; */
+}
+
+
+/* @-moz-document url-prefix() {
     @media only screen and (max-device-width : 415px) and (orientation: portrait) {
         .blurb {
-            font-size: 35px;
+            text-size-adjust: none;
+            -moz-text-size-adjust: none;
+            font-size: 30px;
         }
     }
-}
+} */
--- a/lametro/static/css/city_custom.css
+++ b/lametro/static/css/city_custom.css
@@ -423,20 +423,3 @@ hr .events-line {
     background-position: center center;
     height: 200px;
 }
-/*
-body {
-    /* text-size-adjust: none; */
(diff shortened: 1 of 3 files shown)


hancush (commented):
Interesting! Before we go a clever route, I wonder if fiddling with this property has any effect? https://developer.mozilla.org/en-US/docs/Web/CSS/text-size-adjust

xmedr:
Ah, ty! I tried this before and it didn't seem to work, but messing with it now, I'm finding you can use a `-moz-` prefix with it! So that does reset everything, and the default is that tiny text.

It looks like the default font size for firefox and chrome is a little small, would we want it to be a bit bigger or keep it as is? 

FireFox mobile, just the blurbs adjusted
image

FireFox mobile, everything adjusted
image

Chrome mobile, unchanged
image

lametro/templates/event/_related_bills.html
```diff
@@ -14,7 +14,7 @@ <h4>Board Reports</h4>
       {% with associated_bill=report.related_entities.all.0.bill %}
       <tr>
         <td><strong>{{ report.notes.0 | parse_agenda_item }}</strong></td>
-        <td>{{associated_bill.identifier}} {{report.description | short_blurb}} {{ associated_bill.inferred_status | inferred_status_label | safe }}</td>
+        <td class="blurb">{{associated_bill.identifier}} {{report.description | short_blurb}} {{ associated_bill.inferred_status | inferred_status_label | safe }}</td>
```
hancush, line 17: How about adding this class to the parent table, so all of its contents are a bit larger?

hancush (changes requested):
One tweak, and then let's go with this! I notice that the mobile site overall is pretty tiny. I'll open a separate issue to address that.

hancush:
Oh, actually, @xmedr, the site wide tiny-ness is because we're missing the viewport meta tag! How about adding that to `base.html` and seeing if it addresses the FireFox issue without having to add the additional style? `<meta name="viewport" content="width=device-width, initial-scale=1" />`

xmedr:
@hancush Yuuup, that was exactly it! Adding that fixed things without any other css, and uniformly enlarges text on the site overall

image

lametro/static/css/city_custom.css
```diff
@@ -423,3 +423,20 @@ hr .events-line {
...
+
+/* @-moz-document url-prefix() {
+    @media only screen and (max-device-width : 415px) and (orientation: portrait) {
+        .blurb {
+            text-size-adjust: none;
+            -moz-text-size-adjust: none;
+            font-size: 30px;
+        }
+    }
+} */
```
hancush, line 442: You can remove this dead code!

```suggestion
```

hancush (approved):
Nice! Please remove the dead code, then go ahead and merge.

lametro/static/css/city_custom.css
xmedr: Bah, thanks for catching that! That's what happens when I don't double check myself :expressionless:

PR #967, opened 2023-05-15
Fix typo causing longstanding Sentry error, handles #924

hancush:
## Overview

See title.

Connects #924

## Testing Instructions

 * Run the app locally, and view localhost:8001/events/?view=all – confirm the page loads without error.


xmedr (approved):
all good! page loads without error 👍

Issue #975, opened 2023-05-17
Consistently reference SmartLogic environment via settings


hancush:
We've got a setting for this, but have hard coded the UID in at least one spot. Let's fix that.

hancush:
PR for this is in review: #977

PR #977, opened 2023-05-17
Refactor SmartLogic interactions

hancush:
## Overview

This PR takes most of the code from #771, omitting logic to filter topics without any associated bills. In short, it:

- Creates a new, separate `smartlogic` app containing a SmartLogic client and Django views to retrieve relations and suggestions. These views return the raw responses from the SES API.
- Adds a LA Metro SmartLogic API view, which subclasses the upstream view from the `smartlogic` app, processes the responses, and returns only subjects we have in our database.
- Refactors the autocomplete and related terms JavaScript to use our endpoints, instead querying SES directly. This allows us to move that complex processing out of JavaScript, yahoo!

Connects #975

### Notes

Behavior should not change.

metro-search

## Testing Instructions

 * Configure your local app to use the v5 Test API. The environment and API key can be found by logging into cloud.smartlogic.com using the credentials in Bitwarden.
 * Retrieve subject GUIDs and classifications: `docker-compose run --rm app python manage.py refresh_guid`
 * Update your search index: `docker-compose run --rm app python manage.py update_index --batch-size=25`
 * Bring your app up (`docker-compose up app`), then navigate to the homepage and confirm autosuggestions and term selection work as expected in the search bar.
 * Run a search and confirm related terms are populated beneath the search filters.


lametro/static/js/autocomplete.js
```diff
@@ -9,59 +8,33 @@ var SmartLogic = {
 
     if ( (tokenNeeded || tokenExpired) ) {
       return $.get(
-        '/ses-token/'
+        '/smartlogic/token/'
```
hancush, line 11: The changes in this file are mostly related to factoring out the complicated response processing (it is now done in Python).

smartlogic/client.py
```diff
@@ -0,0 +1,86 @@
...
+    RequestFailed,
+    RequestNotAuthenticated,
+    AuthenticationFailed,
+    ResponseNotSerializable,
+)
+
+
+class SmartLogic(object):
+    BASE_URL = "https://cloud.smartlogic.com"
+    SERVICE_URL = f"/svc/{settings.SMART_LOGIC_ENVIRONMENT}/ses/CombinedModel"
```
hancush, line 27: Woooooo!

smartlogic/client.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def __init__(self, api_key, authorization=None):
+        self.api_key = api_key
+        self._authorization = authorization
```
hancush, line 31: Authorization needs to be provided by views using this client, which is great because it means you can't just request our smartlogic endpoints willy nilly.

smartlogic/views.py
```diff
@@ -0,0 +1,94 @@
...
+        super().setup(request, *args, **kwargs)
+
+        self.smartlogic = SmartLogic(
+            settings.SMART_LOGIC_KEY,
+            authorization=self.request.headers.get("Authorization", None),
+        )
+
+    def dispatch(self, request, *args, **kwargs):
+        try:
+            return super().dispatch(request, *args, **kwargs)
```
hancush, line 33: This calls `get_queryset`, which is where we make our calls to SmartLogic.

smartlogic/views.py
```diff
@@ -0,0 +1,94 @@
...
+        try:
+            return JsonResponse(context["object_list"])
+        except json.JSONDecodeError:
+            return JsonResponse(
+                {
+                    "status": "error",
+                    "message": f"Could not serialize object list: {str(context['object_list'])}",
+                },
+                status=500,
+            )
```
hancush, line 78: I'm not 100% sure this is necessary, but I kept it superstitiously.

docker-compose.yml
```diff
@@ -21,6 +21,9 @@ services:
       - .:/app
     environment:
       DJANGO_MANAGEPY_MIGRATE: "on"
+      # Uncomment the following variable to import the legislative
+      # boundaries that power the Board Members map
+      # DJANGO_MANAGEPY_IMPORT_SHAPES: "on"
```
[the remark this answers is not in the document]
xmedr: Ohhh thats a nice quality of life change. When would those boundaries get imported in production then, if this is commented out here?

xmedr (approved):
@hancush Instructions were clear and everything works as expected!

docker-compose.yml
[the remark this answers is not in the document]
hancush: Great question! Unlike many of our other apps, Metro uses a legacy deployment pattern to an Elastic Cloud Compute (EC2) instance on AWS via CodeDeploy. That pattern runs the scripts in the `scripts/` directory on every deployment. 

Specifically, this and other management commands get run here:

https://github.com/Metro-Records/la-metro-councilmatic/blob/736622f040a7829682429668a21fd98a1f865503/scripts/after_install.sh#L45-L48

Think of it like a release phase on Heroku, but in addition to loading data, it also triggers the processes that run and serve the app.

(You can read more about this pattern in our archived docs, here: https://github.com/datamade/deploy-a-site/blob/master/Zero-Downtime-Deployment.md)

Issue #969, opened 2023-05-15
Improve mobile display


hancush:
About 10% of our users visit the site on a mobile or tablet device. That might not sound like much, but it amounts to ~7,000 people this year. The quality of site display on smaller screens is pretty variable. Let's put some thought into improving display for non-desktop devices.

xmedr:
I did some snooping with browserstack and my phone to see how things are. Here are some thoughts on some pages. Lmk what you think @hancush, I can pick any or all of these up anytime:

- General
  - On browserstack devices other than galaxy s21, the top nav bleeds off of the page a bit, which makes it so that the logo and expanded nav items also render slightly off the page. My phone is an s21 and both mine and theirs seems to be fine here.
- Homepage:
  - Could possibly use dividers between listed meetings
- Board members listing:
  - The recent activity section displays fine when showing less, but when showing more, row items get pushed to opposite edges of the container
  - Related to that: would it be worth considering moving that show more button to the top of that container for mobile? It's potentially a lot to scroll through to get back to the show less button at the bottom if you accidentally clicked it or changed your mind.
  - The map that says "tap to view details" just takes you to the person detail page for that polygon instead of showing details. If this is a feature we want on mobile, then this might be a priority.
- About Us
  - page scrolls horizontally a bit for board report statuses table
- Board report search
  - After searching once, the collapsed nav disappears. comes back on refresh
  - Might be nice to have a separator between related search terms for clarity on mobile
   <img src="https://github.com/Metro-Records/la-metro-councilmatic/assets/114717958/88e7af32-e959-4370-8383-3995d2348533" alt="Girl in a jacket" width="200">

- Events listing
  - The top section with the date selector, list button and refresh could be a little more clear. Instructions in the date picker get cutoff and I initially wasn't sure what the bulleted list button was for
  <img src="https://github.com/Metro-Records/la-metro-councilmatic/assets/114717958/9452c1b3-6818-450a-b016-baa93387dc2c" alt="Girl in a jacket" width="200">

- Minutes search:
  - Same as top section of events listing.
  - The td showing what kind of meeting it was is blank. This also feels like a priority.
- Event detail:
  - The event's heading with details could use some horizontal space
  <img src="https://github.com/Metro-Records/la-metro-councilmatic/assets/114717958/de7b7b88-ca8e-474f-aafe-ed513696f5a5" alt="Girl in a jacket" width="200">

hancush:
[...]

PR #982, opened 2023-06-16
Refactor index and about, upgrade bootstrap

xmedr:
## Overview

This upgrades bootstrap to version 5 and refactors the index and about pages

Connects #969

### Notes

I have this comparing to the `bootstrap-mobile-adjustments` branch so that they can be merged into that piece by piece, and then this bootstrap branch can be merged into main once everything is done.

If there's a better way to do this, please lmk!

## Testing Instructions

 * Check the index and about pages against their live versions on desktop and mobile window sizes
 * Confirm that
   * they look presentable/same
   * the nav works
   * the functionality across those two pages still works (tooltips, table of contents, etc)


lametro/templates/base.html
```diff
@@ -9,7 +9,7 @@
 
     <link href="https://maxcdn.bootstrapcdn.com/font-awesome/4.4.0/css/font-awesome.min.css" rel="stylesheet">
     <link href='https://fonts.googleapis.com/css?family=Open+Sans:400,700,300italic' rel='stylesheet' type='text/css'>
-    <link rel="stylesheet" type="text/css" href="{% static 'css/bootstrap.journal.min.css' %}" />
+    <link rel="stylesheet" type="text/css" href="{% static 'css/bootswatch.min.css' %}" />
```
antidipyramid, line 12: @xmedr Looks like `bootswatch` is used for theming. What do we use this CSS file for?

lametro/templates/index/index.html
```diff
@@ -180,5 +199,10 @@ <h2 style="display: inline-block;">
 
         $('#beta-info').tooltip({html: true, trigger: 'focus click', placement: 'right'});
 
+        var tooltipTriggerList = [].slice.call(document.querySelectorAll('[data-bs-toggle="tooltip"]'))
+        var tooltipList = tooltipTriggerList.map(function (tooltipTriggerEl) {
+            return new bootstrap.Tooltip(tooltipTriggerEl)
+        })
```
antidipyramid, line 205: Can you say a little bit about what these lines do?

antidipyramid (commented):
The index and about pages look good on my end. I left two quick questions.

lametro/templates/base.html
xmedr: This `bootswatch.min.css` seems to contain all bootstrap 5 functionality, in addition to theming. I believe I also have a regular `bootstrap.min.css` in the css directory that isn't actually being used, just in case it turns out that bootswatch was actually missing some things. 

So far it's been sufficient on it's own, and I have a reminder for myself to delete that regular bootstrap file once I'm done. I can delete it now if that makes more sense!

lametro/templates/index/index.html
xmedr: For sure! So this is actually lifted straight from the [bootstrap 5 docs on enabling tooltips](https://getbootstrap.com/docs/5.0/components/tooltips/#example-enable-tooltips-everywhere). This was added for the tooltip above the search bar that describes suggested topics being a new feature.

I believe that this selects all the tooltips by the `data-bs-toggle` attribute and the `bootstrap.Tooltip` plugin generates the content and places it adjacent to the trigger.

lametro/templates/base.html
antidipyramid: @xmedr From the [Bootstrap docs](https://getbootstrap.com/docs/5.0/getting-started/contents/#css-files), it looks like you can just use `bootstrap.min.css` to get all the Bootstrap styles. 

Do we use themes? If not, we could just use the plain Bootstrap CSS file.

lametro/templates/index/index.html
antidipyramid: Cool, thanks!

antidipyramid (approved):

PR #988, opened 2023-06-27
Refactor board members listing and search pages

xmedr:
## Overview
This continues the template upgrade to bootstrap 5, this time for the board members listing and search pages. This also makes some small bug fixes and changes. Some highlights are:

#### Search
- The 'remove' button that renders next to the 'order by' sorting options on the search page wasn't actually removing the sort. This has been fixed. A condition has also been added so that if the order is by 'relevance', then the remove doesn't render because that is the default order anyway.
- The panels that were used to show and hide facets have been upgraded to accordions. The style was changed along with this, so they don't look identical. IMO I don't mind the way they look and animate, but if we want it to be closer to the original then I can make some adjustments.
- The tags that show up with search results can get very long and don't wrap, causing the page to have extra space to the right if they're present. My solution here is to give those tags a max-width at smaller screen sizes with an ellipsis representing the overflow.

#### Board members listing
- The map on the board members page said "Tap ward to view details" on mobile sizes, implying that it would do the same as the hover on desktop. Tapping actually brought you to the member's detail page, so the language has been changed on mobile sizes to reflect this.
- The 'show more' buttons for the Recent Activity and Board of Directors Meetings sections have been moved to the top of those areas. This is to prevent users having to scroll to the bottom of a potentially long list of items to collapse the list.

This one seems to be a bit chunkier than the last, apologies! I'll try to submit these in smaller sections. Bring on any questions you have.

Connects #969 

### Notes

I left a couple todo's in there for now, that I do intend on going back to.

## Testing Instructions

- Check the index and about pages against their live versions on desktop and mobile window sizes
- Confirm that
  - they look presentable/similar to the original
  - the functionality across those two pages still works (maps, clear filter buttons, faceting etc)


lametro/templates/search/_ordered_search_filter_items.html
```diff
@@ -5,11 +5,14 @@
     {% if count %}
         {% if name in selected_list %}
             <li class="small">
-                <a href="#" class="filter-value" data="{{facet_name}}_exact:{{name}}" title="{{ name }}"><strong>{{ name | short_topic_name }}</strong></a>
-                <a href ="#" class="remove-filter-value" data="{{facet_name}}_exact:{{name}}">
-                   <i class="fa fa-times"></i>
+                <a href="#" class="filter-value" data="{{facet_name}}_exact:{{name}}" title="{{ name }}">
+                    <strong>{{ name | short_topic_name }}</strong>
                 </a>
-                <span class="badge badge-facet pull-right">{{ count }}</span>
```
antidipyramid, line 12: The "Filter by Meeting" accordion seems to be missing the count.

Screen Shot 2023-06-28 at 4 03 58 PM

lametro/templates/search/_ordered_search_filter_items.html
```diff
@@ -19,7 +22,7 @@
         {% if name not in selected_list %}
             <li class="small">
                 <a href="#" class="filter-value" data="{{facet_name}}_exact:{{name}}" title="{{ name }}">{{ name | short_topic_name }}</a>
-                <span class="badge badge-facet pull-right">{{ count }}</span>
```
antidipyramid, line 22: I think it'd be nice to have the counts justified to the right.

Screen Shot 2023-06-28 at 4 05 35 PM

lametro/templates/board_members/board_members.html
```diff
@@ -18,134 +18,142 @@
...
-                        </tr>
-                    {% endfor %}
                     <tr>
-                        <td colspan="3" align="center">
+                        <td class="text-end text-lg-center" colspan="3">
                             <a href="" id="more-actions"><i class="fa fa-fw fa-chevron-down"></i>Show more</a>
                             <a href="" id="fewer-actions"><i class="fa fa-fw fa-chevron-up"></i>Show fewer</a>
                         </td>
                     </tr>
+                    {% for action in recent_activity|slice:":100" %}
```
antidipyramid, line 121: It might make sense to keep the "Show more" option at the bottom and just reduce the number of items shown on the page (100 is a lot of items to scroll through!).

What do you think?

antidipyramid (changes requested):

lametro/templates/board_members/board_members.html
xmedr: That sounds good! Maybe 25? Or is that too few?

lametro/templates/search/search.html
```diff
@@ -393,6 +381,14 @@ <h3 class="modal-links">
     var facetTerms = [];
     $.each(selectedFacets, function(facetName, facetValue) {
         topicFacets.includes(facetName) ? facetTerms.push(facetValue) : $.noop();
+
+        // Open the accordions for selected facets on page load
+        let accordionBtn = $("[data-bs-target=collapse_" + facetName + "]");
+        let accordionDiv = $("#collapse_" + facetName);
+
+        accordionBtn.removeClass("collapsed");
+        accordionBtn.attr("aria-expanded","true");  //TODO: this bit isn't working
```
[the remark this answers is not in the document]
xmedr: Thankfully the accordions were expanding correctly for me too, which is great. But I was concerned that the aria attribute wasn't updating and would lead to people with screen readers not being able to interact with it as expected.

But looking into it more, I was definitely just selecting `accordionBtn` incorrectly haha. This is all working as expected in one of the newer commits.

lametro/templates/search/_ordered_search_filter_items.html
xmedr: Ah you're right. There is a conditional in [`_search_filter.html`](lametro/templates/search/_search_filter.html) specifically for sponsorships (the Filter by Meeting section) that I carried over:

[`{% if facets.fields.legislative_session|length > 1 %}` ](https://github.com/Metro-Records/la-metro-councilmatic/pull/988/files/064d119fb3746964906d62104f00c4f8859465ac#diff-fc097a686fb9405f2f370e47babde9e2512d8b8ce3544fe02cc35daf7d7afdcdL32)

that was preventing counts from rendering for those sponsorships. Looking at what it's used for, it only changes whether or not to leave space for and render the count.

In a new commit I got rid of that conditional and now space is always allotted for the count. So it gets rendered if it exists and just leaves a bit of space if it doesn't. I figure it would help keep facet name lengths uniform. What do you think?

Example of if only some have a count:
image

lametro/templates/board_members/board_members.html
antidipyramid: Sure, I think 25 is reasonable.

xmedr: Awesome, got those changes in

antidipyramid (approved):

Issue #991, opened 2023-07-06
Using Google Translate impacts access to navigation


shrayshray:
I noticed today that when using the Google Translate tool embedded on the site, the tool covers up the top navigation. Can you look into what can be adjusted to ensure navigation options are still visible for people translating the site?
![image](https://github.com/Metro-Records/la-metro-councilmatic/assets/28063988/99f63be5-7c26-4129-a134-a77988826f7b)

hancush:
Done and deployed.

PR #992, opened 2023-07-11
Fix Google translate bar position

antidipyramid:
## Overview

See title.

Connects #991 

### Demo

FireShot Capture 025 - ಮುಖಪುಟ - ಮೆಟ್ರೋ ಬೋರ್ಡ್ - localhost

## Testing Instructions

 * Navigate to the home page
 * Select a language to translate the page to
 * Verify that the google translate bar doesn't block the navbar

### DIFF
--- a/lametro/templates/base.html
+++ b/lametro/templates/base.html
@@ -119,6 +119,22 @@
                 'google_translate_element');
         }
 
+		// Make sure the Google Translate toolbar doesn't block the navbar
+		const navBar = document.querySelector("nav"),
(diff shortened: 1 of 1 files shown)


hancush (commented):
Looking good! Have you tested on other browsers? We aren't transpiling JavaScript for compatibility in this app, so I think it would be prudent.

antidipyramid:
@hancush Yes-- it seems to work on Safari, Chrome, and Firefox.

hancush (approved):

PR #993, opened 2023-07-12
Refactor person detail and align logout btn w/other nav items

xmedr:
## Overview

This continues the template upgrade to bootstrap 5, this time for the person detail page.

Connects #969 

### Demo
image

### Notes
On the live site, the map disappears on small screen sizes, which was either because it was overflowing the page or we didn't think it necessary to have on those sizes.

I have it showing up responsively on all sizes here, since I believe it could be useful to have that reminder of the person's jurisdiction regardless of screen size. If we think it's still too cluttered, lmk and I can remove it!

There was also some change in the presentation of the person's image at different screen sizes

## Testing Instructions

- Check some members' detail pages against their live versions on desktop and mobile window sizes
- Confirm that
  - they look presentable/similar to the original
  - the functionality across those two pages still works (maps, switching btwn related committees and board reports etc)

### DIFF
--- a/lametro/static/css/city_custom.css
+++ b/lametro/static/css/city_custom.css
@@ -350,6 +350,10 @@ hr .events-line {
     .current-meeting-img {
         max-width: 75%;
     }
+
+    #person-detail-headshot {
+        max-height: none;
+    }
 }
 
 @media only screen and (max-width : 415px) {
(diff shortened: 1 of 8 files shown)


antidipyramid (approved):
Looks great to me!

PR #1009, opened 2023-08-17
upgrade to bootstrap 5, enhance mobile display and site accessibility

xmedr:
## Overview

This combines all the previous refactor pr's

Connects #969 

## Testing Instructions

 * Navigate through the site and make sure the styling looks consistent and the functionality still works

PR #1014, opened 2023-09-06
Add ability to manage alerts

### LINKED ISSUE #1010: Give Metro ability to add alerts
During the state of emergency, we added alerts to the top of the site. It would be great for Metro to be able to manage these themselves. Propose a plan to enable this functionality.

xmedr:
## Overview

This branch adds ui to allow logged in users to manage alerts. There is a collapsible alert manager on the homepage where users can:
- select the type of alert
- see a small preview of the styling
- enter a description
- view and delete existing alerts

Closes #1010 

### Demo
ezgif-1-fd05881540

### Notes

At the time of writing, the Bootstrap 5 changes in #1009 have yet to be merged. This presents a couple caveats:

- The alert styles "primary" and "secondary" listed in the dropdown are present in bs5, but not the version of bootstrap that's used by this branch, so alerts with those types won't be styled as of now. Those styles will display properly once that pr comes in.
-  I have inline styles here that I will replace with their corresponding bs5 classes once that pr gets merged in.

## Testing Instructions

 * Bring the container up to run the new migration
 * Log in to the site and navigate to the home page
   * Confirm that the alert management section does not show on the home page until after you've logged in
 * Click the show button to expand the alert section and click through some of the alert type choices
   * Confirm that the select box style changes to preview what the alert will look like
 * Enter an alert description and click Submit
   * Confirm that the alert gets added to the top of every page, and to the View Existing section of the alert manager
 * Click the delete button next to your newly added alert in the alert manager
   * Confirm that the alert disappears from the top


lametro/views.py
```diff
@@ -103,6 +106,29 @@ def extra_context(self):
...
+        else:
+            context["alert_form"] = AlertForm()
+
+        return context
+
+    def post(self, request, *args, **kwargs):
+        form = AlertForm(request.POST)
+        description_content = request.POST.get("description")
+
+        if description_content.isspace():
```
hancush, line 123: What about handling this validation on the form model a la https://docs.djangoproject.com/en/4.2/ref/forms/validation/#cleaning-a-specific-field-attribute? Then your view should Just Work™️ without having to do this manual handling.

lametro/views.py
```diff
@@ -103,6 +106,29 @@ def extra_context(self):
...
+        context = super().get_context_data(**kwargs)
+
+        if "invalid_form" in kwargs:
+            context["alert_form"] = kwargs.get("invalid_form")
+        else:
+            context["alert_form"] = AlertForm()
+
+        return context
+
+    def post(self, request, *args, **kwargs):
```
hancush, line 119: If you move validation to the form model, then you can extend [`form_valid()`](https://ccbv.co.uk/projects/Django/4.2/django.views.generic.edit/FormView/#form_valid) here, as you do in the AlertDelete view.

lametro/templates/index/index.html
```diff
@@ -172,5 +176,28 @@ <h2 style="display: inline-block;">
...
+            }
+            return false;
+        });
+
+        $("#id_type").on('change', function() {
+            $(this).removeClass()
+            if (this.value != "") {
+                $(this).addClass("alert-" + this.value)
+            }
+        });
```
hancush, line 200: Love this little touch!

hancush (changes requested):
This looks really good, @xmedr! A couple of suggestions about form validation inline.

lametro/views.py
xmedr: Ah gotcha, sounds good! In order to get that field specific validation to work, I changed the `description` field on the model to be optional, otherwise django would just run it's own validation and spit out a generic "This field is required" error message.

"Just Work™️" is something I'm going to carry with me for the rest of my career

lametro/views.py
xmedr: Cool, got this to work as well! I could have overthought this, but without a post method, I was getting an error `Method Not Allowed (POST): /`. I found that to get past this and also get the `form_valid()` to run, I could have the LAMetroIndexView also inherit from the FormView. So that's what's in there now.

Is that kosher or should I change this?

lametro/views.py
```diff
@@ -100,9 +106,16 @@ def extra_context(self):
...
         extra["form"] = LAMetroCouncilmaticSearchForm()
+        extra["alert_form"] = self.get_form()
 
         return extra
 
+    def form_valid(self, form):
+        success_url = self.get_success_url()
+        form.save()
+
+        return HttpResponseRedirect(success_url)
```
hancush, line 117: Apart from saving the form, this is the default behavior, so you can extend (add extra stuff to) the method instead of overriding (completely replacing) it:

```suggestion
        form.save()
        return super().form_valid()
```

lametro/views.py
```diff
@@ -1036,6 +1049,17 @@ def get_context_data(self, **kwargs):
         return context
 
 
+class AlertDeleteView(DeleteView):
+    model = Alert
+    success_url = reverse_lazy("index")
+
+    def form_valid(self, form):
```
hancush, line 1056: Ditto here.

hancush (commented):
Looking great! One tiny change inline.

hancush (commented):
Looking great! One tiny change inline.

lametro/views.py
hancush: Kosher!

lametro/views.py
xmedr: Ohh gotcha, I'll make sure I make more of a conscious decision to extend or override these kinds of methods!

lametro/templates/index/index.html
```diff
@@ -172,5 +175,47 @@ <h2 style="display: inline-block;">
...
+            switch(this.id) {
+                case "link":
+                    input.value += ("[this is a link!](https://url.com) ")
+                    break;
+                case "bold":
+                    input.value += ("**this is bolded text!** ")
+                    break;
+                case "italic":
+                    input.value += ("*this is italicized text!* ")
+                    break;
```
xmedr, line 216: I'm torn on whether to keep this bit here, what do you think of this?

I initially wanted something like github where you can highlight some text, and clicking one of the buttons shown below would wrap your selection in markdown, but couldn't get `window.getSelection()` to work on firefox. So scrapped that approach in favor of just having the buttons insert snippets of markdown that the user could edit to what they need.

Is this useful enough to keep the way it is now? I feel like I'm leaning towards either getting rid of that feature all together, or figuring out how to get it to work the way I imagined in the first place, but not sure.

image

hancush: I think it's a little idiosyncratic, but ultimately more ergonomic than their having to create the Markdown from scratch. I vote keep, though this is also a great question for the user.

xmedr:
@hancush I added some labels and a ui adjustment, but i imagine this should be good to go since they approved it during our meeting

Issue #1011, opened 2023-08-29
Searches for multiple tags yeild zero results


shrayshray:
I'm currently trying to locate reports tagged with both "Testimony" and "Public Hearing". I was expecting to be able to build a search for multiple tags in the search bar. When I tried to, I got zero results. Here are the steps I took:
1.  Type "Testimony" in the search bar and select the term from the suggestion drop down, then click search
2. Select "Result type" = "Tag Search", which yields 21 results. 
3. Go back to the search bar, type "Public Hearing" and select it from the suggestion drop down, then run the search, which yields 21 results. 
4. After verifying some of these results are tagged with both "Testimony" and "Public Hearing", select "Result type" = "Tag Search"
5. Get zero results. 
Am I incorrect in my assumption I should be able to build a search for multiple tags in the search bar?

I'm able to get the results I expect using this process:
1. Click on the tag "Testimony" either on a search result in the list, or in the side bar of a report
2. On the new results page, clear "Testimony" from the search bar but not from the filters
3. Type "Public Hearing" and select it from the suggestion drop down, then run the search, which yields 21 results. 
4. Select "Result type" = "Tag Search"
5. Get 4 results with both tags

hancush:
There are a couple of relevant sections in the code:

- On the front end, we parse the query from the search bar to send to the server [here](https://github.com/Metro-Records/la-metro-councilmatic/blob/70b6e171b9b0033511b55c43ddd8f87a417a8179/lametro/static/js/autocomplete.js#L139).
- On the back end, we convert the parsed query to a SearchQuerySet [here](https://github.com/Metro-Records/la-metro-councilmatic/blob/70b6e171b9b0033511b55c43ddd8f87a417a8179/lametro/forms.py#L63).

I suspect but am not 100% sure that one of these might not be behaving correctly.

hancush:
Fixed in #1019.

Issue #972, opened 2023-05-16
Search for Project results in error after clicking "tag search"


camachoo:
Type "project" in the search bar on the home page.
Click enter.
Click "Tag Search"
![image](https://github.com/Metro-Records/la-metro-dashboard/assets/28767738/eae49161-41e2-43a2-b9ef-e0d82c1ec626)

hancush:
I think this term is associated with so many board reports that it's breaking Solr. Not sure how to fix, but can look into it.

shrayshray:
I noticed the top tagging terms are all experiencing the same issue. See attached for report on top tagging terms (thanks to the reports Datamade is generating for us!). It does seem like an issue to have this occurring for the most popular terms. There is work to be done on our side to evaluate whether these terms are over-tagging, and on the Datamade side to see if this level of tagging can be accommodated. Even if over-tagging is occuring now, it's possible we'll have this amount of instances in the future.
[report-2023-05-08-board-report-tag-analytics.pdf](https://github.com/Metro-Records/la-metro-councilmatic/files/11503147/report-2023-05-08-board-report-tag-analytics.pdf)

hancush:
@antidipyramid is going to run the full scrape and index all board reports on the Heroku instance, then check to see whether the switch to Elasticsearch resolves this problem.

Let us know what you find out!

antidipyramid:
@hancush I'm getting this error after a full scrape:

<img width="621" alt="Screen Shot 2023-09-06 at 2 30 53 PM" src="https://github.com/Metro-Records/la-metro-councilmatic/assets/36973363/10bcd70b-b38a-4f00-956c-024f04f916ef">

Do we just have to increase storage on the AWS server?

hancush:
Try `sudo docker system prune` first, then run `df -ah` to see how much space you have left.

antidipyramid:
@hancush After `docker system prune`:

<img width="591" alt="Screen Shot 2023-09-08 at 9 38 47 AM" src="https://github.com/Metro-Records/la-metro-councilmatic/assets/36973363/e4fafbd2-2539-49a9-8d09-7f541631edc9">
<img width="472" alt="Screen Shot 2023-09-08 at 9 39 34 AM" src="https://github.com/Metro-Records/la-metro-councilmatic/assets/36973363/f629a077-d8fa-4e91-845f-79382d9628a6">

[...]

PR #1019, opened 2023-10-02
Fix tag search

antidipyramid:
## Overview

This PR fixes two tag search bugs:

1. Searches for multiple tags yielding no results
2. Searches for tags with many associated board reports error-ing out (e.g. `project`)

Connects #1011 
Connects #972 

## Testing Instructions

 * Make sure your local instance has enough board reports (maybe 100+?)
 * Navigate to the search page
 * Execute two tag separate tag searches (e.g. `budget` then `accounting`)
 * Search for the same two tags together (e.g. `budget` and `accounting`)
 * Verify that searching multiple tags yields only board reports with *all* searched tags
 * Verify that full text and normal search still works as expected

 * Verify that tag searches for common tags like `project` don't error out. Locally, you can verify that the search query in the terminal doesn't look like this: 
Screen Shot 2023-10-02 at 2 30 47 PM

I'll verify this behavior on staging as well.

### DIFF
--- a/lametro/forms.py
+++ b/lametro/forms.py
@@ -12,7 +12,7 @@ from haystack.query import EmptySearchQuerySet
 
 from councilmatic_core.views import CouncilmaticSearchForm
 
-from lametro.models import LAMetroBill, LAMetroPerson
+from lametro.models import LAMetroPerson
 
 
 class LAMetroCouncilmaticSearchForm(CouncilmaticSearchForm):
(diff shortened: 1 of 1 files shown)


hancush:
@antidipyramid Want to take a look at that failing test?

antidipyramid:
@hancush Fixed-- had to clean up some stray imports.

Issue #1026, opened 2023-10-04
Tag search doesn't include bill type matches


antidipyramid:
After merging in #1019, tag searches no longer include matches for bill type.

<img width="1400" alt="Screen Shot 2023-10-04 at 12 28 16 PM" src="https://github.com/Metro-Records/la-metro-councilmatic/assets/36973363/e34a254f-8c91-46d6-b047-fe9046ceb6ec">

hancush:
Fixed in #1027.

PR #1027, opened 2023-10-04
Add bill type matches to tag search results

antidipyramid:
## Overview

See title.

Connects #1026 

## Notes

The tags displayed below a search result:

Screen Shot 2023-10-04 at 1 10 44 PM

include not only formal "topics" but also locations, programs, policies, bill types, etc. 

I think it would be a good idea to match users' tag searches over all of those values. We could add a combined `tags` field to the search index to easily query tags with `sqs.filter(tags__in=[search_term])`.

## Testing Instructions

 * Search for a bill type (e.g. `Informational Report`)
 * Click `Tag search`
 * Verify that this yields all matching bill types
 * Verify that a tag search for a topic and a bill type works as expected

### DIFF
--- a/lametro/forms.py
+++ b/lametro/forms.py
@@ -67,10 +67,12 @@ class LAMetroCouncilmaticSearchForm(CouncilmaticSearchForm):
         ]
 
         topic_filter = SQ()
+        bill_type_filter = SQ()
         for term in terms:
             topic_filter &= SQ(topics__in=[term])
+            bill_type_filter &= SQ(bill_type__icontains=term)
 
-        sqs = sqs.filter(topic_filter)
+        sqs = sqs.filter(topic_filter | bill_type_filter)
 
         return sqs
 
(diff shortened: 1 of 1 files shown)


lametro/forms.py
```diff
@@ -68,7 +68,7 @@ def _topic_search(self, sqs):
 
         topic_filter = SQ()
         for term in terms:
-            topic_filter &= SQ(topics__in=[term])
+            topic_filter &= SQ(topics__in=[term]) | SQ(bill_type__in=[term])
```
hancush, line 71: > I think it would be a good idea to match users' tag searches over all of those values.

This is a great idea. As a stopgap to the combined index field, could we update this query to search all of the fields you mention?

hancush (changes requested):

hancush (approved):
Thanks! Let's get this important fix up on the live site.

Issue #1030, opened 2023-10-17
Home: Search suggestions not loading


shrayshray:
In the search bar on the Home page, search suggestions are not showing. I've tried several queries, and the drop down shows "Results could not be loaded." This is happening in both Firefox and Chrome. It is not happening on the Search page, just in the search bar on the Home page.
![image](https://github.com/Metro-Records/la-metro-councilmatic/assets/28063988/e4f6950d-2346-4376-9786-9bcedeb5da3e)

hancush:
Done in #1032

PR #1032, opened 2023-10-20
Fix search suggestions

antidipyramid:
## Overview

This PR fixes a bug where search suggestions were not loading correctly. 

Connects #1030

### Issue summary

When a new visitor loads the page we request a Smartlogic token for them. However, we weren't waiting for the token to be granted and stored before setting up the search bar, causing a 403 error whenever it tried to get suggestions. The token was fetched after the bar had been configured.

On subsequent page loads, the search suggestions functioned as expected since the token was already stored in the browser.

## Testing Instructions

 * Either open a private window or clear your browser's cookies/cache (note that in some browsers, e.g. Firefox, browser data persists across open private tabs)
 * Navigate to the homepage
 * Verify that you are able to load search suggestions

### DIFF
--- a/lametro/static/js/autocomplete.js
+++ b/lametro/static/js/autocomplete.js
@@ -1,267 +1,288 @@
 var SmartLogic = {
   query: {},
-  getToken: function () {
+  getToken: function() {
     tokenNeeded = !window.localStorage.getItem('ses_token')
 
     msInDay = 86400000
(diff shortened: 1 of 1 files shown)


xmedr (approved):
Nice, this works as expected and feels like a pretty elegant solution!

hancush:
@antidipyramid Are we waiting on anything to deploy this?

antidipyramid:
@hancush Nope this one is good!

Issue #1036, opened 2023-10-26
Alert manager not showing alerts on live site


xmedr:
The alerts appear on the site after a delay, but they don't appear in the manager for deletion

xmedr:
@hancush I'm having trouble reproducing this locally.

Since our thought is that caching is what's causing issues, I've changed my cache backend to `django.core.cache.backends.db.DatabaseCache` and ran `createcachetable`. I ran [a test in the shell](https://stackoverflow.com/questions/29707274/how-do-you-confirm-django-is-using-memcached/29709199#29709199) and confirmed it was working. Adding and removing alerts from here poses no issues yet, alerts get removed and displayed accordingly which isn't what I expected.

I then tried to see what would happen if I added a `cache.clear()` command after an alert gets [submitted](https://github.com/Metro-Records/la-metro-councilmatic/blob/d5c077d78cae394ec715638607409d91f0bf496d/lametro/views.py#L112) or [deleted](https://github.com/Metro-Records/la-metro-councilmatic/blob/d5c077d78cae394ec715638607409d91f0bf496d/lametro/views.py#L1042) in the views. From here, deleting an alert doesn't remove it from the top, but clears all previous alerts from the manager:
![image](https://github.com/Metro-Records/la-metro-councilmatic/assets/114717958/ebb02122-befc-470f-b1be-7eb43db066a8)

This is more in line with what Shelly's issue looked like before, but this is with the addition of `cache.clear()`. I feel like caching could still be the culprit, but maybe I'm not enabling it correctly?

hancush:
@xmedr Want to drop some time on the calendar to pair this week? I'll have time tomorrow onward.

xmedr:
@hancush Yes absolutely! I'll find a time for tomorrow then

PR #1039, opened 2023-11-02
Move alert rendering outside of the navigation cache

hancush:
## Overview

See title. This change enables realtime updating of alerts, which were previously not rendering immediately on the live site due to cache interference.

Connects #1036 

### Demo

alerts

### Notes

This project uses a special caching library that allows you to cache fragments of a template. (This feature has since become native to Django; our use of the library predates this change.) 

In order to cache template fragments, you wrap sections of HTML in the `{% cache ...options %}{% endcache %}` tag.

This bug was a result of retrieving the value of `alerts` and rendering alerts within the cached navigation block. By moving the call to `get_alerts` and display of existing alerts outside of the tag, alerts now update in realtime.

## Testing Instructions

 * Update the CACHES setting in your local settings_deployment.py file to:
 ```python
 CACHES = {
     "default": {
         "BACKEND": "django.core.cache.backends.db.DatabaseCache",
         "LOCATION": "councilmatic_cache",
     }
 }
 ```
 * Create a superuser in your local instance: `docker-compose run --rm app python manage.py createsuperuser`
 * Log in to your local instance at http://localhost:8001/metro-login
 * Navigate back to the homepage and create some alerts. Confirm they show up on page load at the top of the page and in the alert manager.
 * Delete some alerts and confirm they are removed from the top of the page and the alert manager.

### DIFF
--- a/lametro/templates/base.html
+++ b/lametro/templates/base.html
@@ -1,4 +1,4 @@
-{% load adv_cache static %}
+{% load adv_cache extras lametro_extras markdownify static %}
 
 <!DOCTYPE html>
 <html lang="en">
@@ -99,9 +99,8 @@
             </div><!-- /.navbar-collapse -->
         </div>
     </nav>
+    {% endcache %}
 
-    {% load extras lametro_extras static %}
-    {% load markdownify %}
     {% get_alerts as alerts %}
     {% if alerts %}
         {% for alert in alerts %}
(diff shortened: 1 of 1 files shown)


xmedr (approved):
Nice, this works perfectly! After the cache tag rundown, the issue makes a lot of sense

Issue #1021, opened 2023-10-03
Improve on WCAG "Perceivable" category - a11y audit


xmedr:
Go through the following criteria for [WCAG's Perceivable section](https://www.w3.org/TR/WCAG21/#perceivable) and improve their issues to meet WCAG requirements:

### Level A
- 1.1.1 Non-text Content:
  - [x] Let's get alt tags onto the thumbnails of the board members listing page table. The committee pages have listing tables and those have alt tags
  - [x] Figure out an accessible alternative to the maps on person detail pages. I've tried adding an alt tag to the map div directly, but VoiceOver didn't read it. Maybe using a caption underneath?
  - [x] The Metro logo in the nav needs an empty alt tag
  - [x] The DataMade logo on the About Us page needs an alt
  - [x] Many icons have an aria-hidden attribute for screen readers, but double check that all of them are taken care of, or given a label if screen readers do need to see them.
 
- 1.3.1 Info and Relationships:
  - [x] Various "Show More/Less" buttons can use the vocabulary of the thing they’re manipulating to make it clearer. Some people navigate using a list of all buttons or links on a page or view, so they might not encounter the surrounding context.
  - [x] Maybe under the board members listing page, the list of links of meetings that just say “Regular Board Meeting” can also include the date in the link?
  - [x] Change search results to use `ul` tags. [Here's a relevant SO answer](https://stackoverflow.com/questions/13009735/html5-tags-for-a-search-result-list#:~:text=You%20should%20use%20them%20only,an%20ordered%20one%20or%20not.)
  - [x] The `a` tags that are being used by the board members listing page and the search tooltip don't have an `href` attribute. We can just put `#`. Also check the rest of the `a` tags on the site.

- 1.3.3 Sensory Characteristics:
  - [x] Instructions use positional words like “below”. Positions can’t easily be determined by screen readers, so we can try using different language there.
  
- 1.4.1 Use of Color:
  - [x] We have a lot of links that are identified solely by color, we should consider making them distinct if, say, the user saw the page in grayscale. Changing the lightness of the links or marking them with underlines would be potential solutions. Though I'm not sure if having every one of those be underlined would look great.

### Level AA
- 1.4.3 Contrast (Minimum) (This section might be a fix as you go situation, but here's some places to start):
  - [x] Text color for links don't have enough contrast, including the ones in the index hero section
  - [x] Same for text color for search facet numbers
[...]

PR #1042, opened 2023-11-09
A11y Improvements - Perceivable

xmedr:
## Overview

This branch aims to improve on the "perceivable" category of WCAG standards for the app. Each of the areas of improvement are listed in the connected issue. Most notable changes are higher color contrast throughout the site, and the addition of underlines to links.

- Connects #1021

### Demo

image

### Notes
The original issues for each category are listed in #1021
- 1.1.1 Non-text Content:
  - I decided on adding an `aria-labelledby` as the alternative to the maps on board member detail pages. Screen readers will read the corresponding heading when focusing on those maps.
  - As far as the board member listing page maps, there's a couple alternatives to that map on that same page, so I've left that mostly untouched. There's both the Learn More tooltip above the map, as well as the table next to it that lists out all the people that would've been described by the map.
- 1.4.1 Use of Color:
  - I figured that since [github changed their default to having all links be underlined for accessibility's sake](https://github.blog/changelog/2023-10-18-new-default-underlined-links-for-improved-accessibility/#:~:text=By%20default%2C%20links%20within%20text,managing%20the%20appearance%20of%20links), then maybe metro can too. If the underlines are a bit much then we can try changing the lightness of the links instead
  - There are some links that probably don't need to be underlined since they don't navigate to another page, but instead manipulate something on the same page. This includes things like the show more links on the board members page. To fix this, in another category I'll be changing `<a>` tags that don't send you to a different page to `<button>` tags instead [as suggested in this article](https://www.a11y-101.com/design/button-vs-link).
image
- 1.4.13 Content on Hover or Focus:
  - Content inside the tooltips wasn't tabbable and the learn more icon's popup appeared at the end of the DOM, preventing screen readers from accessing it at an appropriate time. To fix both of those issues, I've added an aria descriptor to both those elements.

## Testing Instructions

 * Navigate the site (optionally with a screen reader)
 * Asses if the points covered in the issue have been properly addressed
 * Asses the changes made to colors and language

### DIFF
--- a/lametro/static/css/city_custom.css
+++ b/lametro/static/css/city_custom.css
@@ -755,3 +755,8 @@ a, .btn-link, .page-link {
 .btn-secondary {
     background-color: #6C757D;
 }
+
+.search-results {
+    list-style: none;
+    padding-left: 0px;
+}
--- a/lametro/static/css/city_custom.css
+++ b/lametro/static/css/city_custom.css
@@ -5,7 +5,7 @@ body, h1, h2, h3, h4, h5, h6, .h1, .h2, .h3, .h4, .h5, .h6, .navbar {
 }
 
 body, body .table, .popover-body, .badge-facet, .text-body-secondary {
-    color: #777777;
+    color: #6B6B6B;
 }
 
 h1, h2, h3, h4, h5, h6, .h1, .h2, .h3, .h4, .h5, .h6, .navbar, th {
@@ -198,10 +198,10 @@ strong.match-group {
   background-color: #EB6864;
 }
 
-.label-teal { background-color: #49A6AB; }
+.label-teal { background-color: #0E8389; }
 
 .label-received, .label-approved, .label-adopted {
-    background-color: #1D9A43;
+    background-color: #008027;
 }
 
 .label-active {
@@ -221,6 +221,7 @@ strong.match-group {
     margin-left: 1.8em;
     padding-top: 5px;
     padding-bottom: 5px;
+    color: #B82A25;
 }
 
 .nav-inline {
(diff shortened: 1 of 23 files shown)


antidipyramid (commented):
The color contrast and link underlines look great!

To be clear, we no longer show maps at all on board member detail pages?

xmedr:
Nice, I'm glad you like it! 

And we *should* still be okay to display maps on member detail pages, just the accessible alternative for screen readers is now going to be the heading above those maps. So for example on Karen Bass's page, instead of VoiceOver hovering over the map and reading "Map Open this area in Google Maps etc etc",  it should now read "City of Los Angeles map". But the map itself should still be present.

image

antidipyramid:
@xmedr OK cool, that's what I thought.

For some reason on my local instance I don't see a map on the board member detail pages. I tried clearing my cache/cookies but it's still blank.

FireShot Capture 028 - Janice Hahn - Metro Board - localhost

xmedr:
Oh interesting, that's strange. Sounds good, let me know if I can help

antidipyramid:
@xmedr Hmm that didn't fix it. Can you try clearing your cache/cookies and opening the page? If you can't replicate, we should pair quickly to see what's going on. (High likelihood something's off on my end!)

xmedr:
@antidipyramid  Alrighty, I cleared cookies/cache from both chrome and firefox, and also ran `cache.clear()` from the terminal, but I'm still getting the maps to show on my end. I'm absolutely down to pair on this! I'm cool with whenever's easiest for you

antidipyramid (approved):

Issue #1043, opened 2023-11-14
Updates to analytics reporting


shrayshray:
We'd like the following changes to tag analytics reports:
1. File naming convention to start with the date in this format - YYYY-MM-DD
2. Reports run on the first of every month. This seems to be what was established in #812, but it now looks like reports are running quarterly.

shrayshray:
@antidipyramid is this issue still in progress? The latest report I'm seeing are from 10-31-2023. We'd like to prioritize this fix.

antidipyramid:
We should make sure that the year comes first in the file name.

antidipyramid:
Hi @shrayshray, here are the three most recent analytics files on Google Drive. Is this how we expect the files to be named?

<img width="320" alt="Screenshot 2024-03-05 at 3 58 15 PM" src="https://github.com/Metro-Records/la-metro-councilmatic/assets/36973363/ab7914c7-bf03-4e24-b17a-a7caac880056">

PR #1054, opened 2024-01-05
Use environment variable to authenticate to Google Drive

antidipyramid:
## Overview

This PR transitions the app to use an environment variable to authenticate to Google Drive rather than the current JSON file.

Connects Metro-Records/la-metro-dashboard#123 #1043

## Testing Instructions

 * Decrypt `configs/lametro_service_acct_key.json` and copy the contents to `GOOGLE_SERVICE_ACCT_API_KEY` in `.env.local` (you'll have to move off this branch briefly to do + remove newline characters from the JSON string)
 * Set the `REMOTE_ANALYTICS_FOLDER` value in `.env.local` (you can use the value set in the staging app's Heroku env vars)
 * Run `python manage.py generate_tag_analytics`
 * Verify that the analytics file has been uploaded


councilmatic/settings.py
```diff
@@ -174,6 +174,8 @@
 # - Analytics
 ANALYTICS_TRACKING_CODE = env("ANALYTICS_TRACKING_CODE")
 SERVICE_ACCOUNT_KEY_PATH = "configs/lametro_service_acct_key.json"
```
fgregg, line 176: don't you want to remove this line?

fgregg (changes requested):
should we also remove references to settings_deployment that are sprinkled about, or do you want to save that for a separate clean up PR?

fgregg (changes requested):
we should be able to remove the keyrings directory altogether now.

fgregg:
> should we also remove references to settings_deployment that are sprinkled about, or do you want to save that for a separate clean up PR?

no, this should be done in a separate PR

antidipyramid:
@fgregg Any more comments on this one?

fgregg (approved):

Issue #1049, opened 2023-12-14
Long search queries cause Haystack/Elastic error


antidipyramid:
<img width="1440" alt="Screen Shot 2023-12-14 at 11 21 59 AM" src="https://github.com/Metro-Records/la-metro-councilmatic/assets/36973363/d94fd3ac-e86f-41c2-b66f-c9c84410b901">

It seems like this is usually caused by Elastic running out of memory (or at least, it thinks its running out of memory). Example query: `Executive Management Committee File Career Pathways`.

Some observations:
- this can be replicated locally and on Heroku
- on Heroku we've not even used half of the available disk space

Since this happens when a search is executed, I suspect this is caused by inefficient queries.

antidipyramid:
It looks like I can no longer replicate the error with the search string `Executive Management Committee File Career Pathways`.

We'll keep this one open for a little while in case it resurfaces as January traffic to the site picks up.

antidipyramid:
2/2 update: For some reason, I can again replicate this error on long search strings. After some testing, it seems like these errors coincide with memory leak-like behavior on the server. We've updated the server to try to mitigate this issue in #1072.

PR #1057, opened 2024-01-10
Fix slash escaping in searches

antidipyramid:
## Overview

A number of the search errors we've had are due to slashes in queries (e.g. `Westside Subway Extension/Purple Line Extension Phase 2`). It seems like we were double escaping `\` which caused an error in Elastic.

Connects #1049 

## Testing Instructions

 * Execute a search that includes a `/`
 * Verify that you don't get an error

### DIFF
--- a/lametro/forms.py
+++ b/lametro/forms.py
@@ -34,8 +34,8 @@ class LAMetroCouncilmaticSearchForm(CouncilmaticSearchForm):
             q += '"'
 
         # Escape reserved characters
-        reserved_characters = "|&*/\!{[]}~-+'()^:"
-        mapping = {char: f"\{char}" for char in reserved_characters}
+        reserved_characters = "|&*/\!{[]}~-+'()^:"  # noqa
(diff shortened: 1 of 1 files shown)


lametro/forms.py
```diff
@@ -34,10 +34,10 @@ def clean_q(self):
...
 
         # Escape reserved characters
-        reserved_characters = r"""|&*/\!{[]}~-+'()^:"""
-
-        for char in reserved_characters:
-            q = q.replace(char, r"\{}".format(char))
+        reserved_characters = "|&*/\!{[]}~-+'()^:"  # noqa
+        mapping = {char: f"\{char}" for char in reserved_characters}  # noqa
+        table = str.maketrans(mapping)
+        q = q.translate(table)
```
xmedr, line 40: TIL `maketrans()` is useful for swapping out characters, neat!

xmedr (approved):
works as expected!

Issue #1062, opened 2024-01-18
Boardagendas Site Application Error


neilarellano:
We are getting an Application Error when loading the site. Please see below:

![image](https://github.com/Metro-Records/la-metro-councilmatic/assets/110051034/1cce8e9f-b95b-4a88-8962-4d4bf5e5c1f1)

shrayshray:
This occurred during the 1/18/24 Operations, Safety and Customer Experience Committee meeting which began at aprox 1:05pm PST. The site was loading intermittently as of 1:22pm, and consistently as of 1:32pm. 
Per @antidipyramid via email, "It seems like a large number of people are trying to access the site right now. We’ve increased the server capacity and will continue to monitor the site as the meeting goes on.
We’ll also continue investigating what caused the bottleneck and how to avoid it in the future."

antidipyramid:
[Link to load test results](https://docs.google.com/document/d/1r7Byb14QHqBsoa6Y4Sd0JBT4Dd1dEJv8kNk2l5Vz7kE/edit)

antidipyramid:
2/2 update: we've doubled the server memory of our dynos on Heroku. So far, this seems to fix our memory woes.

PR #1068, opened 2024-01-24
Add sentry performance tracking

antidipyramid:
## Overview

See title. This will allow us to have more fine-grained analytics on app performance in Sentry.

Connects #1062 

## Notes

I've enabled the [Heroku Dyno Metadata add on](https://devcenter.heroku.com/articles/dyno-metadata) for both staging and production so we have access to `HEROKU_RELEASE_VERSION` and `HEROKU_APP_NAME` environment variables.


fgregg (approved):
cool! you might also want to add this github action to give sentry even richer info about releases.  https://github.com/datamade/chi-councilmatic/blob/main/.github/workflows/associate_commits.yml

Issue #718, opened 2021-04-14
Some searches mysteriously time out


hancush:
We've only seen one instance of this so far: http://boardagendas.metro.net/search/?order_by=desc&q=consolidated%20audit%20package%20a&search-all=on&sort_by=date

Sorts can be expensive, but, oddly, this search (sans sort) [only has 731 results](https://boardagendas.metro.net/search/?order_by=desc&q=consolidated%20audit%20package%20a&search-all=on). Larger result sets [can be sorted without issue](https://boardagendas.metro.net/search/?q=red+line+and+expo+reports&search-all=on&sort_by=relevance&order_by=asc).

hancush:
Don't see any errors in the Solr logs.

hancush:
Another: `se volvera a pagar pasaje`

hancush:
Pause barring update on ElasticSearch migration until December meeting.

antidipyramid:
Looks like this is still happening with ElasticSearch. 

My suspicion is that this is because both of these searches (`consolidated audit package a` and `se volvera a pagar pasaje` both contain the (very short!) word `a`. Searching `consolidated audit package e` also produces the same behavior. Might have something to do with finding every occurrence of such common strings, even in the middle of words.


![Image](https://github.com/Metro-Records/la-metro-councilmatic/assets/36973363/92a1e52e-a4a8-4ae5-8fb0-dfc9d4d6f426)

antidipyramid:
@antidipyramid will re-check on these problematic strings to see if we're still experiencing these errors.

xmedr:
@antidipyramid I've successfully tested these problematic strings today on staging and prod without errors. I also tried searching/sorting results for shorter terms like `e`, `a`, `the`, and `have`, all of which worked well.

PR #1074, opened 2024-02-07
Don't highlight tiny words in search results

antidipyramid:
## Overview

Previously we were using a more naive [highlighting scheme](https://github.com/datamade/django-councilmatic/blob/b550d380365a1e525f2f721cae5a480a4ef59da5/councilmatic_core/utils.py#L8) that would catch small words like "a" or "the". It's likely that this has been causing memory errors as we've set up Haystack to highlight all board report attachment text.

This PR attempts to make search result highlighting more efficient by only considering words with of >3 length. 

Connects #718


derekeder (approved):
this looks good. lets bring it in

Issue #1076, opened 2024-02-12
Home: Character limit, formatting and editing alerts


shrayshray:
A couple questions about the alerts (#1010) banner functionality: 
1. Is there a character limit on alerts?
2. Can we add line breaks in alerts? 
4. Can existing alerts be edited? Currently, any changes to an existing alert can only be made by creating a new alert and deleting the existing alert.

xmedr:
@shrayshray great questions!

1. The practical answer is no, there is no character limit for alerts. In reality there is a limit but it's super high, 1GB of text in total which translates to books worth of pages/words. Is a smaller character limit something you folks wanted in place?
2. We can absolutely add the ability for line breaks! I'll add a new button as well to make inserting those a bit easier.
3. You're right, as of now that's the only way to edit alerts. We'll work on adding an edit button to help with that!

xmedr:
@shrayshray update: we've got the functionality working for line breaks and editing (see gif below). we'll just be moving the alert manager to a dedicated page and it'll be good to go! with the addition of the edit form, we're hoping giving the manager its own page will help to clarify and declutter things.

![edit-alerts](https://github.com/Metro-Records/la-metro-councilmatic/assets/114717958/e6db703c-3e86-4871-991c-c56d1876537f)

shrayshray:
@xmedr looks great!

xmedr:
@shrayshray great! this has been added into the live site. we've added a new _Admin_ dropdown to the navbar when logged in, in case we add more of these kinds of features. and the alerts have been moved into a link there. here's an updated preview of what that looks like

https://github.com/Metro-Records/la-metro-councilmatic/assets/114717958/9fac9bc2-2290-465b-b3bf-f661beace82e

PR #1087, opened 2024-03-11
Allow for alert editing and line breaks

xmedr:
## Overview

This branch adds the ability to edit existing alerts within the alert manager. It also allows for line breaks to be added to alerts using `<br>` tags, and supplies a button to the editor do so.

- Connects #1076 

### Demo
edit-alerts

### Notes

There's an adjustment to the for loop that renders alerts, that makes sure the alerts stay in the same orde when edited. Without that `dictsort` filter, newly edited alerts may appear at the top of the list.

Let me know what you think about the UI/UX for this process. The alert manager was initially kept on the same page as the index for ease of use, so I tried to keep that pattern going with this new editor. The other option I could think of was to change the existing "add new" form to an "edit" form with some jquery when the edit button was clicked. We could also just move all of this to a separate page for simplicity's sake.

Also, because of the way the view is set up, I had trouble getting backend validation to make sure alerts don't have empty descriptions. If there's a way to use the AlertForm I previously created, that would be ideal, since that already has the validation. Otherwise I could change the Alert model to remove `null=True, blank=True` from the description.

## Testing Instructions

 * Login to the app
 * Create an alert with some line breaks using the editor button or by typing in the break tag `<br>`
   * Confirm that line breaks appear in the alert
 * Use the alert manager to edit the description and type, then save.
   * Confirm that the alert is changed
 * Optional: create/edit alerts and confirm they stay in the same spot at the top and on the list afterwards


xmedr:
@antidipyramid Ah I see what you mean. I do think moving it to a separate page would give it more space to breathe and be less cluttered by other elements. It'd also future proof it a bit in terms of code maintenance and ui design in the event more features get added to it later on.

I'm down to move it to another page! I'll get started on that when I've got more time next week. I'll also update folks in the issue

xmedr:
preview of the version split off on a separate page

https://github.com/Metro-Records/la-metro-councilmatic/assets/114717958/117dd436-05ae-44ed-8769-6d9032a3b17b

antidipyramid (approved):
Looks great!

A couple small UI tweaks for your consideration:

1. A navbar drop down titled "Admin" with "Alerts" as an option, in case we want to add more admin pages.
2. Putting the "Edit" and "Delete" buttons side-by-side.

Feel free to implement one or both of those if you agree.

Issue #1024, opened 2023-10-03
Improve on WCAG "Robust" category - a11y audit


xmedr:
Go through the following criteria for [WCAG's Robust section](https://www.w3.org/TR/WCAG21/#robust) and improve their issues to meet WCAG requirements:

### Level A
- 4.1.1 Parsing ([used this html validator](https://validator.w3.org/nu/?doc=https%3A%2F%2Flametro.datamade.us%2F)):
  - [ ] PDFs are structured a bit strangely and it affects screen reader performance. Instead of a paragraph being it's own `div`, one word or a whole phrase in the middle of a sentence will be.
  - [x] The calendar pdf also reads things out of order, but it doesn’t look like we have control over that.
  - [x] We’ve got some `img` and other tags that have stray, unneeded ending tags, like the Metro logo `img` tag
  - [x] There are buttons that have `href` attributes. Let's try to change those to just `a` tags
  - [x] Figure out if we need `autocomplete` attributes on radio buttons. If we do then keep them. If not, we can remove.
  - [x] Many `small` tags are encasing `div`s, whereas the `div`s should encase the `small`s
  - [x] Some table elements are missing a `scope` attribute
  - [x] The board report search bar has an `aria-describedby` that doesn’t lead to anything. This is identical to the issue for 3.3.2 Labels or Instructions
  - [x] Run the validator again to see what else can use some improvement.

- 4.1.2 Name, Role, Value:
  - [x] Consider replacing the `title` attribute on various elements with a `tooltip` instead. [Suggestion taken from this site](https://inclusive-components.design/tooltips-toggletips/).
  - [x] Many `div`s may be able to be changed into more semantic tags like sections, articles, etc.
  - [x] ~~The buttons above the board members listing page map are `label`s instead of `button`s. Consider making changing that to be more semantic.~~ These labels contain inputs, so this is semantically correct.

### Level AA
- 4.1.3 Status Messages:
  - [x] Can add an [aria role of alert](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA19) to errors

Connects #1012

xmedr:
Closed by #1088

PR #1088, opened 2024-03-26
A11y Improvements - Robust

xmedr:
## Overview

This branch aims to improve on the "robust" category of WCAG standards for the app. Each of the areas of improvement are listed in the connected issue.

This also adjusts the version of black that pre-commit uses to match the version used in github actions.

- Connects #1024

### Notes
The original issues for each category are listed in #1024

- 4.1.1 Parsing
  - There are a few errors that come up when using an html validator such as [the one provided by the W3C](https://validator.w3.org/), that weren't directly resolved in this branch
    - The `autocomplete` attribute on the search radio buttons raises errors in that validator, but I believe we can keep them there. [This github comment](https://github.com/squidfunk/mkdocs-material/issues/4023#issuecomment-1155490344) notes that it used to be required to set that attribute to "off" to prevent some browsers from persisting the choice across pages. The metro app handles that itself in a few different ways. I have a couple more resources on this topic if interested.
    - There are a number of errors that exist because of dependencies we're using. Some of these have solutions in updated versions, while others do not. I'll open issues for the ones that would benefit from an update.
  - Many `data` attributes were adjusted to be more specific, and the functions that use them have been adjusted as well.
  - Moment.js has been removed because it wasn’t effecting anything about the datepickers on either chrome or firefox, but was causing a validation error.

Dependencies that cause validation errors
- May be resolved with an update:
  - mozilla pdf.js
  - recaptcha
  - leaflet
  - datatables
- Cannot be resolved with an update:
  - select2
  - google maps
  - google translate
  - google programmable search engine
  - google tag analytics

## Testing Instructions

 * Get to the site
 * Asses if the points covered in the issue have been properly addressed.
   * Firefox has an add-on available called [HTML Validator](https://addons.mozilla.org/en-US/firefox/addon/html-validator/) that can be used to test pages on localhost
image


antidipyramid (approved):
Looks great!

I think the recaptcha scheme would be a particularly good site component for us to update. It'll have the side benefit of moving away from the old, super long URLs we generate for searches.

xmedr:
Merging this into main instead of #1009 since this has everything that pr does as well as resolved merge conflicts.

PR #1095, opened 2024-04-17
Enable review apps with the staging database

### LINKED ISSUE #1094: Consider review apps
The initial reason that we decided against review apps for metro is because of all the data that would need to get created every time an app does. To circumvent this, we could set the database connection to the staging/production app instead. For chicago councilmatic, they've ensured that migrations don't get applied to the database they're connecting to. This does mean that any changes that require migrations would still have to be reviewed locally.

> https://github.com/datamade/chi-councilmatic/blob/main/scripts/release.sh
we just add an env variable called `PRODUCTION`
the trick is you don’t want to run any of the database migrations when not on production
so there are some limitations to it

Reviewing different parts of this project involves making sure folks have all the data they need for the issue, and that occasionally presents its own challenge. Review apps would help alleviate this reviewing process.

xmedr:
## Overview

This alters the release script to look for either a signal that the app is running on either a production or staging environment.

The staging/production apps on heroku have a new boolean config var added to denote whether setup commands should be run.

Review apps have been set up with config vars nearly identical to the staging environment prior to opening this pr. They are also prepped to use the staging database instead of creating a new one every time a new review app gets made. The change in the release script is meant to prevent potential migrations on review apps from changing the staging db.

Smaller changes include updating the readme and clearing the cache on new releases. The nav would occasionally get formatted strangely and clearing the cache always solves this.

- Closes #1094 

## Testing Instructions

 * Confirm that a review app gets made with the functionality and data available on the staging site.
 * Run `heroku releases:output v6 -a la-metro-cou-feature-re-zibkyf` to see the output of the release marked as "Deploy"
   * Confirm that the output shows `PRODUCTION or STAGING is not set, skipping setup management commands` to note that migrations were not run

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -1,6 +1,13 @@
 #!/bin/bash
-set -euo pipefail
 
-python manage.py migrate --noinput
-python manage.py createcachetable
-python manage.py import_shapes data/final/boundary.geojson
+if [ -n "${PRODUCTION}" ] || [ -n "${STAGING}" ]; then
+    set -euo pipefail
+
+    python manage.py migrate --noinput
+    python manage.py createcachetable
+    python manage.py import_shapes data/final/boundary.geojson
+    python manage.py clear_cache
+
+else
+    echo "PRODUCTION or STAGING is not set, skipping setup management commands"
+fi
--- a/README.md
+++ b/README.md
@@ -325,6 +325,9 @@ docker-compose -f docker-compose.yml -f docker-compose.locust.yml run --service-
 This will start the Locust web server on http://localhost:8089. For more details,
 see the [Locust documentation](https://docs.locust.io/en/stable/).
 
+## Review Apps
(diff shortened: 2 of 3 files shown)


scripts/release.sh
```diff
@@ -1,6 +1,13 @@
...
-python manage.py migrate --noinput
-python manage.py createcachetable
-python manage.py import_shapes data/final/boundary.geojson
+if [ -n "${PRODUCTION}" ] || [ -n "${STAGING}" ]; then
+    set -euo pipefail
+
+    python manage.py migrate --noinput
+    python manage.py createcachetable
+    python manage.py import_shapes data/final/boundary.geojson
+    python manage.py clear_cache
```
derekeder, line 9: this line seems new. i wonder why the cache wasn't getting cleared before. seems fine to include now

derekeder (approved):
nice - confirmed this works as expected and looks good. yay review apps!

PR #1096, opened 2024-04-18
Add management command to clear cache

xmedr:
## Overview

Quick addition of a `clear_cache` command. The newest version of the release script in #1095 uses this command but deploying to staging raises an error because it currently doesn't exist.

## Testing Instructions

 * Run `python manage.py clear_cache` on the review app
 * Confirm that the output ends in a success message

### DIFF
--- a/lametro/management/commands/clear_cache.py
+++ b/lametro/management/commands/clear_cache.py
@@ -1,0 +1,19 @@
+# Adapted from https://github.com/rdegges/django-clear-cache
+from django.conf import settings
+from django.core.cache import cache
+from django.core.management.base import BaseCommand, CommandError
+
+
+class Command(BaseCommand):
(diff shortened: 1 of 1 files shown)


derekeder:
@xmedr do we need this command? would it be better to remove it from the deploy script?

xmedr:
@derekeder hm, no i don't think it's _needed_, but it might be a QOL addition. if we ever run into deployment issues that come about because of caching, we've just been going in and clearing it ourselves. but it does require that manual step after we notice it first.

i feel like @antidipyramid might have more insight on how useful this would be than i do, so i don't mind taking it out of the deploy script now and asking his opinion when he gets back!

derekeder:
ok lets go ahead and bring this in so our deploy script doesn't error. it is a standard command in our other projects

Issue #1104, opened 2024-05-07
FY25 Calendar


neilarellano:
Greetings!

Can we please add a link to the FY25 Calendar.
![image](https://github.com/Metro-Records/la-metro-councilmatic/assets/110051034/1e86f68c-33fc-453f-b4c7-1751237b7ff8)

Here is the PDF file for the calendar.
[FY25 Committee  Board Calendar.pdf](https://github.com/Metro-Records/la-metro-councilmatic/files/15238822/FY25.Committee.Board.Calendar.pdf)

Thanks!

antidipyramid:
Hi @neilarellano, just making sure: since the FY25 calendar starts in July, do we want to add a link to next year's calendar like this:

<img width="648" alt="Screenshot 2024-05-07 at 1 55 13 PM" src="https://github.com/Metro-Records/la-metro-councilmatic/assets/36973363/f5e0e345-0342-4abb-a420-37a88ac15bf7">

or replace the FY24 links with FY25 like this:

<img width="419" alt="Screenshot 2024-05-07 at 1 57 35 PM" src="https://github.com/Metro-Records/la-metro-councilmatic/assets/36973363/4be50e91-e44d-47df-8492-20fab1617b7d">

camachoo:
Please add it.  Make it so the FY 25 link stacks below the FY24 link.

example:
[cid:3dc2c334-5ada-4e1f-ac91-0bc908063fed]

________________________________

PR #1105, opened 2024-05-09
Add FY2025 calendar links

antidipyramid:
## Overview

See title.

Connects #1104 

### Demo

Screenshot 2024-05-09 at 3 31 41 PM

### Notes

I kept the calendar links inline on the events page because I thought extra whitespace would throw off the balance, but let me know if you think it'd look better stacked.

Screenshot 2024-05-09 at 3 28 03 PM

## Testing Instructions

 * Verify that links have been added and that they work


xmedr (approved):
all good to go! i agree with keeping it inline on the events page, looks better that way

Issue #1109, opened 2024-05-23
URGENT - Site Application Error During Board Meeting


neilarellano:
Hi Team!

The site will not load and errors out with an Application error. Can we please look into this ASAP?

antidipyramid:
We'll estimate the cost for autoscaling in Heroku (using Performance tier dynos).

antidipyramid:
We'll track future site optimization work in the related milestone.

PR #1110, opened 2024-05-23
Lower gunicorn timeout

antidipyramid:
## Overview

See title. [Heroku recommends a gunicorn timeout of 10-20 seconds.](https://devcenter.heroku.com/articles/preventing-h12-errors-request-timeouts#add-a-timeout-to-the-webserver)

Connects #1109

### DIFF
--- a/heroku.yml
+++ b/heroku.yml
@@ -14,4 +14,4 @@ release:
   image: web
 # The command that runs your application. Replace 'app' with the name of your app.
 run:
-  web: gunicorn --max-requests 1000 --max-requests-jitter 50 -t 180 --log-level debug councilmatic.wsgi:application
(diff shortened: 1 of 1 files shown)


heroku.yml
```diff
@@ -14,4 +14,4 @@ release:
   image: web
 # The command that runs your application. Replace 'app' with the name of your app.
 run:
-  web: gunicorn --max-requests 1000 --max-requests-jitter 50 -t 180 --log-level debug councilmatic.wsgi:application
+  web: gunicorn --max-requests 1000 --max-requests-jitter 50 -t 20 --log-level debug councilmatic.wsgi:application
```
antidipyramid, line 17: I chose to start with 20 seconds because we've got some slower database requests, especially for search queries. But we can also try lower.

hancush (approved):
Let's give it a shot.

PR #1113, opened 2024-05-31
Reduce homepage db queries

antidipyramid:
## Overview

This PR:
- optimizes the homepage by removing unnecessary database queries in the`LAMetroEvent.has_passed` property
- reduces noise in Sentry by no longer tracking performance of static asset pages

Connects #1109 

## Testing Instructions

 * Verify that tests pass

### DIFF
--- a/councilmatic/settings.py
+++ b/councilmatic/settings.py
@@ -203,6 +203,16 @@ if SENTRY_DSN:
 
     from councilmatic.logging import before_send
 
+    def custom_sampler(ctx):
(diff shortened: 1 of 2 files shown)


councilmatic/settings.py
```diff
@@ -203,6 +203,16 @@
 
     from councilmatic.logging import before_send
 
+    def custom_sampler(ctx):
```
hancush, line 206: Cool!

PR #1114, opened 2024-06-05
Optimize database calls on index and event detail page

antidipyramid:
## Overview

This PR removes duplicate database calls on the index and event detail pages. Specifically, it tackles database logic that's activated when there are meetings "today."

Locally, on a day with 4 meetings, homepage database queries reduced by 50% (44 -> 22) and event detail page queries reduced by ~35% (17 -> 11).

Connects #1109 

## Testing Instructions

 * Verify tests pass

### DIFF
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -838,7 +838,7 @@ class LAMetroEvent(Event, LiveMediaMixin, SourcesMixin):
     def is_ongoing(self):
         return self in type(self)._streaming_meeting()
 
-    @property
+    @cached_property
     def has_passed(self):
         if self.broadcast.exists():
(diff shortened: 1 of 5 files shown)


lametro/models.py
```diff
@@ -838,7 +838,7 @@ def is_upcoming(self):
     def is_ongoing(self):
         return self in type(self)._streaming_meeting()
 
-    @property
+    @cached_property
```
hancush, line 841: How does this work?

antidipyramid: `cached_property` caches a property value for just for the life of a model instance.

hancush: Oh, sick!

hancush (approved):

PR #1117, opened 2024-06-07
Optimize person detail page

antidipyramid:
## Overview

See title. Locally, I saw database calls on person detail pages reduce by about half. 

## Notes

Evaluating the Queryset returned by `LAMetroPerson.committee_sponsorships` on Heroku takes anywhere from several minutes to sometimes 10+ seconds (along with spiking memory usage). Since we're only evaluating the Queryset once, I dropped in a [iterator](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#iterator) to see if the lack of caching can improve the performance.

Connects #1109

## Testing Instructions

 * Verify tests pass

### DIFF
--- a/lametro/views.py
+++ b/lametro/views.py
@@ -707,7 +707,8 @@ class LAPersonDetailView(PersonDetailView):
             context["sponsored_legislation"] = []
 
         context["memberships_list"] = (
-            person.current_memberships.exclude(organization__name="Board of Directors")
+            person.current_memberships.prefetch_related("organization")
+            .exclude(organization__name="Board of Directors")
             .annotate(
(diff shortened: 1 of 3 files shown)


lametro/views.py
```diff
@@ -701,13 +701,15 @@ def get_context_data(self, **kwargs):
         except AttributeError:
             context["map_geojson"] = None
 
-        if person.committee_sponsorships:
-            context["sponsored_legislation"] = person.committee_sponsorships
+        sponsored_legislation = list(person.committee_sponsorships.iterator())
```
hancush, line 704: Beautiful.

Issue #1118, opened 2024-06-11
Granicus outage causes homepage to crash


antidipyramid:
This morning, a Granicus outage caused the homepage to crash. It seems like the running meetings endpoint resolved but took too long (>30 secs) to close the connection.

PR #1119, opened 2024-06-11
Cap running events request

antidipyramid:
## Overview

Granicus seems to be having intermittent issues with their running meetings endpoint. This PR aims to put a hard limit on requests to Granicus to make sure that our pages remain accessible regardless of the status of the running meetings endpoint.

Connects #1118

## Testing Instructions

 * Verify that tests pass
 * Verify that the review app's homepage works correctly

### DIFF
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -5,6 +5,8 @@ import logging
 import os
 from pathlib import Path
 import pytz
+import sys
+import time
 
 import requests
 from django.conf import settings
@@ -697,17 +699,32 @@ class LAMetroEvent(Event, LiveMediaMixin, SourcesMixin):
         queryset.
         """
 
+        def trace_function(frame, event, arg):
+            """
+            Makes sure we cap the request for running events.
+            See https://stackoverflow.com/a/71453648
+            """
+            if time.time() - start > TOTAL_TIMEOUT:
+                raise Exception("Request timed out.")
+
+            return trace_function
+
+        TOTAL_TIMEOUT = 5
+
         running_events = cache.get("running_events")
         if not running_events:
+            start = time.time()
+            sys.settrace(trace_function)
+
             try:
                 running_events = requests.get(
-                    "http://metro.granicus.com/running_events.php", timeout=5
+                    "http://metro.granicus.com/running_events.php", timeout=3
                 )
-            except (
-                requests.exceptions.ConnectTimeout,
-                requests.exceptions.ReadTimeout,
-            ):
+            except Exception as e:
+                logger.error(e)
                 return cls.objects.none()
+            finally:
+                sys.settrace(None)
 
             # Cache running events for one minute
             cache.set("running_events", running_events, 60)
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -40,7 +40,7 @@ from councilmatic_core.models import (
 from lametro.utils import (
     format_full_text,
     parse_subject,
-    get_url,
+    timed_get,
     LAMetroRequestTimeoutException,
 )
 from councilmatic.settings_jurisdiction import BILL_STATUS_DESCRIPTIONS, MEMBER_BIOS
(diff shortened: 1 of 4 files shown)


antidipyramid:
The running meetings endpoint isn't the only Granicus endpoint we rely on on the homepage (we also use it on event pages). I'm thinking that if we like this approach, we should make a custom `requests.get` function.

What do you think, @hancush?

hancush:
@antidipyramid Sure, we could make a reusable thing to use wherever we make external requests.

lametro/utils.py
```diff
@@ -106,3 +109,38 @@ def get_list_from_csv(filename):
...
+
+
+def get_url(url, params=None, **kwargs):
+    """
+    Convenience function to ensure GET requests that time out raise an exception.
+
+    See https://stackoverflow.com/a/71453648
+    """
+
+    TOTAL_TIMEOUT = kwargs.get("timeout", 5)
```
hancush, line 125: What do you think about making this configurable (i.e., defining it in settings.py)? We can set a reasonable default there.

lametro/utils.py
```diff
@@ -106,3 +109,38 @@ def get_list_from_csv(filename):
...
+
+    def trace_function(frame, event, arg):
+        """
+        Makes sure our request raises an exception if the total time from
+        start to finish exceeds TOTAL_TIMEOUT.
+        """
+        if time.time() - start > TOTAL_TIMEOUT:
```
hancush, line 132: If we don't define this in settings.py, let's define it inside this method, so it shares the same scope.

lametro/utils.py
```diff
@@ -106,3 +109,38 @@ def get_list_from_csv(filename):
         my_list = [row for row in reader]
 
     return my_list
+
+
+class LAMetroRequestTimeoutException(Exception):
+    pass
+
+
+def get_url(url, params=None, **kwargs):
```
hancush, line 118: How about something a bit more descriptive? `timed_get`? Open to ideas!

tests/test_requests.py
```diff
@@ -0,0 +1,16 @@
+import pytest
+from freezegun import freeze_time
+from datetime import datetime, timezone, timedelta
+
+from lametro.utils import get_url, LAMetroRequestTimeoutException
+
+
+def test_get_url(mocker):
```
hancush, line 8: Bless.

lametro/models.py
```diff
@@ -56,17 +61,21 @@ def api_source(self):
 
     @property
     def api_representation(self):
-        response = requests.get(self.api_source.url)
+        try:
+            response = get_url(self.api_source.url)
+            response.raise_for_status()
```
hancush, line 66: What does `raise_for_status` do for us here?

lametro/utils.py
```diff
@@ -106,3 +109,38 @@ def get_list_from_csv(filename):
...
+
+    TOTAL_TIMEOUT = kwargs.get("timeout", 5)
+
+    def trace_function(frame, event, arg):
+        """
+        Makes sure our request raises an exception if the total time from
+        start to finish exceeds TOTAL_TIMEOUT.
+        """
+        if time.time() - start > TOTAL_TIMEOUT:
+            raise LAMetroRequestTimeoutException("Request timed out.")
```
hancush, line 133: How about passing the message containing the problematic URL to the exception here, that way our logging is consistent?

hancush (changes requested):
Looking good! Few things line.

lametro/utils.py
antidipyramid: `timed_get` works for me!

lametro/models.py
antidipyramid: `raise_for_status` raises an `HTTPError` if the request returns a non-200 response. I just added it to keep all the logic in a try/except block.

lametro/utils.py
antidipyramid: Yeah, that's a good call.

hancush (approved):

Issue #1123, opened 2024-06-17
Update Barger and Hahn Headshots


neilarellano:
Hi Team!

Since we are unable to upload these, can you please upload it for us while we wait for the upload issue to get resolved. Can we please get this done soon? Thanks!

Janice Hahn
![Janice Hahn 2024](https://github.com/Metro-Records/la-metro-councilmatic/assets/110051034/67f82b16-15e7-4fc2-9e57-1a1b6ee3c753)

Kathryn Barger
![Kathryn Barger 2024](https://github.com/Metro-Records/la-metro-councilmatic/assets/110051034/d3eb7030-c1d9-4b41-9a16-a420790e11fb)

antidipyramid:
Hi @neilarellano-- the headshots have been updated.

PR #1124, opened 2024-06-18
Update Hahn, Barger, and Dutra headshots

antidipyramid:
## Overview

See title.

Connects #1123 

## Testing Instructions

 * Verify that the headshots for councilmembers Hahn, Barger, and Dutra have been updated


xmedr (approved):
all three display updated photos on the review app as expected 👍

Issue #1121, opened 2024-06-15
Agenda Items List Issue


neilarellano:
Hi Team,

The item list on agenda pages seem to not be available. I checked previous meetings that we know had the agenda item list and the list is not there.

xmedr:
@neilarellano We've pushed an update that brings back those agenda list items. Let us know if anything else comes up here!

<img width="1416" alt="image" src="https://github.com/Metro-Records/la-metro-councilmatic/assets/114717958/f30debb1-d3f2-4611-94f6-1f2c5a99bd7a">

PR #1127, opened 2024-06-21
Fix missing agenda list items

xmedr:
## Overview

This branch gives the event detail template access to the `documents` variable which it was expecting in order to render the `_related_bills` partial. This should allow the related board reports to show up again.

This also pins the version of pip used in docker to `24.0`. Version `24.1` was recently released (June 20th) and only allows pinning packages with an asterisk when using `==` or `!=` . Something like `package-name<=1.2.*` no longer works, which is what `textract` tries to do with one of its dependencies.

- Connects #1121

## Testing Instructions

 * Head to a past event's detail page within the review app
 * Confirm that the Board Reports heading appears as expected
 * Confirm that related bills show up if available

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -54,7 +54,8 @@ jobs:
         python-version: '3.10'
     - name: Install dependencies
       run: |
-        python -m pip install --upgrade pip setuptools
+        python -m pip install pip==24.0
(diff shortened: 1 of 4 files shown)


antidipyramid (approved):
Thanks, @xmedr!

PR #1128, opened 2024-06-28
Prevent manual broadcast button from appearing when one exists

xmedr:
## Overview

This is a fix for a bug in pr #1112. On staging, the logic for preventing a user from publishing another broadcast when one already existed was not working because the `manual_events` variable that the event template was expecting didn't exist. This sets that up properly.

This also disallows the option to publish a manual broadcast both for past events that already have a regular broadcast available, and cancelled events.

## Testing Instructions
**Note: since that last PR applied the migration for manual broadcasts, this can now be tested on the review app (using the staging database)

 * Publish an event
 * Go to another event's page. Confirm that:
   * An event **without** a proper broadcast does not allow publishing a second one, and provides a link to the initial event
   * An event **with** a proper broadcast or a **cancelled** event does not allow publishing a new one at all

### DIFF
--- a/lametro/templates/event/event.html
+++ b/lametro/templates/event/event.html
@@ -42,6 +42,8 @@
                     {% endfor %}
                 {% endif %}
                 {% if user.is_authenticated %}
+                    {% get_events_with_manual_broadcasts as manual_events %}
+
                     {% if not event_ok %}
                         <div class="well" style="margin-top: 20px">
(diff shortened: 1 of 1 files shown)


xmedr:
whenever this gets approved i'll test on staging proper one more time before deploying

antidipyramid (approved):

Issue #1073, opened 2024-02-07
Error Uploading Katy Yaroslavsky's Portrait


neilarellano:
I tried uploading a portrait for Director Yaroslavsky but I keep getting an error:
![image](https://github.com/Metro-Records/la-metro-councilmatic/assets/110051034/a5414835-1bf8-49f9-9158-78d8231140c0)

antidipyramid:
Hi @neilarellano, can you post/send us the image you're trying to upload?

neilarellano:
![Councilwoman Yaroslavsky - Portrait (Jan 2023)](https://github.com/Metro-Records/la-metro-councilmatic/assets/110051034/8a720340-e05b-471c-a49b-66a72819543c)

Hi @antidipyramid,

Here it is.

antidipyramid:
Update: this is going to require an update to the way we store headshots in the database. We're aiming to have a fix in by the end of the week.

antidipyramid:
Fundamentally, our issues with headshots stem from the fact that we're using the `Person.headshot` field that's defined in [`django-councilmatic`](https://github.com/datamade/django-councilmatic/blob/5b074f376667766e4b6dbf093871d294bb35fc51/councilmatic_core/models.py#L52C1-L62C1):

```python
class Person(opencivicdata.core.models.Person):

    ...

    headshot = models.FileField(upload_to='images/headshots',
                                storage=static_storage,
                                default='images/headshot_placeholder.png')
```

`Person.headshot` hard codes the local filesystem as the storage backend. `LAMetroPerson.headshot_url` tries to get around that by returning the URL in the bucket. After explicitly `councilmatic.custom_storage.MediaStorage` to save the headshot to the S3 bucket, we're stuffing the actual URL of the object in S3 into `LAMetroPerson.headshot`:

```python
    elif "headshot_form" in request.POST:
            form = PersonHeadshotForm(request.POST, instance=person)
            file_obj = request.FILES.get("headshot", "")
            ...
            person.headshot = self.get_file_url(request, file_obj)
            person.save()
```

This sort of works in templates because `LAMetroPerson.headshot_url()` returns the headshot `FileField` itself which converts to the raw URL of the bucket file we passed to it. But because the field isn't using the correct file system, other properties like `LAMetroPerson.headshot.path` or `LAMetroPerson.headshot.size` don't return correctly. 

`LAMetroPerson` is a proxy model so we can't override and change the `headshot` storage backend. 

[...]

Issue #1055, opened 2024-01-05
Remove documentation and code related to the old deployment pattern


fgregg:
includes 

- [docker-compose.deployment.yml](https://github.com/Metro-Records/la-metro-councilmatic/blob/main/docker-compose.deployment.yml)
- [test_settings_deployment.py](https://github.com/Metro-Records/la-metro-councilmatic/blob/main/configs/test_settings_deployment.py)
- [secrets.py.gpg](https://github.com/Metro-Records/la-metro-councilmatic/blob/main/lametro/secrets.py.gpg)

and the README

PR #1134, opened 2024-07-08
Upload headshots to S3 via django-storages

hancush:
## Overview

This PR makes use of an upstream change to `django-councilmatic` allowing the headshot storage backend to be configured by Councilmatic instances. When credentials are provided, it configures the Metro app to store and retrieve headshots from S3.

This PR also performs some housekeeping (#1055):

- Clean up files from old deployment strategies
- Update README to reflect current stack and development practices
- Document environment variables in `app.json`
- Default environment variables to development settings

Connects #1073

### Notes

Depends on new release of django-councilmatic. Current draft pinned to branch with change.

https://github.com/datamade/django-councilmatic/pull/292

## Testing Instructions

- Log in to the review app using the staging credentials in Bitwarden.
- View the board member listing and confirm headshots are rendering. (Most are placeholders. Janice Hahn has a real one.)
- Navigate to a board member and upload a new headshot. Confirm that it appears on their detail page and the listing page.


requirements.txt
```diff
@@ -1,4 +1,5 @@
-django-councilmatic[convert_docs]>=3.0
+# django-councilmatic[convert_docs]>=3.0
+git+https://github.com/datamade/django-councilmatic.git@hcg/configurable-headshot-backend#egg=django-councilmatic
```
hancush, line 2: TODO: Update this to new `django-councilmatic` release once it becomes available.

Dockerfile
```diff
@@ -1,20 +1,11 @@
...
-
 FROM python:3.10-slim-bullseye
 LABEL maintainer "DataMade <info@datamade.us>"
 
 RUN apt-get update && \
     apt-get install -y libpq-dev gcc gdal-bin gnupg && \
     apt-get install -y libxml2-dev libxslt1-dev antiword unrtf poppler-utils \
                        tesseract-ocr flac ffmpeg lame libmad0 libsox-fmt-mp3 \
-                       sox libjpeg-dev swig libpulse-dev curl && \
+                       sox libjpeg-dev swig libpulse-dev curl git && \
```
hancush, line 8: Need `git` to install `django-councilmatic` from the branch, will remove when new release is cut.

councilmatic/settings.py
```diff
@@ -8,8 +8,33 @@
 
 env = environ.Env(
     # Set default values
-    DEBUG=(bool, False),
-    LOCAL_DOCKER=(bool, False),
+    LOCAL_DOCKER=(bool, True),
```
hancush, line 11: We default to development settings in other apps. Think it's a good idea to preserve this pattern and keep the size of the local env file small, since these are overridden in deployment already.

councilmatic/settings.py
```diff
@@ -19,7 +44,7 @@
 environ.Env.read_env(os.path.join(BASE_DIR, ".env.local"))
 
 SECRET_KEY = env("DJANGO_SECRET_KEY")
-DEBUG = env.bool("DEBUG")
+DEBUG = env.bool("DJANGO_DEBUG")
```
hancush, line 47: It's conventional to prefix settings with the app they're relevant to.

councilmatic/settings.py
```diff
@@ -185,11 +210,23 @@
 GOOGLE_API_KEY = env("GOOGLE_API_KEY")
 
 # - AWS
-AWS_KEY = env("AWS_KEY")
-AWS_SECRET = env("AWS_SECRET")
+AWS_S3_ACCESS_KEY_ID = env("AWS_S3_ACCESS_KEY_ID")
+AWS_S3_SECRET_ACCESS_KEY = env("AWS_S3_SECRET_ACCESS_KEY")
+AWS_STORAGE_BUCKET_NAME = env("AWS_STORAGE_BUCKET_NAME")
```
hancush, line 215: Django Storages automatically discovers these variables.

lametro/models.py
```diff
@@ -454,11 +454,7 @@ def ceo(cls):
     @property
     def headshot_url(self):
         if self.headshot:
-            # Assigning this to image_url would make 'return static(image_url)'
-            # at the end of this property concatenate 'static' to the url.
-            # Returning here solves that.
-            if "headshot_placeholder.png" not in self.headshot.path:
-                return self.headshot
+            return self.headshot.url
```
hancush, line 457: We previously needed to load all images with `static()` because `django-councilmatic` was storing them with static files. Now, since we've told `django-councilmatic` to upload headshots to S3, if there is a headshot, we can return its URL directly, and only get the manual headshots and placeholder from static.

lametro/views.py
```diff
@@ -668,7 +667,7 @@ def post(self, request, *args, **kwargs):
                     self.get_context_data(form=form, headshot_error=error)
                 )
 
-            person.headshot = self.get_file_url(request, file_obj)
+            person.headshot = file_obj
```
hancush, line 670: `person.headshot` is a file field, so we can pass it this file directly.

councilmatic/settings.py
xmedr: neat! that makes things easier. if we're removing these non-S3 specific vars here, what do you think of these heroku config var adjustments as well?

**staging** - in addition to `AWS_KEY` and `AWS_SECRET`, there's also `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` which seem to hold duplicate values of their S3 counterparts. i imagine these can all be removed

**production** - the S3 versions do not exist, so we can likely change `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` to match the S3 var names. django storages also finds these automatically, but i'm unsure if the mismatch would cause issues considering these `settings` are looking for the S3 specific versions.

review apps are using only the S3 specific variables, so no change there

README.md
xmedr, line 1: line 17 here refers to a legacy way of running the app without docker. is this still useful to us? those old docs talk about using Solr which has been removed here, so they're also out of date. If we want to keep the reference to the old magicks, we could maybe change the language to something like:

 >These days, we run apps in containers for local development. More on that [here](https://github.com/datamade/how-to/docker/local-development.md). **For reference, if you'd like to read up on how we used to run this app locally,** see the [legacy setup instructions](https://github.com/datamade/la-metro-councilmatic/blob/b8bc14f6d90f1b05e24b5076b1bfcd5e0d37527a/README.md).

The line in question:
https://github.com/Metro-Records/la-metro-councilmatic/blob/04df15e524a76cde55913f6b8d9c8de95851b5b9/README.md?plain=1#L17

xmedr (changes requested):
the README updates and env var cleanup are very much appreciated! and the headshot uploads work as expected. just a couple questions inline

README.md
hancush: Great call. Since we don't run apps without Docker any more, I just removed the reference to the old setup instructions.

hancush:
@xmedr Yes, those config var changes are a good idea. Best to have everything standardized.

antidipyramid (approved):
Works great. Thanks for also including the README updates!

xmedr (approved):
sweet, i just made those config var changes for prod and staging, so this looks good to me!

PR #1135, opened 2024-07-09
Update issue templates

antidipyramid:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,27 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: An actionable title e.g. "Update the icon for..." or "Fix the behavior of..."
```
antidipyramid, line 4: I think this is what will appear in the title section of the issue as a placeholder.

xmedr (approved):
this looks great!

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,27 @@
+---
+name: Bug report
+about: Create a report to help us improve
```
hancush, line 3: How about something more specific, like "Let us know that the board agendas site is behaving differently than expected"?

.github/ISSUE_TEMPLATE/bug_report.md
hancush: I think for bugs, the actionable thing is always going to be "Fix the behavior of..." What if, instead, we request a summary of the unexpected behavior, e.g., "Uploading board member headshot results in an error"?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,27 @@
...
+title: An actionable title e.g. "Update the icon for..." or "Fix the behavior of..."
+labels: ''
+assignees: ''
+
+---
+
```
hancush, line 9: How about a link to this resource from Mozilla? https://bugzilla.mozilla.org/page.cgi?id=bug-writing.html

hancush (changes requested):

hancush:
Superseded!

PR #1146, opened 2024-07-24
Make sort_name sortable 🙃

hancush:
## Overview

This PR updates the configuration of the sort_name field so ElasticSearch will actually sort by it. Reference: https://github.com/django-haystack/django-haystack/issues/569

Connects #1142 

## Testing Instructions

This change requires an index rebuild to take effect, so you'll have to test locally.

- Run the app as instructed in the README. Make sure to follow the instructions for populating your search index.
- Run a search.
- On the results page, sort by title and confirm the sorted results load without error.


councilmatic/settings.py
```diff
@@ -32,8 +32,10 @@
     AWS_S3_ACCESS_KEY_ID=(str, ""),
     AWS_S3_SECRET_ACCESS_KEY=(str, ""),
     AWS_STORAGE_BUCKET_NAME=(str, ""),
-    RECAPTCHA_PUBLIC_KEY=(str, ""),
-    RECAPTCHA_PRIVATE_KEY=(str, ""),
+    # Test keys from:
+    # https://developers.google.com/recaptcha/docs/faq#id-like-to-run-automated-tests-with-recaptcha.-what-should-i-do
+    RECAPTCHA_PUBLIC_KEY=(str, "6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI"),
```
antidipyramid, line 37: Locally, I had to remove the corresponding (empty) recaptcha variables from `.env.local` to get this to work. `django-environ` seems to parse empty variables as present and won't use the default. 

We should probably remove the these variables from `.env.local.example`.

antidipyramid (approved):
Works great! Just one comment.

councilmatic/settings.py
hancush: Great catch. Done!

Issue #1151, opened 2024-07-24
Reduce the size of search results


hancush:
One search result can be several MB in size. That can really add up! I suspect the biggest culprits are the full text fields. Figure out a way to index that information, but not return it in the results. Looks like ElasticSearch provides for this, if django-haystack does not: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-fields.html

hancush:
Also related to efforts to reduce page load time.

antidipyramid:
It doesn't look like there's a way to add a `fields` parameter to the Elastic query with Haystack. We can try extending Haystack's `Elasticsearch7SearchBackend` class.

antidipyramid:
@hancush I tested this on staging and found that it didn't lower the size of the index by very much. I've read that elastic uses a compression algorithm by default in indexes, so that may be why.

 It seemed to reduce memory usage when searching, but not by much either.

antidipyramid:
Update: I was able to get Elastic to exclude the text fields of bills from search results by [customizing Haystack's Elastic backend](https://github.com/Metro-Records/la-metro-councilmatic/blob/74c6211961af5bacdd94eee4676da79a89730f53/lametro/custom_search.py#L11-L68). I deployed that branch and found that, again, it reduced memory usage somewhat but not enough to be considered a fix.

PR #1152, opened 2024-07-25
Don't store full text in search index

antidipyramid:
## Overview

This PR overrides `django-councilmatic`'s `BillIndex.full_text` to no longer store a board report's `full_text` in the index.

Connects #1151 

## Testing Instructions

This PR requires an index rebuild and should be tested locally. 

* Run `docker compose run --rm app python manage.py rebuild_index`
* Verify that search still works as expected

Once this is merged in, I'll rebuild the staging database's index and see if this addresses our memory problems when searching.


hancush (approved):
✨

Issue #1138, opened 2024-07-23
Reduce page load times on meeting days


antidipyramid:
## Overview

Many site pages slow down considerably on meeting days. There are various code blocks (many duplicated) sprinkled throughout views and templates that trigger requests to Granicus. 

## Current Benchmarks

Performance on 7/18/24:

| Page | P95 |
| --- | ----------- |
| / | 983.79ms |
| /events/ | 826.56ms |
| /event/{slug} | 759.09ms |

## Goal Benchmarks

We should shoot to resolve 95% of requests in under 500ms.

PR #1156, opened 2024-08-20
Don't load all events on default events page

antidipyramid:
## Overview

Previously, `LAMetroEventsView` was iterating over and then serializing all events, both past and future, each time the page was loaded. The template simply truncated the events listing. I believe part of the high load times on the events page was caused by serializing so many events for each request.

This PR aims to reduce the load time of the main events page by only computing and serializing three days of events by default. The user can then request to view all future or past meetings.

Connects #1138 

## Benchmarks

I added a [template profiler](https://github.com/node13h/django-debug-toolbar-template-profiler) to Django Debug Toolbar to be able to benchmark template load times locally.

This PR reduced template load time on the default events page to ~40ms from ~475ms (-91%). Of course, showing all events will always greatly increase load time. But even that is faster now that we're only serializing either all past or future events instead of both.

### Notes

The inclusion of both future and past events on the same page leads to a little quirkiness in the code and a less than ideal UX. I think we could simplify things and lean more on Django's built-in `Paginator` in `ListView`s by separating out the future and past event listings. 

The default events page could stay the same. But the "Show upcoming meetings" button, for example, could lead to a separate page with pagination. This would also avoid ever having to serialize thousands of events in one request.

## Testing Instructions
 * Verify that the date picker on the events page still functions
 * Verify that clicking the show upcoming/past meetings buttons work as expected

### DIFF
--- a/lametro/templates/events/events.html
+++ b/lametro/templates/events/events.html
@@ -79,7 +79,7 @@
 								{% if 'show' in request.GET and request.GET.show == 'future' %}
 									<a href="{% url 'events' %}" class="btn btn-salmon" id="fewer-upcoming-events"><i class="fa fa-fw fa-chevron-up"></i>Show fewer upcoming meetings</a>
 								{% else %}
-									<a href="{% url 'events' %}?show=future#future-meetings" class="btn btn-salmon" id="more-upcoming-events"><i class="fa fa-fw fa-chevron-down"></i>Show all upcoming meetings</a>
+									<a href="{% url 'events' %}?show=future#upcoming-meetings" class="btn btn-salmon" id="more-upcoming-events"><i class="fa fa-fw fa-chevron-down"></i>Show all upcoming meetings</a>
 								{% endif %}
 
                 <h2 id="past-meetings" style="margin-top: 40px;"><span>Past {{ CITY_VOCAB.EVENTS }}</span>
--- a/lametro/templates/events/events.html
+++ b/lametro/templates/events/events.html
@@ -50,7 +50,7 @@
             {% endif %}
 
             {% if future_events %}
-                <h2><span>Upcoming {{ CITY_VOCAB.EVENTS }}</span>
+                <h2 id="upcoming-meetings"><span>Upcoming {{ CITY_VOCAB.EVENTS }}</span>
                 <br class="non-desktop-only"/>
                 <small>
                     <a href="rss/" title="RSS feed for Upcoming and Recent Events">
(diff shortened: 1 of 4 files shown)


lametro/templates/events/events.html
```diff
@@ -76,10 +76,13 @@ <h2><span>Upcoming {{ CITY_VOCAB.EVENTS }}</span>
                     </div>
                 {% endfor %}
 
-                <a href="" class="btn btn-salmon" id="more-upcoming-events"><i class="fa fa-fw fa-chevron-down"></i>Show all upcoming meetings</a>
-                <a href="" class="btn btn-salmon" id="fewer-upcoming-events"><i class="fa fa-fw fa-chevron-up"></i>Show fewer upcoming meetings</a>
+								{% if 'show' in request.GET and request.GET.show == 'future' %}
+									<a href="{% url 'events' %}" class="btn btn-salmon" id="fewer-upcoming-events"><i class="fa fa-fw fa-chevron-up"></i>Show fewer upcoming meetings</a>
+								{% else %}
+									<a href="{% url 'events' %}?show=future#upcoming-meetings" class="btn btn-salmon" id="more-upcoming-events"><i class="fa fa-fw fa-chevron-down"></i>Show all upcoming meetings</a>
+								{% endif %}
```
hancush, line 83: What do you think about breaking this logic out into a partial that can be used for both future and past events?

lametro/templates/events/events.html
```diff
@@ -174,53 +181,8 @@ <h2><span>All {{ CITY_VOCAB.EVENTS }}</span>
     <script type="{% static 'js/lib/bootstrap-datetimepicker.js' %}"></script>
     <script>
     $(document).ready(function() {
-
         $('#from').datepicker();
         $('#to').datepicker();
-
-        function collapseUpcomingEvents(){
```
hancush, line 181: You love to see it.

lametro/views.py
```diff
@@ -379,54 +378,52 @@ def day_grouper(x):
...
-            for event_date, events in itertools.groupby(past_events, key=day_grouper):
-                events = sorted(events, key=attrgetter("start_time"))
-                org_past_events.append([date(*event_date), events])
-
-            context["past_events"] = org_past_events
+            org_past_events = itertools.groupby(past_events, key=day_grouper)
+
+            # Only show three days of events by default
+            past_events = itertools.islice(org_past_events, 3)
+            future_events = itertools.islice(org_future_events, 3)
```
hancush, line 413: Hmm, rather than retrieving all events, grouping, and the limiting to three days, what if we filtered the queryset for events on the most recent past three days? Something like:

```python
media_events.filter(start_date__in=LAMetroEvent.objects.filter(
    start_time__lt=timezone.now()
).order_by("-start_time")[:3])
```

And then applying grouping?

lametro/views.py
```diff
@@ -379,54 +378,52 @@ def day_grouper(x):
...
+            show_all_future_events = self.request.GET.get("show") == "future"
+            if show_all_past_events:
+                past_events = org_past_events
+            elif show_all_future_events:
+                future_events = org_future_events
+
+            context["future_events"] = [
+                (d, list(events)) for d, events in future_events
+            ]
+            context["past_events"] = [(d, list(events)) for d, events in past_events]
```
hancush, line 426: Rather than loading these into memory, is it possible to pass the generator to the template and iterate it there instead?

hancush (changes requested):
Looking good, @antidipyramid. A few questions inline.

lametro/views.py
antidipyramid: That would be ideal but the template throws an error when passed the generators.

I think this is one of the reasons we should eventually move to a more standard pagination approach for the all past/future events view.

lametro/templates/events/events.html
antidipyramid: Hm, since the block is fairly small I'm thinking it might be clearer to keep it inline? I imagine the partial would also require an extra variable to know whether it's meant to render the upcoming or previous meeting buttons.

hancush: Good point!

hancush (approved):

hancush:
Stoked for these changes to come in, @antidipyramid! Thanks for this.

PR #1161, opened 2024-09-20
Optimize event detail page

antidipyramid:
## Overview

This PR reduces redundant database calls on the event detail page. Locally, I see a 30% faster load time but I suspect it might be different on staging.

Connects #1138 

## Testing Instructions

* Verify that the event detail page functions as expected, especially the agenda item list
* Verify that agenda item links (View/Download) are correct

### DIFF
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -542,31 +542,6 @@ class LAMetroEventManager(EventManager):
             Prefetch("media", queryset=mediaqueryset)
         ).prefetch_related("media__links")
 
-    def with_related_bills(self):
-        latest_action = BillAction.objects.filter(bill=OuterRef("pk")).order_by(
-            "-order"
-        )
-
-        related_bills = (
-            LAMetroBill.objects.filter(eventrelatedentity__agenda_item__event=event)
-            .defer("extras")
-            .prefetch_related(
-                Prefetch(
-                    "versions",
-                    queryset=BillVersion.objects.filter(
-                        note="Board Report"
-                    ).prefetch_related("links"),
-                    to_attr="br",
-                ),
-                "packet",
-            )
-            .annotate(
-                last_action_description=Subquery(
-                    latest_action.values("description")[:1]
-                )
-            )
-        )
-
 
 class LiveMediaMixin(object):
     BASE_MEDIA_URL = "http://metro.granicus.com/mediaplayer.php?"
--- a/lametro/models.py
+++ b/lametro/models.py
@@ -15,7 +15,7 @@ from django.utils import timezone
 from django.utils.functional import cached_property
 from django.core.cache import cache
 
-from django.db.models import Prefetch, Case, When, Value, Q, F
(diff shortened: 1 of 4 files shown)


lametro/templatetags/lametro_extras.py
```diff
@@ -327,3 +330,10 @@ def get_events_with_manual_broadcasts():
     broadcasts = EventBroadcast.objects.filter(is_manually_live=True)
     events = [b.event for b in broadcasts]
     return events
+
+
+@register.filter
```
antidipyramid, line 335: To avoid repeated database calls for the latest action, I annotated the event's board reports with its latest action's description. Then, I use this template filter to parse the bill status.

lametro/views.py
```diff
@@ -209,10 +221,37 @@ def get_context_data(self, **kwargs):
...
+                Prefetch(
+                    "versions",
+                    queryset=BillVersion.objects.filter(
+                        note="Board Report"
+                    ).prefetch_related("links"),
+                    to_attr="br",
+                ),
+                "packet",
+            )
+            .annotate(
```
antidipyramid, line 241: This annotation is meant to bypass the many queries in `LAMetroBill.inferred_status` to get the last action for each bill. 

@fgregg mentioned that it might be a good idea to bypass `LAMetroBill.inferred_status` in all other views/templates by adding this annotation to the standard `LAMetroBill` manager.

hancush: What are your thoughts on replacing it across the board?

hancush: I think it might be nice to have it accessible when requested, so it's not baked into `objects`, but you can get at it easily, like `LAMetroEvent.objects.with_media`: https://github.com/Metro-Records/la-metro-councilmatic/blob/a63e3359080b818de6df932321ca406f9086a4a9/lametro/models.py#L511

hancush (approved):
This is some fancy footwork, and I'd be interested to see how it performs on staging. Left a question inline.

lametro/views.py
antidipyramid: Yeah, I like adding it as an optional method on the `LAMetroBill` manager.

We use the inferred status property on some board report detail pages so we should probably keep it for that. It's also used in search faceting but optimizing that might require some more thought.

lametro/models.py
```diff
@@ -148,6 +149,17 @@ def get_queryset(self):
 
         return qs
 
+    def with_latest_actions(self):
```
hancush, line 152: 🔥

antidipyramid:
@hancush Before merging, I'm noticing that the download link for the first agenda item for the [7/25 board meeting](https://la-metro-cou-feature-11-8oyjzc.herokuapp.com/event/regular-board-meeting-d6e87548255a/) goes to the current event page rather than Legistar. Do you know why that might be?

Screenshot 2024-09-30 at 10 33 43 AM

antidipyramid:
Pushed up a fix for the agenda item download links. Do you mind taking one last look, @hancush?

PR #1174, opened 2024-11-22
Migrate alert management to Wagtail

hancush:
## Overview

See title. Connects #1175.

### Demo

Screenshot 2024-11-22 at 9 58 46 AM
Screenshot 2024-12-04 at 1 34 07 PM
Screenshot 2024-12-04 at 1 34 14 PM
Screenshot 2024-12-04 at 1 34 59 PM

## Testing Instructions

_Credentials for the datamade user are in Bitwarden under `LA Metro Wagtail User - Staging`_

- Using the review app, add some alerts with various rich text elements. Confirm they appear on the site as expected.
- Also confirm that there's a link to manage alerts in the Wagtail userbar in the bottom left.


lametro/models/cms.py
```diff
@@ -18,3 +24,38 @@ class AboutPage(Page):
...
+
+    panels = [
+        FieldPanel(
+            "type",
+            help_text="Select a style for your alert.",
+        ),
+        FieldPanel("description", widget=MarkdownTextarea),
+    ]
+
+    def content(self):
```
hancush, line 55: The ModelAdmin can render custom model methods in the list display. This method renders the alert description and adds a styled label for the alert type.


Screenshot 2024-11-22 at 10 04 39 AM

lametro/models/legislative.py
```diff
@@ -1237,17 +1237,3 @@ def __str__(self):
 
         else:
             return self.name
-
-
-class Alert(models.Model):
```
hancush, line 1242: Migrated verbatim to `models/cms.py`.

lametro/static/css/wagtail_custom.css
```diff
@@ -0,0 +1,53 @@
+[class*="alert"] {
```
hancush, line 1: Low key obsessed with CSS variables.

lametro/static/js/wagtail_custom.js
```diff
@@ -0,0 +1,37 @@
...
+	})
+};
+
+// Create an observer instance linked to the callback function
+const observer = new MutationObserver(styleAlertTypeSelect);
+
+const config = {childList: true, subtree: true};
+
+// Start observing the target node for configured mutations
+observer.observe(document.body, config);
```
hancush, line 37: This whole rigamarole is necessary because Wagtail adds form fields to the DOM in JavaScript. In order to update the alert type select, we have to wait for it to show up, give it an initial style, then add an event listener for changes. The `MutationObserver` API is pretty dope: https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver

hancush: h/t to @xmedr for this UI!

lametro/templates/base.html
```diff
@@ -115,6 +97,8 @@
             {% endblock %}
         </main>
 
+        {% wagtailuserbar 'bottom-left' %}
```
hancush, line 100: Replaces conditional admin link in top nav.

lametro/wagtail_hooks.py
```diff
@@ -0,0 +1,63 @@
...
+        self.modeladmin = modeladmin_cls()
+
+    def render(self, request):
+        return (
+            f'<li class="w-userbar__item" role="presentation"><a href="{self.modeladmin.url_helper.index_url}" '
+            + f'target="_parent" role="menuitem">Manage {self.modeladmin.get_menu_label()}</a></li>'
+        )
+
+
+@hooks.register("construct_wagtail_userbar")
```
hancush, line 40: https://docs.wagtail.org/en/v5.2.7/reference/hooks.html#construct-wagtail-userbar

lametro/wagtail_hooks.py
```diff
@@ -0,0 +1,63 @@
...
+        '<script src="{}"></script>', static("js/markdowntextarea_overrides.js")
+    )
+
+
+@hooks.register("insert_global_admin_js", order=500)
+def insert_custom_wagtail_javascript():
+    return format_html('<script src="{}"></script>', static("js/wagtail_custom.js"))
+
+
+@hooks.register("insert_global_admin_css", order=500)
```
hancush, line 59: https://docs.wagtail.org/en/v5.2.7/reference/hooks.html#insert-global-admin-css

lametro/wagtail_hooks.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+@hooks.register("construct_wagtail_userbar")
+def add_modeladmin_links(request, items):
+    for modeladmin_cls in (AlertAdmin,):
+        items.append(ModelAdminLink(modeladmin_cls))
+    return items
+
+
+@hooks.register("insert_global_admin_js", order=100)
```
hancush, line 47: https://docs.wagtail.org/en/v5.2.7/reference/hooks.html#insert-global-admin-js

.pre-commit-config.yaml
```diff
@@ -15,5 +15,5 @@ repos:
     rev: '9f60881'
     hooks:
       - id: flake8
-        files: lametro councilmatic tests
+        files: ^(lametro|councilmatic|tests)
```
hancush, line 18: TRICKY LITTLE BUGGER

lametro/static/css/wagtail_custom.css
xmedr: Did some reading up on this and oh yea I can totally see why 👀

lametro/templates/base.html
xmedr: Since they'll have to sign in to wagtail to manage the alerts after this, would it be useful to change up the Metro Login link in the footer? Maybe by making it link to the wagtail login instead. That is, as long as logging in through wagtail would still allow them to manage things like manual broadcast links or events that don't appear in legistar, which I imagine it would.

If we think yes, then I can take care of it when I branch off of this!

(Also something to consider in the future is whether wagtail should also somehow manage manual broadcasts.)

lametro/wagtail_hooks.py
```diff
@@ -0,0 +1,56 @@
...
+            f'<li class="w-userbar__item" role="presentation"><a href="{self.modeladmin.url_helper.index_url}" '
+            + f'target="_parent" role="menuitem">Manage {self.modeladmin.get_menu_label()}</a></li>'
+        )
+
+
+@hooks.register("construct_wagtail_userbar")
+def add_modeladmin_links(request, items):
+    for modeladmin_cls in (AlertAdmin,):
+        items.append(ModelAdminLink(modeladmin_cls))
+    return items
```
xmedr, line 44: I'm curious why this is a loop, could you talk a bit about that?

lametro/static/js/wagtail_custom.js
xmedr: Ah so is the mutation observer being used here mostly because we can't attach an event listener the usual way since the selector doesn't exist yet?

xmedr (changes requested):
This works great! Do you think it'd be worth to add a border to the type select? I had trouble noticing it at first with a light theme and nothing selected (pic below). Otherwise, I've just got a couple questions inline.

image

lametro/wagtail_hooks.py
hancush: Good catch! Factored out.

lametro/templates/base.html
hancush: Oh, good idea. Yes!

lametro/static/css/wagtail_custom.css
```diff
@@ -1,16 +1,21 @@
...
 
 .button[class*="alert"]:hover {
   background-color: rgba(var(--r), var(--g), var(--b), var(--a));
   color: var(--font-color);
 }
 
+/* Add default border to empty alert select */
+.alert- {
+  border: 1px solid var(--w-color-border-field-default)!important;
+}
```
xmedr, line 17: Nice, tyty!

xmedr (approved):
Awesome, looks good!

PR #1180, opened 2024-12-03
Fix duplicate search filters

hancush:
## Overview

The accessibility changes did not address an issue where search filters were duplicated. This PR applies a patch with the updated markup.

Connects #1009, #979 

#### Before (Staging)

Screenshot 2024-12-03 at 11 27 18 AM

#### After

Screenshot 2024-12-03 at 11 26 40 AM

## Testing Instructions

- On the review app, view the search page and confirm all of the facets display correctly.
- Double check that the markup retains the accessibility changes (namely, data attributes and the aria-* attributes).

### DIFF
--- a/lametro/templates/search/_search_filter.html
+++ b/lametro/templates/search/_search_filter.html
@@ -1,7 +1,7 @@
 {% load extras %}
 {% load lametro_extras %}
 
-{% if facet_label != 'Controlling Body' %}
+{% if facet_label != 'Controlling Body' and item_list|length > 0 %}
 
     <div class="accordion-item">
         <h3 class="accordion-header">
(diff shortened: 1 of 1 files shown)


lametro/templates/search/_search_filter.html
```diff
@@ -21,40 +21,31 @@ <h3 class="accordion-header">
...
-                                </li>
-                            {% endif%}
-
-                        {% endfor %}
-                    {% else %}
-
-                        {% for name, count in item_list %}
-                            {% if count %}
-                                {% include 'search/_ordered_search_filter_items.html' %}
+                                    </a>
```
xmedr, line 45: Voiceover is reading each facet link in the Meetings filter twice, and it looks like this produces nested links (which then resolves to just two links next to each other).

Removing the outermost anchor tag might be what we're looking for here. Maybe something like:

```html
<div class="skinny-list">
  {% if name in selected_facets.sponsorships %}
      <strong>{{ name | title }}</strong>
      <a href="#" class="remove-filter-value btn btn-primary btn-sm" data-param="{{facet_name}}_exact:{{name}}" aria-label="Remove {{name}} filter">
          <i class="fa fa-times" aria-hidden="true"></i>
      </a>
  {% else %}
      <a href="#" class="filter-value" data-param="sponsorships_exact:{{name}}" title="{{ name|title }}">
          {{ name | title }}
      </a>
  {% endif %}
</div>
```

xmedr (changes requested):
Ah ty for catching this! Looks good and the a11y changes are all intact. Just got one change in there

hancush:
Great catch, thank you @xmedr! Patch in and ready for your eyeballs.

xmedr (approved):
This is perfect!

Issue #1177, opened 2024-12-02
Create pattern for editing content that appears across a set of detail pages


hancush:
E.g., public comment instructions on event detail pages

N.b., the CMS will not enable customizing information about particular events / board reports - that information should be entered into Legistar.

hancush:
Connects #1168

xmedr:
Let's try to see if we can include custom "blocks" in the rich text editor based on the event itself, like the ecomment url. This link might help: https://docs.wagtail.org/en/latest/advanced_topics/customization/page_editing_interface.html#rich-text-html

hancush:
In our jam session, @xmedr and I contemplated a model to contain editable text that varies based on when it is and when the event takes place. 

```python
# model

class EventBlurb
	CONDITION_CHOICES = future, upcoming, ongoing, concluded
	
	condition = arrayfield(charfield(choices))  # this is always OR (AND is not possible)
	blurb = textfield

# template

for blurb in event blurb
	for condition in condition
	if condition
		blurb
```

hancush:
@xmedr This might be helpful! https://docs.wagtail.org/en/v6.3.1/extending/rich_text_internals.html

xmedr:
@hancush Thanks for that link! Through that I've found that I should be [extending the draftail editor to include a new block](https://docs.wagtail.org/en/v5.2.7/extending/extending_draftail.html#creating-new-blocks). And this almost works. I can create a new link type and give it an aria-label, target, and href. And I was hoping that I could default the href to `{{event.ecomment_url}}`, save that to the model, and have that render, however this is not the case. The href itself shows up as that template variable.

Do you have any ideas on making django render template variables from the database? If not, the alternative could be to have some JS on event detail pages that attaches to these ecomment links if they exist, and either changes the href, or just takes care of opening the right link itself.

The rendered link:
![image](https://github.com/user-attachments/assets/9f58e78f-77f7-4844-9c2a-99e76cda7f08)

The editor with a new ecomment button:
![image](https://github.com/user-attachments/assets/95684256-e386-4887-be48-c0189b72a88d)

hancush:
@xmedr Hm, let's pair on this one. Want to grab a time?

xmedr:
After #1185 comes in @xmedr to migrate the usage of the wagtail `modeladmin` module to instead use `snippets`

hancush:
Done in #1168!

PR #1185, opened 2024-12-05
Allow for notices on all event pages

xmedr:
## Overview

This branch replaces the hard coded public comment instructions with wagtail-editable objects from an `EventNotice` model. These objects will render according to conditions that Metro can select for each notice, and those condition options came from the conditions that were present within the `event_header_*_public_comment` templates. There are two sets of conditions: the status of the broadcast, and its public comment setting.

I've also updated the fixtures with a set of event notices that match what metro currently has on the site.

- Connects #1177

## Testing Instructions
** Note: This PR requires a database migration, the review app will not work and this will have to be tested locally.
 * Pull down this branch and run migrations
   * You may instead run an event scrape if needed, and this would run migrations
 * To have a set of EventNotices ready, load in the cms content with the `load_content` management command
 * Create a super user if needed and log in to the cms `localhost:8001/cms`
   * hannah's from the staging site is in the fixture
 * Head to Event Notices and confirm they have loaded correctly
   * Also confirm that the About Us page content is still intact
 * Make some notices for events that exist for you.
   * Confirm that they render as expected on event detail pages, conditions and all


lametro/templates/event/_event_header_live_public_comment.html
```diff
@@ -1,5 +1,5 @@
 <section class="row">
-    {% if has_agenda and not event.has_passed %}
+    {% if has_agenda and event_notices %}
```
xmedr, line 2: This branch is allowing for a notice to be posted even when an event has_passed. Because of this, the styling conditionals have been adjusted to allow space for notices if they exist at all.

lametro/templates/event/_event_header_live_public_comment.html
```diff
@@ -53,78 +53,41 @@
...
         <article class="col-md-5">
-            <h2 class="h4">Submit public comment remotely</h2>
+            {% if not event.has_passed %}
+                {% for notice in event_notices %}
+                    {% if "upcoming" in notice.broadcast_conditions and event.is_upcoming %}
+                        {{ notice.message|safe }}
+                    {% elif "ongoing" in notice.broadcast_conditions and event.is_ongoing %}
+                        {{ notice.message|safe }}
```
xmedr, line 63: This pattern makes it so that notices only ever show if the event's broadcast status matches the notifications' conditions. This is repeated across the `public_comment` partials

lametro/templates/event/_event_header_live_public_comment.html
xmedr: Also something to note, any notice showing up is still contingent on the event having an agenda, to mimic the way public comment instructions only displayed if they had one as well. Lmk if we want to change this!

lametro/models/cms.py
```diff
@@ -57,3 +59,75 @@ def content(self):
             + f"<span class='button button-small alert-{self.type}'>{self.get_type_display()}</span>"
             + "</div>"
         )
+
+
+class CheckboxSelectMultipleList(forms.CheckboxSelectMultiple):
+    def format_value(self, value):
+        if isinstance(value, str):
+            return value.split(",")
```
xmedr, line 67: This is helping to actually display our arrayfield choices in the create/edit views

lametro/templates/event/_event_header_live_public_comment.html
```diff
@@ -53,78 +53,41 @@
...
+
+                        {% if USING_ECOMMENT %}
+                            <p>
+                                <strong>On the web:</strong>
+                                <a href="{{ event.ecomment_url }}" target="_blank" aria-label="Go to public comment - link opens in a new tab">
+                                    <i class="fa fa-fw fa-external-link" aria-hidden="true"></i>
+                                    Go to public comment
+                                </a>
+                            </p>
+                        {% endif %}
```
xmedr, line 73: The ecomment portions of the partials have been kept for now.

lametro/views.py
```diff
@@ -283,6 +284,20 @@ def get_context_data(self, **kwargs):
 
         context["USING_ECOMMENT"] = settings.USING_ECOMMENT
 
+        # Only provide notices for this event's public comment setting
+        if event.accepts_live_comment:
+            context["event_notices"] = EventNotice.objects.filter(
+                comment_conditions__contains=["accepts_live_comment"]
+            )
```
xmedr, line 291: Deciding which notices to pass through based off of public comment status in the view to take some burden off of the template.

lametro/models/cms.py
```diff
@@ -57,3 +59,75 @@ def content(self):
...
+        ("future", "Future events"),
+        ("upcoming", "Upcoming events"),
+        ("ongoing", "Ongoing events"),
+        ("concluded", "Concluded events"),
+    ]
+    COMMENT_CONDITION_CHOICES = [
+        ("accepts_live_comment", "Events that accept live public comments"),
+        ("accepts_comment", "Events that accept public comments when not live"),
+        ("accepts_no_comment", "Events that do not accept public comments at all"),
+    ]
```
hancush, line 91: Hm, I'm a little confused by the comment conditions. I think all events accept comment?

lametro/templates/event/_event_header_live_public_comment.html
```diff
@@ -53,78 +53,41 @@
         {% endif %}
     </section>
 
-    {% if has_agenda and not event.has_passed %}
+    {% if has_agenda and event_notices %}
```
hancush, line 56: I'm wondering now if it would be easier to get the appropriate notice/s in the view so we can simplify the templates?

lametro/templates/base.html
```diff
@@ -109,7 +109,7 @@
             <ul class="nav col-md-6 justify-content-around justify-content-md-end list-unstyled">
                 <li class="ms-3"><a href="{% url 'lametro:contact' %}">Contact Us</a></li>
                 <li class="ms-3"><a href="{% slugurl 'about' %}">About Us</a></li>
-                <li class="ms-3"><a href="{% url 'metro_login' %}">Metro Login</a></li>
+                <li class="ms-3"><a href="{% url 'wagtailadmin_home' %}">Metro Login</a></li>
```
hancush, line 112: ❤️

hancush (changes requested):
This is looking really snazzy! Left a few questions inline. How about also adding a link to edit the event notice to the Wagtail userbar? (See the hook that adds the manage alerts link.)

hancush:
P.s., I pushed an updated fixture omitting the OCD models – it should only contain content managed by the CMS!

lametro/models/cms.py
xmedr: Ah okay, I was basing this off of the below section in the event template. Is this is a holdover from a time when not all events accepted comment, or maybe I'm just misinterpreting the below conditions?

https://github.com/Metro-Records/la-metro-councilmatic/blob/431a943513bca38620de620cd9c19ad5067c3280/lametro/templates/event/event.html#L109-L116

hancush: Looks like some committee meetings don't accept comments: https://github.com/Metro-Records/la-metro-councilmatic/pull/823/files

But that's determined on the back end – this logic shouldn't be exposed in the CMS!

xmedr: Gotcha, I've removed it from the cms!

In that case, what do you think about me bringing back the public comment instructions in the template then? Since it says different things depending on which template is being rendered, and the same "upcoming" notice would render across all those templates regardless of whether the event accepted comment or not.

lametro/templates/event/_event_header_live_public_comment.html
xmedr: I think it would be easier once we figure out how to get those template variables to work within these notices. As of now, the conditionals that are left within these three comment templates have event specific variables in them that we wouldn't currently be able to render in the editable notice. Unless you had something else in mind!

lametro/models/cms.py
hancush: Hmmm.................................. maybe you were right to include the comment conditions after all. Can we change the labels to:

- Events that accept public comment before and during the meeting
- Events that accept public comment before but not during the meeting
- Events that do not accept public comment

hancush: Sorry for the run around!

lametro/templates/event/_event_header_live_public_comment.html
hancush: Got it, ok. Let's put a pin in this.

lametro/models/cms.py
xmedr: No worries! It's an easy adjustment and I could've explained myself more haha.

xmedr:
@hancush public comment changes reverted and notices added to userbar!
image

hancush:
I want to look at this with fresh eyes in January, but for now, it looks great, @xmedr!

lametro/templates/event/_event_header_live_public_comment.html
xmedr: Awesome, that made sense and I have those changes in. I did keep a little bit of the conditionals in this file in particular concerning the actual ecomment link. But this template and the other two now have their notices filtered in the view by event status/condition

lametro/views.py
```diff
@@ -283,8 +284,44 @@ def get_context_data(self, **kwargs):
...
+        # Returns the event's broadcast status as a string
+
+        if event.is_upcoming:
+            return "upcoming"
+        elif event.is_ongoing:
+            return "ongoing"
+        elif event.has_passed:
+            return "concluded"
+        else:
+            return "future"
```
xmedr, line 323: I have this method in the view returning the matching string per event status.

The LAMetroEvent model has something similar in the `display_status` property, but the statuses that it returns are just different enough that I decided against using it. It returns a `cancelled` status which we're not using and lumps upcoming and future events into the `upcoming` status. I could change the latter bit to fit our needs if we want to! Provided that it's okay to have a new Future label on events

lametro/templates/event/_event_header_live_public_comment.html
```diff
@@ -53,78 +55,24 @@
...
-                    <i class="fa fa-fw fa-exclamation-circle" aria-hidden="true"></i>
-                    <em>
-                        When you call in to comment, you will be able to hear the live stream of the meeting while you wait to comment.
-                        <strong>If you are also listening to the meeting on another device, please lower the volume on the second device to avoid feedback and ensure you can be heard clearly.</strong>
-                    </em>
-                </p>
-
-                {% if event.is_ongoing and USING_ECOMMENT %}
+            {% if USING_ECOMMENT %}
+                {% if event.is_ongoing %}
```
hancush, line 65: Can we use the `event_status` here, too?

hancush (commented):
One small thing inline, then this is good to go! I'll use the wagtail-epic branch for the new staging instance. Can you change the base for this PR before you merge?

xmedr:
Change pushed and base branch updated! I'll merge it in now then

Issue #1176, opened 2024-12-02
Create pattern for editing content on pages with some data elements


hancush:
E.g., Board Members, Committees, and their detail pages.

As part of this issue, migrate management of board member bios and headshots to Wagtail.

antidipyramid:
After #1189 is in, we'll want to think about extending this pattern to committee detail views.

antidipyramid:
Bios and headshots done. We'll want to ask Metro whether they want editable committee detail elements as well.

PR #1189, opened 2024-12-20
Implement board member details as snippet

hancush:
## Overview

This PR adds a Django model for board member headshots and bios and registers as a snippet for editing in the CMS. It also deprecates the `headshot_url` property, instead creating a shared template block that renders the right thing given a person instance and, optionally, a way to request the original aspect ratio, instead of a square thumbnail.

Connects #1176

### Demo

Screenshot 2024-12-20 at 12 15 58 PM
Screenshot 2024-12-20 at 12 16 03 PM
Screenshot 2024-12-20 at 12 16 15 PM
Screenshot 2024-12-20 at 12 16 26 PM

### Notes

I decided to go ahead and implement this as a snippet, since the `modeladmin` module is deprecated in the next version of Wagtail, so using snippets now makes our migration path easier moving forward. I'm wondering if we might want to do this in #1185, too. https://docs.wagtail.org/en/v5.2.1/reference/contrib/modeladmin/migrating_to_snippets.html

## Testing Instructions

- Log into the review app, then check out the Board Member Details section of the CMS.
- Try some filters and verify they work as expected.
- Try editing some headshots and bios. Confirm the previews render correctly.
- Publish some of your edits. Confirm the changes are visible on the live site.


lametro/signals/handlers.py
```diff
@@ -0,0 +1,12 @@
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from lametro.models import LAMetroPerson, BoardMemberDetails
+
+
+@receiver(post_save, sender=LAMetroPerson)
```
hancush, line 7: Will handle instantiating details for new members moving forward. To backfill, we simply need to call save on all existing board members.

lametro/models/cms.py
```diff
@@ -24,6 +26,64 @@ class AboutPage(Page):
...
+        max_length=256, blank=True, null=True, default="Metro"
+    )
+    bio = RichTextField(blank=True, null=True)
+    _revisions = GenericRelation(
+        "wagtailcore.Revision", related_query_name="member_details"
+    )
+
+    @property
+    def revisions(self):
+        return self._revisions
```
hancush, line 57: https://docs.wagtail.org/en/v5.2.7/topics/snippets/features.html#saving-draft-changes-of-snippets

lametro/models/cms.py
```diff
@@ -24,6 +26,64 @@ class AboutPage(Page):
...
+
+    def get_preview_context(self, request, mode_name):
+        context = super().get_preview_context(request, mode_name)
+        context["person"] = self.person
+        context["person_details"] = self
+
+        council_post = self.person.latest_council_membership.post
+        context["qualifying_post"] = council_post.acting_label
+
+        context["preview"] = True
```
hancush, line 79: The map and committee/board report modules require a bunch of extra context, so I use this to purposefully omit them from page previews. (I also added a note to the admin interface.)

lametro/templates/snippets/related_object_status_tag.html
```diff
@@ -0,0 +1,4 @@
+{% load wagtailadmin_tags %}
```
hancush, line 1: More or less stolen from the Wagtail source code, to make the status icon in the board member details snippet list link to the person detail page.

lametro/wagtail_hooks.py
```diff
@@ -23,22 +38,189 @@ class AlertAdmin(ModelAdmin):
...
+    Model permission that doesn't allow creating new instances.
+    """
+
+    def user_has_permission(self, user, action):
+        if action in ("add", "delete"):
+            return False
+        return user.has_perm(self._get_permission_name(action))
+
+
+class BoardMemberFilterSet(django_filters.FilterSet):
```
hancush, line 58: This is a `django-filter` filter set, which I'm using to add custom filters for whether a person is a current board member and what body they belong to, on the board member detail snippet listing page.

lametro/wagtail_hooks.py
```diff
@@ -23,22 +38,189 @@ class AlertAdmin(ModelAdmin):
...
+            PublishingPanel(),
+        ]
+    )
+
+
 modeladmin_register(AlertAdmin)
+register_snippet(BoardMemberDetailsViewSet)
+
+
+class UserBarLink:
```
hancush, line 155: I pulled some reusable things out of the model admin link so I could also use them for the snippet edit link.

lametro/wagtail_hooks.py
```diff
@@ -23,22 +38,189 @@ class AlertAdmin(ModelAdmin):
...
+        except KeyError:
+            return
+
+        try:
+            snippet = BoardMemberDetails.objects.get(person__slug=slug)
+        except BoardMemberDetails.DoesNotExist:
+            return
+
+        finder = AdminURLFinder()
+        return finder.get_edit_url(snippet)
```
hancush, line 215: TIL! https://docs.wagtail.org/en/latest/extending/admin_views.html#adminurlfinder

lametro/models/cms.py
xmedr: The map currently isn't showing up on the non-preview version of the person detail page. This might be because it's on a review app that doesn't have access to `map_geojson` but wanted to make sure!

xmedr (changes requested):
This looks really cool so far! Got a few a questions and comments below, but the status links and "edit this page" links work exactly as expected, and the whole editing process is smooth.

lametro/models/cms.py
hancush: Great catch! The `populate_shapes` command (and, indeed, the whole release script) wasn't running, since this is a review app. I've added a new config variable, `WAGTAIL_STAGING`, a default to True for review apps in our CMS workstream, and a value of True for the actual Wagtail staging instance, so the release script will run there, too. The effect of all of this _should_ be that review apps created from branches where `WAGTAIL_STAGING` is set in `app.json` should run the release script, and so should the Wagtail staging deployment. Does that make sense?

lametro/models/cms.py
```diff
@@ -24,6 +26,64 @@ class AboutPage(Page):
...
+        return reverse("lametro:person", kwargs={"slug": self.person.slug})
+
+    def __str__(self):
+        loaded_obj = (
+            type(self)
+            .objects.select_related("person")
+            .prefetch_related("person__memberships")
+            .get(id=self.id)
+        )
+        return f"{loaded_obj.person.name}{' (current)' if loaded_obj.person.current_memberships.exists() else ''}"
```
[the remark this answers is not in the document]
hancush: Excellent question. It _is_ useful when we know we'll be working with related objects! Any time you see yourself writing `{model}.{related_object}.{attribute}`, you should check the initial queryset to see whether you've loaded in that related object, in order to avoid the extra database queries. I did a lunch and learn on this way back in the day, when I was learning this lesson in a very painful way, lol: https://docs.google.com/document/d/1giTexSmZ0hXd1sPBU8AUuT_O2VWgpXaGqDZnMzW05kE/edit?usp=sharing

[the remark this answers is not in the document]
hancush: It's extra useful when you're working in a loop, or if you're accessing multiple attributes from a related object.

lametro/signals/handlers.py
hancush: Good idea! Signals aren't fired in migrations, so I added a `RunPython` step to the existing migration that creates detail pages for any members we already have in our database.

docker-compose.yml
```diff
@@ -64,7 +64,6 @@ services:
         condition: service_healthy
     environment:
       DATABASE_URL: 'postgis://postgres:postgres@postgres/lametro'
-      SHARED_DB: "True"
```
hancush, line 67: Byeeee https://github.com/Metro-Records/scrapers-lametro/pull/35

lametro/migrations/0018_boardmemberdetails.py
```diff
@@ -0,0 +1,140 @@
+# Generated by Django 3.2.25 on 2024-12-20 15:25
+
+from django.db import migrations, models
+import django.db.models.deletion
+import wagtail.fields
+import wagtail.models
+
+
+def initialize_detail_pages(apps, schema_editor):
```
hancush, line 9: Great suggestion, @xmedr!

xmedr (approved):
I got a couple error pages about relations not existing on the Pages and Board Member Details tabs when I first started up since a set of migrations did not automatically apply themselves. But after running `migrate` manually, everything worked as expected! It's probably because the `WAGTAIL_STAGING` env var wasn't set on this new review app.

This can come in, but what do you think about adding a note about that var in the readme of this or an upcoming pr, and then also adding that var in to subsequent review app configs as False?

PR #1202, opened 2025-01-28
Migrate usage of wagtail modeladmin to snippets

### LINKED ISSUE #1177: Create pattern for editing content that appears across a set of detail pages

xmedr:
## Overview

This migrates usage of wagtail's modeladmin module throughout the app, to the newer snippets module. This also refines the filters used in the event notice listing page.

- Closes #1177

### Demo

image

### Notes

~~This is comparing to (and branched from) the board-member-details cms branch since that had a start to this migration. I can change the comparison to the cms/epic branch after the board member one gets merged.~~ Base branch changed!

## Testing Instructions
 * Log in to the cms and make an event notice
 * Confirm that the notice gets created as expected
 * Confirm that filtering notices on the list view works as expected

### DIFF
--- a/lametro/wagtail_hooks.py
+++ b/lametro/wagtail_hooks.py
@@ -1,8 +1,7 @@
 from django.contrib import admin  # noqa
 from django.templatetags.static import static
-from django.utils.html import format_html, strip_tags
+from django.utils.html import format_html
 
-from html import unescape
 import django_filters
 from wagtail import hooks
 from wagtail.admin.admin_url_finder import AdminURLFinder
@@ -14,7 +13,6 @@ from wagtail.admin.panels import (
     HelpPanel,
 )
 from wagtail.admin.ui.tables import UpdatedAtColumn, LiveStatusTagColumn
-from wagtail.contrib.modeladmin.options import ModelAdmin, modeladmin_register
 from wagtail.permissions import ModelPermissionPolicy
 from wagtail.snippets.models import register_snippet
 from wagtail.snippets.views.snippets import SnippetViewSet
@@ -22,9 +20,9 @@ from wagtail.snippets.views.snippets import SnippetViewSet
 from lametro.models import Alert, BoardMemberDetails, LAMetroOrganization, EventNotice
 
 
-class AlertAdmin(ModelAdmin):
+class AlertViewSet(SnippetViewSet):
     model = Alert
-    base_url_path = "alerts"
+    icon = "warning"
     menu_icon = "warning"
     menu_order = 100
     add_to_settings_menu = False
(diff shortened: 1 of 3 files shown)


lametro/wagtail_hooks.py
```diff
@@ -94,6 +92,19 @@ class Meta:
         fields = ["current_member"]
 
 
+class EventNoticeFilterSet(django_filters.FilterSet):
+    broadcast_conditions = django_filters.ChoiceFilter(
+        lookup_expr="icontains", choices=EventNotice.BROADCAST_CONDITION_CHOICES
+    )
+    comment_conditions = django_filters.ChoiceFilter(
+        lookup_expr="icontains", choices=EventNotice.COMMENT_CONDITION_CHOICES
+    )
```
xmedr, line 101: The filters on the listing page are using the easier to read (but often lengthy) version of the condition choices, while the list itself is using the shorter, less clear version.

It's probably best to use just one of those across both, but I wanted to check which you thought was the better format. I'm leaning towards to more clear one.

lametro/wagtail_hooks.py
```diff
@@ -215,17 +214,17 @@ def render(self, request, href=None):
...
     icon_name = "plus"
 
-    def __init__(self, modeladmin_cls):
-        self.modeladmin = modeladmin_cls()
+    def __init__(self, viewset_cls):
+        self.viewset = viewset_cls()
 
     def get_href(self, request):
-        return self.modeladmin.url_helper.index_url
+        return f"/cms/{self.viewset.url_prefix}/"
```
xmedr, line 224: I tried to programmatically get the `/cms/` portion of the string by using methods on the viewset and using reverse on the viewset's namespace, but had zero luck. Thoughts?

lametro/wagtail_hooks.py
antidipyramid: Yeah, I think going with the longer version is worth it for the clarity.

lametro/wagtail_hooks.py
antidipyramid: What were you passing to `reverse`?

antidipyramid (commented):
The listings look and work great.

lametro/wagtail_hooks.py
xmedr: It was things like
```python
self.viewset.url_prefix
# returns "snippets/lametro/alert"

self.viewset.name
# returns "alert"

self.viewset.get_url_name(self.viewset.name)
# returns "wagtailsnippets_lametro_alert:alert"

self.viewset.chooser_admin_url_namespace
#returns None
```

I also tried things like
```python
reverse_lazy('wagtailsnippets', args=["lametro", self.viewset.url_prefix])
# this returned "django.urls.exceptions.NoReverseMatch: Reverse for 'wagtailsnippets' not found. 'wagtailsnippets' is not a valid view function or pattern name."
```

antidipyramid: I'm not having any luck either.  Explicitly adding`/cms/` is an OK compromise!

PR #1211, opened 2025-02-13
Fix PDF Viewer 

antidipyramid:
## Overview

This PR fixes a PDF viewer error by moving core councilmatic URLs above the new wagtail ones. 

Connects #1211

## Testing Instructions
 * Verify that PDF viewers load as expected


hancush (changes requested):
Hm, the empty path ("") will swallow all URLs. So now, our Wagtail pages aren't resolving. (The About page is the default Councilmatic page, not the Wagtail-managed page.) I think we need to figure out a smarter way to route so that, if there is a Wagtail page at a given path, then we prefer it, but otherwise dump into Councilmatic (or something like that). What do you think?

hancush:
@antidipyramid We're already overriding most of the Councilmatic URLs, so what do you think about including an explicit entry for all the Councilmatic URLs we use (compare our urls.py to [the root one](https://github.com/datamade/django-councilmatic/blob/3.x/councilmatic_core/urls.py)) and letting Wagtail be the catch all?

antidipyramid:
@hancush Sounds like a good plan (with a side effect of no longer having to consult `django-councilmatic`'s `urls.py`!).

antidipyramid:
@hancush In order to pull in all of the councilmatic patterns to our `urls.py`, it seems like we'd also have to reach into the template files and update namespaces like `{% url 'events' %}` to `{% url 'lametro:event' %}`.

Also, there are `reverse` calls scattered throughout inherited event/person/bill from `django-councilmatic` that we'd have to account for. 

Since this change would require some careful combing through of various files, what do you think about patching this particular bug by explicitly declaring the `pdfviewer` pattern in `lametro/urls.py` for now? After this fix is in, we can pull in the rest of the councilmatic urls.

```py
    path("smartlogic/", include("smartlogic.urls", namespace="smartlogic")),
    url(r"^pdfviewer/$", pdfviewer, name="pdfviewer"),
    path("cms/", include(wagtailadmin_urls)),
    path("documents/", include(wagtaildocs_urls)),
    path("", include(wagtail_urls)),
    url(r"", include("councilmatic_core.urls")),
```

hancush:
@antidipyramid That sounds like a good plan, provided the `url` and `reverse` calls are not affected by the catch all snake pit?

hancush (approved):

Issue #1221, opened 2025-03-11
Nav: Update link to "Archive Search"


shrayshray:
Please update the link to "Archive Search" in the top navigation. The new link is: https://mtasearch02.metro.net:23352/apps/boardarchives/

![Image](https://github.com/user-attachments/assets/aa14babf-d45a-4b3a-bfc7-9d4d7b5a07ee)

PR #1222, opened 2025-03-11
Update archive search link

antidipyramid:
## Overview

See title

Connects #1221


xmedr (approved):
Looks good! Just one more spot to change and this can come in:
https://github.com/Metro-Records/la-metro-councilmatic/blob/742b692941601575eff7e6f2d6cb532acf7171c0/lametro/templates/search/_empty_search_message.html#L1-L3

Also, they may need to change the link in their About Us page. It currently leads to the version of the search that I don't think they're using:
image

antidipyramid:
@xmedr thanks for catching that!

PR #1223, opened 2025-03-12
Feat/expiring alerts

### LINKED ISSUE #1198: Expiring site alerts
From @neilarellano:

> Are we able to add a feature to have a duration for alerts in which they expire/get removed after?

xmedr:
## Overview

tk

- Closes #1198

#### Alerts in the cms:
image

---

#### Alerts on the site:
image

### Notes

Because there are some previously approved changes from main in here, I've commented on all the changes relevant to this pr to make those clear. You can also check the latest commits to see these changes.

## Testing Instructions
** Note: if this PR requires a database migration, the review app may not work and this may have to be tested locally.
 * tk


lametro/models/cms.py
```diff
@@ -122,6 +127,9 @@ def content(self):
             + "</div>"
         )
 
+    class Meta:
+        ordering = ["pk"]
+
```
xmedr, line 132: This way they're always listed in chronological order of when they were added whether when rendered or when listed in the cms.

lametro/templates/base.html
```diff
@@ -83,7 +83,7 @@
 
         {% get_alerts as alerts %}
         {% if alerts %}
-            {% for alert in alerts|dictsort:"pk" %}
+            {% for alert in alerts %}
```
xmedr, line 86: Remove the previous sort now that they're always sorted chronologically.

lametro/templatetags/lametro_extras.py
```diff
@@ -327,7 +332,7 @@ def sort_topics(topics):
 
 @register.simple_tag
 def get_alerts():
-    return Alert.objects.all()
+    return Alert.objects.exclude(expiration__lt=timezone.now())
```
xmedr, line 335: The actual date filtering

lametro/wagtail_hooks.py
```diff
@@ -34,7 +34,7 @@ class AlertViewSet(SnippetViewSet):
     add_to_settings_menu = False
     exclude_from_explorer = False
     add_to_admin_menu = True
-    list_display = ("content",)
+    list_display = ("content", "expiration")
```
xmedr, line 37: Show expiration on the alerts list

tests/conftest.py
```diff
@@ -316,6 +318,36 @@ def build(self, **kwargs):
     return LAMetroSubjectFactory()
 
 
+@pytest.fixture
+@pytest.mark.django_db
+def alert(db):
+    """
+    An alert builder. The expiration can be entered as either a timezone aware
+    datetime object, or a datetime string in the format '%Y-%m-%d %H:%M:%S'.
+    """
```
xmedr, line 327: A shiny new builder!

tests/test_cms_models.py
```diff
@@ -0,0 +1,27 @@
...
+        ("2025-03-20 10:00:00", True),
+        ("2025-03-15 11:00:00", True),
+        ("2025-03-15 09:00:00", False),
+        ("2025-03-14 10:00:00", False),
+    ],
+)
+def test_alert_expiration(client, alert, expiration_str, should_display):
+    """
+    Test that alerts do not show on the page when expired
+    """
```
xmedr, line 20: Testing for the alert behavior

hancush:
@xmedr You can follow our instructions for Wagtail review apps in the README to run the scrape locally with a token against the review app's DB: https://github.com/Metro-Records/la-metro-councilmatic?tab=readme-ov-file#review-apps-for-wagtail-changes

lametro/templates/base.html
hancush: Great call to move the ordering up into the model so that it's consistent anywhere we're displaying the alerts!

lametro/templates/index/_meeting_details_next.html
```diff
@@ -34,16 +34,19 @@ <h3 class="h4">{{  meeting.link_html | safe }}</h3>
...
 
         <!-- Supplementary links -->
-        {% if meeting.documents.all %}
-          <div class="row">
-            <div class="col-7">
-              <a class="btn btn-teal" id="pdf-download-link" target='_blank' href='#' aria-label="Get Agenda PDF - link opens in a new tab">
-                <i class='fa fa-fw fa-download' aria-hidden="true"></i>
-                Get Agenda PDF
-              </a>
+        {% with agenda_url=meeting.documents.all|find_agenda_url %}
```
hancush, line 37: What's up with this change?

lametro/templatetags/lametro_extras.py
hancush: Hm, if dates are stored in the DB in UTC, should we pass a local time or a UTC time for filtering?

hancush (changes requested):
Looking good! 

Love the idea of entering and displaying expiration dates in local time while storing them in UTC time. Is this an out of the box thing when managing times in Wagtail, or did you configure it someplace?

Also, what do you think about defaulting to showing only active alerts, with the option to show all? The board member detail filtering might be instructive here, but happy to chat more about an approach if you like!

lametro/templates/index/_meeting_details_next.html
xmedr: This was part of [a fix in this other pr](https://github.com/Metro-Records/la-metro-councilmatic/pull/1218). I merged in main to keep the cms branch up to date (since learning my lesson with the a11y branches 😱).

But that fix was mostly to get the pdf agenda links working again, since they were previously empty.

lametro/templatetags/lametro_extras.py
xmedr: UTC for sure! And `timezone.now()` operates in UTC in this case.

At time of writing, `timezone.now` returns this in the review app's shell:
`datetime.datetime(2025, 3, 13, 14, 13, 32, 744054, tzinfo=<UTC>)`

And the current time in UTC is 2:13 PM, so this lines up.

xmedr (commented):
@hancush Ah okay so it's not the review app's vars that's the issue. Running the same scrape again this morn worked without an issue so maybe it was just a weird case (though there is a super small fix coming to get the rest of the make command working). The review app is now working!

As far as your questions:
- The app was already configured to do this! Since the `TIME_ZONE` in our settings is for LA, Django uses that to [display datetimes in templates and to interpret datetimes entered in forms](https://docs.djangoproject.com/en/5.1/ref/settings/#std-setting-TIME_ZONE). The backend keeps times in UTC by default to avoid conflicts with DST changes.
- Yea I like that idea, I'll start on that!

lametro/wagtail_hooks.py
```diff
@@ -41,6 +66,11 @@ class AlertViewSet(SnippetViewSet):
         "description",
     )
 
+    def get_queryset(self, request):
+        return Alert.objects.filter(
+            Q(expiration__gt=timezone.now()) | Q(expiration__isnull=True)
+        )
```
xmedr, line 72: Success 🎉

lametro/wagtail_hooks.py
```diff
@@ -26,11 +28,34 @@
...
+        empty_label="Not Expired",
+    )
+
+    def filter_expired_alerts(self, queryset, name, value):
+        if value == "expired":
+            return Alert.objects.filter(expiration__lte=timezone.now())
+        elif value == "all":
+            return Alert.objects.all()
+        else:
+            return queryset
```
xmedr, line 50: I kept the `else` for readability's sake, but removing it would still return the default already evaluated queryset.

xmedr (commented):
@hancush We've got default filtering working on the cms, thanks for your help again!

hancush (approved):

Issue #631, opened 2020-07-08
Board Reports page: Add info tips on facet headers


shrayshray:
Add hover over/tool tips to headers of each section of filters explaining what each facet is.

hancush:
Thanks, @shrayshray! 

@antidipyramid, @xmedr - Let's plan to implement these in the CMS, that way they can be managed by Metro staff directly.

PR #1227, opened 2025-03-24
Wagtail manageable tooltips for facets/topics

xmedr:
## Overview

This branch allows users to manage tooltips that appear next to search facets and topic categories listed in board report detail pages. To do so, the user will simply add/edit a tooltip with the name of the label they're looking to manage. If a label appears in both the search facets and detail categories, the same tooltip will be used for both. Users can also disable each tooltip if they'd like to keep what they've made, but want it to stop displaying. Only one tooltip for each label can exist at a time.

So in order to add a tooltip for Board Report Types, a user would add a Tooltip in the cms with the target "Board Report Type" with whatever content they'd like to insert in the rich text editor, and save. The tooltip will then appear in the search and detail pages, expanding when hovered or clicked.

- Connects #631 

### Demo
In the cms:
image

---

On the search page:
image

---

On the bill detail page:
image

## Testing Instructions
 - Log in and add some tooltips for facets/categories
 - Confirm that the process makes sense
 - Confirm that tooltips show up on the site as expected
 - Confirm that the new test is checking for the right behavior and passes


lametro/templates/legislation.html
```diff
@@ -293,5 +296,10 @@ <h1 class="h3">Legislation not found</h1>
         {% for source in legislation.sources.all %}
             console.log('source: {{source}}');
         {% endfor %}
+
+        var tooltipTriggerList = [].slice.call(document.querySelectorAll('[data-bs-toggle="tooltip"]'))
+        var tooltipList = tooltipTriggerList.map(function (tooltipTriggerEl) {
+            return new bootstrap.Tooltip(tooltipTriggerEl)
+        })
```
xmedr, line 303: Get bootstrap tooltips working on bill detail pages

lametro/templates/search/_search_filter.html
```diff
@@ -8,14 +8,16 @@ <h3 class="accordion-header">
             <button class="accordion-button collapsed" type="button" data-bs-toggle="collapse" data-bs-target="#collapse_{{facet_name}}" aria-expanded="false" aria-controls="collapse_{{facet_name}}">
                 Filter by
                 {% if 'Legislation' in facet_label %}
-                    Report Type
+                    Board Report Type
```
xmedr, line 11: Making this facet read exactly the same as the corresponding topic category on the bill detail page would make it clear to Metro that the same tooltip would apply to both spots.

lametro/wagtail_hooks.py
```diff
@@ -230,10 +231,50 @@ class FiscalYearCalendarViewSet(SnippetViewSet):
     )
 
 
+class TooltipFilterSet(django_filters.FilterSet):
```
xmedr, line 234: Following a pattern similar to what we did with alerts, the tooltips will default to showing only those that are enabled, and the filters can adjust from there.

tests/test_cms_models.py
```diff
@@ -25,3 +27,23 @@ def test_alert_expiration(client, alert, expiration_str, should_display):
...
+    metro_subject.build(name="Kiosks")
+
+    # Add this classification to make sure the bill is discoverable by the model manager
+    test_bill = bill.build(classification=["Board Box"], subject=["Kiosks"])
+    test_content = "A test tooltip for a subject!"
+    Tooltip.objects.create(target="Subject", content=test_content, disabled=disabled)
+
+    url = reverse("lametro:bill_detail", kwargs={"slug": test_bill.slug})
+    response = client.get(url)
+    assert (test_content in response.content.decode("utf-8")) == should_display
```
xmedr, line 49: Ideally I would've loved to have this also test whether the tooltip showed up in the search view, but had a hard time getting the `test_bill` to show up in the search results response.

xmedr (commented):
Something to note is while I was opening this, there was an issue with `setuptools` and a dependency of one of requirements (`docx2txt` from `textract`) where it required the formatting of its setup.cfg file to be a bit different - [see more here](https://setuptools.pypa.io/en/stable/history.html#v78-0-0). But after some time, they reverted the change and the build works again. They will reinstate these changes sometime in 2026.

Before they did that, I figured the best option would be to pin setuptools to an earlier version without that formatting requirement change. But for some reason even though the github action would say that an older version of setuptools was installed, it would still raise the error and cause tests to fail. Things are fine now, but I'm curious why that didn't work.

antidipyramid (commented):
What do you think about using a set dropdown menu with options instead of a free text field when adding tooltip text?

Screenshot 2025-03-25 at 4 46 00 PM

It looks like our facets/topics correspond to values that are already hard-coded in `lametro/models/legislative.py`:

```py
class LAMetroSubject(models.Model):
    CLASSIFICATION_CHOICES = [
        ("bill_type_exact", "Board Report Type"),
        ("lines_and_ways_exact", "Lines / Ways"),
        ("phase_exact", "Phase"),
        ("project_exact", "Project"),
        ("metro_location_exact", "Metro Location"),
        ("geo_admin_location_exact", "Geographic / Administrative Location"),
        ("significant_date_exact", "Significant Date"),
        ("motion_by_exact", "Motion By"),
        ("topics_exact", "Subject"),
        ("plan_program_policy_exact", "Plan, Program, or Policy"),
    ]
```

That way the tooltip entry/update process would be less prone to typos or other user error.

lametro/models/cms.py
```diff
@@ -272,7 +272,27 @@ def clean(self):
 class Tooltip(models.Model):
     include_in_dump = True
 
-    target = models.CharField(max_length=255, unique=True)
+    TARGET_CHOICES = [
+        ("board_report_type", "Board Report Type"),
+        ("fiscal_year", "Fiscal Year"),
+        ("geographic_administrative_location", "Geographic / Administrative Location"),
```
xmedr, line 278: I went with defining a separate set of similar choices here mostly because of the 3 extra options, and because we probably don't need the `_exact` suffix that the Subject model's choices have

lametro/templatetags/lametro_extras.py
```diff
@@ -374,6 +374,9 @@ def show_tooltip(label):
     else:
         tooltip_label = label
 
-    if Tooltip.objects.filter(target=tooltip_label).exists():
-        tooltip = Tooltip.objects.get(target=tooltip_label)
+    pattern = re.compile(r"[\W_]+")
+    tooltip_value = pattern.sub("_", tooltip_label.lower())
```
xmedr, line 378: Formatting the label to match our new values in the Tooltip model

xmedr (commented):
That's a great idea! Looks like there are 3 extra labels on the live site so I've added those as options as well

antidipyramid (changes requested):
Brilliant, @xmedr! I just noticed something a bit off with the "Disabled" option when creating/editing a tooltip:


https://github.com/user-attachments/assets/06639603-95c6-42b5-bbc6-4fd8bb41c131

One suggestion for your consideration: I think it'd be nice to have a little description explaining what it means for a tooltip to be disabled, similar to the helptexts under "Target" and "Content".

xmedr:
Ah interesting! It turns out that clicking on the icon for any panels in wagtail will collapse the content inside of it, so the disabled checkbox looks like it's just disappearing. I've changed the icon for that panel so that it's less confusing which item you're supposed to click.

Also good idea! I've got that help text in there.

image

antidipyramid (approved):
Amazing!

Issue #265, opened 2018-03-08
New page: Board Policies


shrayshray:
We'd like to add a new page about Board Policies, with a list of policies linking to the PDF of each in our library/archive, and its date of adoption. Initially I thought these links could be on the About page, but there are approximately 70 ... so, is it possible to create a new page for this? 
I can send a mock-up of design, and spreadsheet with links/dates.

reginafcompton:
@shrayshray - we can create a new page for this. The mock-up, as always, would be very helpful!

shrayshray:
@reginafcompton below is a a mock-up, and I'll post the spreadsheet of links soon!
![boardpolicies](https://user-images.githubusercontent.com/28063988/37232777-4733aa64-23a5-11e8-9ad7-05621938b2b3.png)

shrayshray:
@reginafcompton @hancush would it be possible for Omar and I to have access to edit/update this page once it's set up?

reginafcompton:
@shrayshray - we need a little time to discuss options. Can we get back to you by early next week?

shrayshray:
Of course!

reginafcompton:
@shrayshray - we talked about implementation options, and we see two ways forward:

(1) DataMade creates a new page and periodically makes content changes, whenever requested by Metro.
(2) DataMade adds a Content Management System to Councilmatic - it would allow Metro admins to edit text on any of the static views (e.g.,  the new board-policy page and the about page).

The first option is very lightweight, but does not give Metro much freedom, i.e., only DataMade can make editorial changes.

The second option would require considerable time to implement (around 40 hours), but it will offer Metro greater autonomy. 

We can put together a more precise estimate for the second option, if it sounds like something you'd like to consider. It could be an additional scope or we could use maintenance hours for it. Let us know!

reginafcompton:
(Notes for development)

Relevant links for building a CMS:
https://github.com/datamade/openness-project-nmid/blob/e6fd35e56cfd17f84b6a1fcbe046893ea463180e/pages/models.py#L7

https://github.com/datamade/openness-project-nmid/blob/e6fd35e56cfd17f84b6a1fcbe046893ea463180e/pages/admin.py#L9

https://github.com/datamade/openness-project-nmid/blob/e6fd35e56cfd17f84b6a1fcbe046893ea463180e/camp_fin/base_views.py#L524

https://github.com/datamade/openness-project-nmid/blob/master/camp_fin/views.py#L46

[...]

PR #1229, opened 2025-04-04
Add board policies page type

xmedr:
## Overview

This branch adds a new board policies page type that takes much from the about us page, but also adds a new aside block to match the wireframe in the connected issue.

The asides will reserve a portion of the page if any exist. Otherwise the section content can continue to take up the full width. All asides will be stacked to the right as shown, and will be pushed to the bottom of the page at smaller screen sizes.

- Connects #265

### Demo
**Large screen with aside**
image

---

**Smaller screen with aside**
image

---

**Large screen without aside**
image

### Notes

Once this comes in, we can create an empty Board Policies page for them (with the correct slug) so that the nav link doesn't lead to an error.

## Testing Instructions
 * Head to the review app
   * It has been set up for the migrations involved


.pre-commit-config.yaml
```diff
@@ -16,4 +16,4 @@ repos:
     hooks:
       - id: flake8
         files: ^(lametro|councilmatic|tests)
-        exclude: lametro/migrations councilmatic/settings.py
+        exclude: ^(lametro/migrations|councilmatic/settings\.py)
```
xmedr, line 19: Migrations weren't actually being excluded from flake8 without this patterning.

README.md
```diff
@@ -150,9 +150,11 @@ Grab the values for the `AWS_ACCESS_KEY_ID` and the `AWS_SECRET_ACCESS_KEY` and
 Now you should be able to start uploading some files!
 
 ## Review Apps
-This repo is set up to deploy review apps on Heroku, and those pull from the staging database to match the experience of deploying as closely as possible! However, note that in order to prevent unapproved model changes from effecting the staging database, migrations are prevented from running on review apps. So those will still have to be reviewed locally.
+This repo is set up to deploy review apps on Heroku, and those pull from the staging database to match the experience of deploying as closely as possible! However, note that in order to prevent unapproved model changes from effecting the staging database, migrations are prevented from running on review apps by default.
 
-### Review apps for Wagtail changes
+If a branch involves a migration, either review those changes locally, or check out the following instructions to provision and populate a database for review apps.
```
xmedr, line 155: After doing this a few times, it seems that the instructions below would work fine for more than just wagtail changes and we wouldn't have to review any migrations locally if we don't want to. So I've adjusted the language to make that clear. Let me know if this isn't the case!

xmedr:
@antidipyramid ah they do now! I just had to add localhost to the allowed hosts. i'll change the review apps to always include that

image

antidipyramid (commented):
Board policies page editing and publishing works great, @xmedr! After publishing, the link at `/board-policies-page/` resolves correctly.

One thing I noticed was that the nav bar link to the board policies page doesn't seem to update correctly. On my end it links to `/board-policies-page/None/`.

Did I create the page in the right place in Wagtail?

Screenshot 2025-04-21 at 9 28 56 AM

xmedr:
@antidipyramid You definitely did create the right kind of page. The issue was that since it was named "Board policies page", the slug for it was also `board-policies-page`, but the navbar was looking for a shorter slug `board-policies`. I changed the slug for your page in wagtail and now the link works again.

I imagine when this page gets set up, either we or metro will have to make sure the slug is `board-policies`. Unless there's a more elegant solution I'm not thinking of.

image

image

antidipyramid (approved):
Works great, @xmedr.

Can the board policies page slug with be set using [this field](https://docs.wagtail.org/en/6.3/reference/pages/model_reference.html#wagtail.models.Page.slug)?

xmedr:
@antidipyramid Ah you're right it probably can! I could work on adding a migration that creates an empty board policies page, but I think it might be easier at this point to get this in and create it for them since we're trying to get this ready this week.

Issue #1233, opened 2025-04-30
FY26 Calendar


neilarellano:
Can we please add the new calendar for FY26. Can we have it show right underneath the current FY25 calendar?

Please see attached.

[FY26 Committee and Board Meeting Calendar.pdf](https://github.com/user-attachments/files/19983007/FY26.Committee.and.Board.Meeting.Calendar.pdf)

neilarellano:
Hi Team,

Can we please get this put up on the website soon?

Thanks!

antidipyramid:
Hi @neilarellano, thanks for re-upping. We'll definitely get the new calendar up today.

PR #1234, opened 2025-05-05
Add FY26 calendar links

antidipyramid:
## Overview

See title.

Connects #1233

### Demo

Screenshot 2025-05-05 at 3 24 52 PM

## Testing Instructions
 * Verify that there are fy26 calendar links on the home page and events page


xmedr (approved):
Works as expected! The links in the events page might look better if they were on the same line, but not a huge deal if we want to post it as is!

image

Issue #1215, opened 2025-02-18
Create pattern for handling general downloadable files


xmedr:
The Fiscal Year calendar is something that we manually update whenever Metro has a new one available. Let's start having things like that managed within Wagtail! In this case, we can have the app check for a fiscal year file within our wagtail managed files using something like a standardized naming convention and use that. If it doesn't exist, then we can fallback to the version uploaded locally.

If there are other spots that have files like this, then we can apply this pattern there as well.

antidipyramid:
@xmedr to add functionality to have two calendars on the site at a time. (see #1233 for example)

PR #1237, opened 2025-05-14
Allow for two wagtail manageable calendars

xmedr:
## Overview

This branch allows users to manage two calendars in wagtail - one `current` and one `upcoming`. After implementing this initially with just one calendar, it became apparent that metro will occasionally need to have the next fiscal year's calendar available for viewing before it starts.

- Connects #1215 

### Demo

image

---

image

---

image

### Notes

The review app's db has been set up for testing

## Testing Instructions
 * Log in to the review app and head to the Fiscal Year Calendars nav link
   * Confirm that two empty calendar objects exist, one labeled `current`, and the other `upcoming`
 * Upload a calendar to one or both of them
   * Confirm that the links on the index and events pages update to show one or two calendars, and that their links direct to the files you uploaded
 * Try to change the type of one calendar to the type of the other
   * Confirm that you get an error about only allowing one calendar of each type to exist
 * Try to add a new calendar
   * Confirm that you get an error about only allowing 2 calendars to exist at a time


lametro/models/cms.py
```diff
@@ -276,23 +281,36 @@ class FiscalYearCalendar(models.Model):
...
+        num_types = len(self.TYPE_CHOICES)
+        is_new_cal = not model.objects.filter(id=self.id).exists()
+
+        if model.objects.count() >= num_types and is_new_cal:
             raise ValidationError(
-                "Only one calendar can exist at a time. "
-                "Please edit the existing calendar object."
+                f"Only {num_types} calendars can exist at a time. "
+                "Please edit the existing calendars."
             )
```
xmedr, line 307: I figure that basing this first conditional off of the number of possible calendar types would make it easier to just drop in new types if/when they want another.

antidipyramid (approved):
Can confirm everything works as expected.

Just a suggestion for possible future iterations: the Wagtail docs have a nice [example](https://docs.wagtail.org/en/stable/extending/admin_views.html#creating-admin-views) of adding a calendar view to the admin interface. I wonder if we could simplify it (from Metro's perspective) by doing something like they show.

Screenshot 2025-05-23 at 11 22 46 AM

Also, Metro tends to prefer the calendars to stack but we can tackle that in the future as well.

Issue #1238, opened 2025-05-21
QA for CMS v0.2


hancush:
### New features

- #1189 
- #1202
- #1216
- #1223
- #1227
- #1229
- #1220

### Diff

https://github.com/Metro-Records/la-metro-councilmatic/compare/main..feat/cms-epic

## Testing instructions

- Follow the testing instructions in each of the listed PRs to confirm all functionality works as intended
- Check core areas and functionality of the site to confirm existing features have not be affected

xmedr:
I just tested all of these pr’s ahead of handing the testing over to metro and found a small handful of issues that I can address:
- **Board member details**: For some reason, the “member of” filter returns members that don’t match.
- **Alert expiration**: The expiration part works just fine, but the colors of alerts don't match between the admin and the site.
- **Board policies page**: Would be nice to have a check for whether to display the link for this in the nav. That way metro has some time to create the page and there isn't a dead link on the site.
- Every other pr worked as expected.

One thing to note is that for the "wagtail modeladmin to snippets" change, I noticed an event notice was appearing on events that had passed even if the notice wasn't set that way. But I'm fairly certain this is because this is a staging site that is probably missing some scrape or check for past events. The same does happen locally for me on the main branch which doesn't have this change, but does not happen on regular staging and prod. I don't believe this will be an issue for those environments, but we should double check regular staging when we do push this up.

xmedr:
Fixes and changes from main have been merged into the `cms-epic` branch, and testing has been passed on to Metro! I've suggested a deadline of June 6th.

PR #1220, opened 2025-03-11
Migrate event agenda management to CMS

hancush:
## Overview

See title. Part of our grand transition to Wagtail. As part of cleanup, this PR also breaks most of the event view code into a service and refactors the event admin actions into a unified interface.

Connects #1214

### Demo

Screenshot 2025-04-18 at 10 38 32 AM
Screenshot 2025-04-18 at 10 38 49 AM
Screenshot 2025-04-18 at 10 38 41 AM
Screenshot 2025-04-18 at 10 38 57 AM

## Testing Instructions
* Follow the instructions in the README to initialize and populate a database for the review app
* Visit some events with an agenda and confirm the event admin module says "This event's agenda has been scraped."
* Visit an event without an agenda and confirm the event admin module prompts you to upload one.
* After you add one, confirm the event admin module has a link to manage the manual agenda.
* Delete the manual agenda you added and confirm that the corresponding document still exists in Wagtail, but the event admin module flipped back to the prompt to upload one (i.e., the document link was removed).

PR #1239, opened 2025-05-21
Handle non-event views in EventAgendaLink

hancush:
## Overview

This PR addresses a bug introduced by #1220 which prevented any detail pages besides event detail pages from loading.

Connects #1238 

## Testing Instructions

* Visit event, person, and board report detail pages and confirm they render without error.


xmedr (approved):
Nice catch, all detail pages work as expected! I tested locally instead of on the review app since waiting for the scrape to run and the db to build would've taken a long time

Issue #1224, opened 2025-03-13
Expose tag analytics in CMS


hancush:
We currently run tag analytics via the dashboard and upload reports to a Google Drive once a month. We could easily create a section in the CMS that allows Metro admins to export that report at any time instead. This section could also include some basic reporting, if desired.

Thoughts, @shrayshray, @neilarellano, @camachoo?

PR #1263, opened 2025-07-21
Generate tag analytics from cms

xmedr:
## Overview

This branch allows logged in users to generate google tag analytics whenever they'd like using either a link in the Reports section of the admin or a link in the wagtail userbar on any front end page.

After a short load, the page will refresh and a message will display confirming that a report is being created. Once the report is made and uploaded, the user who requested the report will receive an email notification.

If the user has already queued up a job to make those analytics that is still in progress, they will be shown a message asking them to wait until the last one is done before trying to make a new one.

- Connects #1224 

### Demo

image

---

image

### Notes

I've given this review app and my local app the contents of "LA Metro Google Service Account Key" in bitwarden as a string in order to upload to google drive, and I get a success message from both environments.

However, I'm not sure if I'm looking at the right folder and can't quite confirm that it's actually been uploaded. I'm looking at a folder called "LA Metro Reports" and the latest report is from June 30th. Maybe the folder id I'm using is for a different folder entirely. Alternatively, I may be missing something to upload to that folder.

## Testing Instructions
 * Create a superuser if needed
 * Try to navigate to `/generate-tag-analytics/` from the address bar without having logged in first
   * Confirm that you get redirected to a log in page and it does not generate the analytics yet
 * Do the same after logging in
   * Confirm that you get a message asking you to do it from one of the links on the page. 
 * Log in, and either click Reports > Generate Tag Analytics from the cms, OR head to the front end, open the wagtail userbar, and click "Generate tag analytics"
   * Confirm that you get a success message at the top of the page and that the analytics are being generated.
   * Try starting another job immediately, and confirm that a message displays asking the user to wait until the current job is done before starting another.
   * Confirm that 10 - 30 minutes later, you receive an email saying the report has been uploaded.


antidipyramid (changes requested):
One suggestion for you inline for easier testing.

lametro/management/commands/generate_tag_analytics.py
```diff
@@ -27,17 +28,45 @@ def add_arguments(self, parser):
...
+        email_kwargs = {
+            "from_email": getattr(settings, "DEFAULT_FROM_EMAIL", None),
+            "recipient_list": [email],
+            "fail_silently": False,
+        }
 
         if local:
             with open(output_file_name, "wb") as f:
                 copyfileobj(csv_string, f)
+            self.stdout.write(
```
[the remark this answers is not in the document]
xmedr: @antidipyramid no worries, tysm! this worked perfectly, so i've got that pushed up here. this is an example of what this looks like when i print it to the console:

```bash
Content-Type: text/plain; charset="utf-8"
MIME-Version: 1.0
Content-Transfer-Encoding: 7bit
Subject: Councilmatic: Tag Analytics Generated!
From: lametro@councilmatic.org
To: xavier.medrano@datamade.us
Date: Wed, 30 Jul 2025 14:32:24 -0000
Message-ID: <175388594406.1.1957392949959102496@536da156dcd2>

Your google tag analytics have been generated, and are now available in the google drive folder. Check it out here: https://drive.google.com/file/d/1oJsrC-4VKaxewHNoSioliKa9jgHw1t-8/view
```

lametro/views.py
```diff
@@ -840,6 +845,54 @@ def get_context_data(self, **kwargs):
...
+            try:
+                response.raise_for_status()
+            except requests.exceptions.HTTPError:
+                messages.error(
+                    request,
+                    f"Received a {response.status_code} status code from Heroku. "
+                    "Analytics not generated.",
+                )
+            else:
+                messages.info(request, "Tag analytics are being generated...")
```
antidipyramid, line 886: It'd be nice to let the user know to expect an email containing a link to the generated file. We should also say that the file take some time to generate.

lametro/views.py
```diff
@@ -840,6 +845,54 @@ def get_context_data(self, **kwargs):
...
+            logger.info(f"Sending command to Heroku: {command}")
+            response = requests.post(url, json=data, headers=headers)
+            logger.info(
+                f"Got {response.status_code} response from Heroku: {response.json()}"
+            )
+
+            try:
+                response.raise_for_status()
+            except requests.exceptions.HTTPError:
+                messages.error(
```
antidipyramid, line 880: What do you think about defining some custom exceptions for failures during this process (and in the management command itself)? Otherwise, we might have silent errors or have to rely on users to notify us of issues.

antidipyramid (changes requested):
Left two suggestions for you inline!

lametro/views.py
xmedr: Good point, I've got that in there now!

antidipyramid:
@xmedr Do you also get an error when typing `https://la-metro-cou-feat-cms-a-shawj0.herokuapp.com/generate-tag-analytics/` into your browser?

Screenshot 2025-08-20 at 4 45 12 PM

One other question: what happens if someone starts one tag analytics job but then goes to the page again before the first has a chance to finish? Does that start two simultaneous jobs?

xmedr:
@antidipyramid Ah I do! That'd actually be a great place for another try block. I'll make it so that we check to see if an `HTTP_REFERER` exists before trying to start the command, and redirect to the cms homepage or something if it doesn't. That way we can ensure this only ever gets initiated from the app instead of pasting a link.

You're right, that would start two jobs doing the same thing. It's definitely possible to check the previously started job's status in heroku when someone attempts a new one. I'll just have to figure out how best to either keep track of that previous job's id, or check if any active workers are running this command

xmedr (commented):
@antidipyramid This is ready for your eyes again! I've got those changes in, and set the review app's `DEBUG` value to False so that the cache actually gets used. I've also updated the pr description to match this

PR #1274, opened 2025-10-15
Speed up search result highlighting

antidipyramid:
## Overview

This PR should take care of an N+1 query in the search result template. It also makes sure we don't load unnecessary related objects during searches by using Haystack's `load_all` parameter.

Connects #1151 #1049

## Testing Instructions
 * Verify that searches work as expected


hancush (approved):

xmedr (approved):
Nice, searches feel noticeably faster than on staging!

PR #1291, opened 2026-03-04
Set up notifications for translation suite

xmedr:
## Overview

This branch adds two management commands that will notify the translation suite of new documents to be processed

### Notes
This pr includes a migration, so the review app has been set up with it's own database. The review app is also set to send notifications to the translation suite's staging app.

Also despite the migration and extra service methods, this branch is currently safe to bring in before the translation suite is finished. Any extra entries to the new notifications model would only be added after we start running the management commands.

## Testing Instructions
 * Run the management command on the review app to build notifications
   * `heroku run -a la-metro-cou-feat-trans-18nuxo python manage.py build_translation_notifications`
 * After that completes, run the management command on the review app to send a notification
   * `heroku run -a la-metro-cou-feat-trans-18nuxo python manage.py notify_translation_suite`
 * Confirm that the logs for both print useful information and a success message
 * Run the same two commands again
 * Confirm that the logs say that all documents/notifications are up to date
 * Open a shell on the translation suite's staging app to confirm those documents got made
```python
heroku run -a la-metro-translations-staging python manage.py shell

>>> from la_metro_translations.models import Document
>>> Document.objects.count()
# Example output: 55
```


lametro/management/commands/notify_translation_suite.py
```diff
@@ -0,0 +1,95 @@
...
+    def handle(self, *args, **options):
+        """
+        Send notifications for bills and events recently created/updated if the entity:
+        - has not yet had a notification sent
+        - has been updated more recently than its last notification
+        - does not have a successful notification
+        """
+
+        now = datetime.now()
+        cutoff_date = now - timedelta(weeks=12)  # 3 months ago
```
xmedr: Metro mentioned that 3 months worth of documents should be fine for the initial run of this command during the pilot, and we can work on getting more historical documents in gradually.

lametro/management/commands/notify_translation_suite.py
```diff
@@ -0,0 +1,95 @@
...
+
+        was_successful = str(response.status_code).startswith("20")
+        if was_successful:
+            logger.info(f"Success: {response.json()}")
+        else:
+            logger.warning(f"Failed: {response.json()}")
+
+        # Upsert notification objects for each entity
+        self.upsert_notifications(bills, "bill", was_successful, now)
+        self.upsert_notifications(events, "event", was_successful, now)
```
xmedr: I was hoping to work some more `bulk_create` + `update_conflicts` magic, but turns out the conflicts flag doesn't exist until django 4.1, and we're using 3.2 here.

lametro/models/legislative.py
```diff
@@ -1181,3 +1191,46 @@ def __str__(self):
 
         else:
             return self.name
+
+
+class TranslationNotification(models.Model):
+    """
+    Tracks notifications sent to the Translation Suite
+    """
```
xmedr, line 1205: I figure this model would be a decent way of tracking which entities have had notifications sent for them. This is restricted to needing a relation to either an `LAMetroBill` or an `LAMetroEvent`, but not both.

Also, I'm relating the notifications to the entities instead of the `BillVersions` (board reports) or `EventDocuments` (agendas), because neither of these has an `updated_at` field, just a date field. And board reports don't seem to populate that field anyway.

xmedr (commented):
After this comes in, I'll need to:
- [ ] add the three new env vars to the rest of our councilmatic apps

lametro/utils.py
```diff
@@ -145,3 +145,50 @@ def trace_function(frame, event, arg):
         sys.settrace(None)
 
     return resp
+
+
+def get_alternate_files(entity, type, friendly_name) -> dict:
```
hancush: Hm, I'm not sure about this approach (fires a lot of requests, repeatedly), and it doesn't seem related to this changeset. Want to open an issue, but remove this and address it separately?

lametro/management/commands/notify_translation_suite.py
```diff
@@ -0,0 +1,94 @@
...
+        logger.info("Finished processing notification")
+
+    def upsert_notifications(self, entities, type, was_successful, now):
+        for entity in entities:
+            data_kwargs = {
+                # Populate either the 'bill' or 'event' field with this entity
+                type: entity,
```
hancush: type or entity_type?

lametro/services/event_service.py
```diff
@@ -161,3 +162,35 @@ def get_notices(event) -> QuerySet:
                 )
 
         return notices_by_comment.filter(broadcast_conditions__contains=["future"])
+
+    @staticmethod
+    def build_events_notification(
```
hancush: This should operate on one event. Can you move the loop upstream?

lametro/services/bill_service.py
```diff
@@ -0,0 +1,43 @@
...
+from django.db.models import QuerySet
+
+from lametro.models.legislative import LAMetroBill
+
+logger = logging.getLogger(__name__)
+
+
+class BillService:
+    @staticmethod
+    def build_bills_notification(
```
hancush: Ditto operating on one instance.

lametro/management/commands/notify_translation_suite.py
```diff
@@ -0,0 +1,94 @@
...
+            logger.info(f"Success: {response.json()}")
+        else:
+            logger.warning(f"Failed: {response.json()}")
+
+        # Upsert notification objects for each entity
+        self.upsert_notifications(bills, "bill", was_successful, now)
+        self.upsert_notifications(events, "event", was_successful, now)
+        logger.info("Finished processing notification")
+
+    def upsert_notifications(self, entities, type, was_successful, now):
```
hancush: I think we want to create new notifications, even if they related to the same entity. You can tell your sending management command to only send the most recent notification for each entity (and maybe introduce a discard flag, in the event that a notification was never successful but was superseded by a subsequent notification). Make sense?

hancush (changes requested):
This is a really good start! I think we need two management commands: one to create the notifications, and another to send them (which gives us retries in the event of failures). What do you think?

lametro/management/commands/build_translation_notifications.py
```diff
@@ -0,0 +1,143 @@
...
+        "or events that need their documents OCR'd and translated."
+    )
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--weeks",
+            type=int,
+            default=12,
+            help="Consider a entity, if it was updated within the last N weeks.",
+        )
```
xmedr, line 36: Found myself adjusting the cutoff date a lot, so now we just have an optional flag

lametro/management/commands/notify_translation_suite.py
```diff
@@ -0,0 +1,109 @@
...
+        for related entities.
+
+        On failure, retry a set number of times, ultimately marking them as "failed"
+        if we exceed our retry limit. The `build_translation_notifications` command
+        will make new "waiting" notifications for these entities next time it runs.
+        """
+
+        notifications = TranslationNotification.objects.select_related(
+            "bill", "event"
+        ).filter(status="waiting")
```
xmedr: The way the two commands currently work with each other, there can only be a max of one "waiting" notification per entity, and only if that entity needs it in the first place. That "waiting" notification will also always be the most recent one, if it exists. So sending all that are "waiting" and updating them on success feels like a safe choice in this case.

lametro/management/commands/build_translation_notifications.py
```diff
@@ -0,0 +1,143 @@
...
+            f"and {len(event_notifications)} events"
+        )
+        logger.info("~~ Finished building notifications! ~~")
+
+    def create_notifications(self, qs):
+        """
+        Create a notification object for each entity in the queryset if that entity
+        has relevant related documents. If an entity has no documents to report on,
+        a notification will not be made.
+        """
```
xmedr, line 122: I wanted to make sure that we're only making notifications for entities that have documents to report on. So I'm performing a check here with the bill/event service details method to see if it does.

I'm also now saving the details that will be sent with the notification, _to_ that notification. That way, the sending command doesn't have to make this same check again to build the details.

xmedr (commented):
Alrighty, I've got all of those suggestions in! We're now able to assign multiple notification objects to one entity, so we can queue up a new notification while having info on old ones.

We've also now got two mgmt cmds:
- `build_translation_notifications`: makes new notifications with a status of "waiting", and data on documents that the suite needs to know about, while cleaning up old "failed" notifications for related entities.
- `notify_translation_suite`: sends the data contained in all "waiting" or "failed" notifications. if it fails, it waits and retries 2 more times, marking notifications as "failed" if it never ends up working, or marking as "delivered" if it does work

lametro/management/commands/build_translation_notifications.py
```diff
@@ -124,7 +124,7 @@ def create_notifications(self, qs):
         # Determine which entity type this queryset is for, and choose the right builder
         if not qs:
             return
-        elif type(qs[0]) is LAMetroBill:
+        elif qs.model is LAMetroBill:
```
hancush, line 127: cc @xmedr !

hancush (approved):
👏

Issue #1273, opened 2025-10-11
Item Listed Without Agenda Published


neilarellano:
Hi Team,

I am seeing that our 10/15/2025 Planning and Programming Committee meeting is showing 2025-0805 listed before the agenda was exported/published. This file ID also exists within the Executive Management Committee meeting which was exported/published first.

Should we be preventing items from being listed before an agenda is published for the meeting?

<img width="2145" height="1111" alt="Image" src="https://github.com/user-attachments/assets/81d155f7-827d-4fcb-8e8e-d5bb44ebcffe" />

PR #1292, opened 2026-03-05
Add person titles to wagtail + handle events with unpublished agendas

antidipyramid:
## Overview

Connects #1287 
Connects #1273 

## Testing Instructions
* Verify that events with unpublished agendas do not have any listed agenda items
* Verify that you can add a title for board/committee members
* Verify that the title set in wagtail displays on the person's detail page


xmedr (commented):
All looked good. The regular board meeting on 3/26 on production does not have an agenda but currently displays agenda items, whereas the same meeting on the review app does not. Title customization works as expected too 👍 

My only question is, should the custom titles also appear on the list of board members on the About > Board of Directors page?

antidipyramid:
We can confirm with Metro but I imagine not? I think custom titles are most relevant for committee members who aren't board members. So those members would only have their titles show up on a committee detail page.

PR #1299, opened 2026-05-06
Add workflows for notifying translation suite

### LINKED ISSUE #3: Create a webhook to receive new document notifications from BoardAgendas app
TK

xmedr:
## Overview

This pr adds a couple of workflows to notify the translation suite once a day about new documents. One workflow builds the notifications and then sends them. The other workflow sets it up to run once a day for both staging and prod.

Closes https://github.com/datamade/la-metro-translations/issues/3

### Notes
The prod job is commented out for now, but we can uncomment it when we start showing translations on this app!

If this workflow ever fails in the distant future, it'd be worth looking into whether the [database url for either environment has automatically changed on heroku](https://help.heroku.com/SMVK5PMS/why-have-my-heroku-postgres-credentials-changed).

We can also look into building a listener for that event so we can update it automatically on our side if/when that happens. We could use a [webhook on this app that listens for env var changes](https://devcenter.heroku.com/articles/app-webhooks) on prod/staging, and then connects to the [github api to change the secret](https://docs.github.com/en/rest/actions/secrets?apiVersion=2026-03-10#create-or-update-a-repository-secret) for each environment.

## Testing Instructions
 * Check [this run](https://github.com/Metro-Records/la-metro-councilmatic/actions/runs/25450165638) of the action and confirm it worked as expected
   * It currently says no notifications need to be sent because it sent 657 of them between both staging apps during local testing 🙌

### DIFF
--- a/.github/workflows/daily_scheduler.yml
+++ b/.github/workflows/daily_scheduler.yml
@@ -1,0 +1,20 @@
+name: Daily translation notifications dispatcher
+
+on:
+  pull_request:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 23 * * *" # Runs every day an hour before midnight UTC
+
+jobs:
+  staging:
+    uses: ./.github/workflows/notify_translations.yml
+    with:
+      environment: staging
+    secrets: inherit
+
+  # production:
+  #   uses: ./.github/workflows/run_translation.yml
+  #   with:
+  #     environment: production
+  #   secrets: inherit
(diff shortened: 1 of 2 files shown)


xmedr:
I just noticed there's a trigger in the translation suite workflow to [translate/convert documents on pull request](https://github.com/datamade/la-metro-translations/blob/f254fef90275a35b48aaba79f725ba068791f846/.github/workflows/daily_scheduler.yml#L4). Are we okay if I remove that trigger later? I imagine that was for testing the action, but lmk if we'd rather keep it

xmedr:
After this comes in, I'll need to:

- [x] configure a `production` github environment with its own secrets
- [x] update an appropriate issue in the project to remind us to uncomment the production section in this workflow when we're ready

antidipyramid (approved):

Issue #1296, opened 2026-04-03
No Meetings in The Past Two Weeks Text


neilarellano:
Hi Team!

I noticed that the "Recent Meetings" section on the boardagendas homepage shows that there are no meetings in the past 2 weeks.

<img width="708" height="175" alt="Image" src="https://github.com/user-attachments/assets/b045704f-4443-40ee-8c18-7346934b73ad" />

Metro was having technical difficulties for the past few meetings and was not able to stream these meetings. We eventually uploaded the meeting videos for the last few meetings but still this section shows that there are no recent meetings.

Can we please check the logic on this and how we can correct this?

Thank you.

antidipyramid:
Hi @neilarellano, the site will currently only list meetings in the "recent meeting" section that have had live broadcast links (manual or automated).
 
We could adjust the recent meetings section to display all past meetings whether or not there were live broadcast links. In that case we'd also want to account for cancelled meetings. What do you think?

neilarellano:
I like this idea. Do we have exclusions set for this such as test meetings?

yshng:
Closed by #1302.

yshng:
I've removed the broadcast requirement for listing recent past meetings, relying on "cancelled" status to tell us not to display meetings that didn't happen. (Test meetings are already excluded by earlier checks)

PR #1302, opened 2026-05-20
Show recent meetings that did not have a live broadcast

yshng:
## Overview

Look up recent meetings without checking broadcast status. 

Connects #1296 

### Notes

Initially discussed (with @derekeder) keeping the former `has_passed` method as a first pass and doing something else if that list was empty, but it could be the case that some meetings within the past 2 weeks had broadcasts and some didn't. Then our list would still be missing recent meetings that had no live broadcast, even though it wouldn't be showing the "No Recent Meetings" text. 

Assuming setting the `has_passed` cached property is still a fine way to mark meetings that have ended (versus checking for media, for example). Any place that relies on it will be wrong about events where broadcast media was added later, despite lack of live broadcast. Kept development scoped to Recent Meetings list only (barring creation of named constant due to reuse of magic number).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,7 +12,7 @@ repos:
       - id: black
 
   - repo: https://github.com/pycqa/flake8
-    rev: '9f60881'
+    rev: 'c48217e1'
     hooks:
       - id: flake8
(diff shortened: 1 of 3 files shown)


antidipyramid (approved):
Looks good to me!

Issue #1267, opened 2025-09-09
Hide the 2028 Olympic & Paralympic Games/Quarterly Special Board Meeting Body


neilarellano:
Greetings Team,

We disabled the 2028 Olympic & Paralympic Games/Quarterly Special Board Meeting body from showing insite and it no longer shows within here:
https://metro.legistar.com/Departments.aspx

How can we remove this body from showing under About > Committees?

https://boardagendas.metro.net/committee/2028-olympic-paralympic-gamesquarterly-special-board-meeting-17a3a85110d0/

hancush:
Hi, @neilarellano - There isn't a way for you all to manage committees at the minute, but I've removed the quarterly meeting body from the site. Traditionally, we've used the body type (committee, independent taxpayer oversight committee) to determine which bodies to scrape, which gives you all some control of what comes through. To prevent the quarterly meeting from coming back in the nightly person/organization scrape, you can change the body type, or we can add an exception in the scraper to skip this particular meeting body. What makes most sense on your end?

hancush:
@neilarellano - We don't, because deleted committees will be rescraped if they still meet the scraper's conditions. We could add a way for you all to disable the display of committees we've scraped in the CMS, though. Thoughts?

hancush:
Morning, @shrayshray - You're absolutely right! Committees with no active memberships will not be displayed, so that should cover the case where committees are retired, like this one.

I was thinking more in terms of putting the ability to manage what shows up on the board agendas site in your hands instead of hard coding it into the scraper. Right now, if you wanted to bring in a new body type, for instance, we'd have to update the scraper, but we could consider a model where we scrape everything and you all are able to control what shows up on the site, either on organizations directly and/or via some attribute. Maybe we can chat more at our next monthly priorities meeting?

neilarellano:
Hi Team,

I just wanted to revisit this. I like the idea of being able to control what shows up. If we're looking at the meeting bodies from [here](https://metro.legistar.com/Departments.aspx), we can just use the number of members as an indicator of whether or not a meeting body is active/shown.

I like the idea of having a way to hide/show meeting bodies manually in the CMS for one off situations like the one that came up for this issue.

Thank you.

PR #1303, opened 2026-05-26
Allow selective display of committees using Wagtail

yshng:
## Overview

Create a Wagtail setting for selectively ~~displaying~~ hiding committees under Settings which allows you to turn off select committees and save. The view will then use the display setting as a filter and ~~show only what is still~~ hide what's selected.

~~This is set up such that~~
~~- ALL boxes being checked AND~~
~~- NO boxes being checked~~
~~will both display ALL committees with members.~~

~~In the first case, it's because everything is affirmatively checked.~~

~~In the second case, it behaves as if the display settings do not exist, and goes back to default behavior of showing all committees with more than one member.~~

~~Helpfully, the help text correctly describes this behavior.~~

(Turns out the above stricken was a hint that the solution was wrong.)

Connects #1267

### Notes

Tried with allow list first, then changed to deny list. See commit messages and comment below.

### Testing

No additional tests written.

Manual testing: 
- Log into Wagtail Admin
- Go to Settings > Committee Display Settings
- Check some boxes and save
- Go to Committees page and see if your committees are there.


yshng:
Realized the approach was wrong when writing my own notes on the PR. New approach allows you to decide to hide certain committees rather than deciding which ones to let through. This way if the scraper finds new committees, they don't need to be "made visible" by being added to the allow list; they are visible by default by virtue of not being on the deny list. No extra admin work needed.

antidipyramid (changes requested):
Thanks, @yshng. In the review app I'm getting an error on the Committee Display Settings page. Do you know why that is?

Image

yshng:
@antidipyramid It's because the PR includes a migration, you'll need to test locally unless you want to walk me through provisioning a db.

antidipyramid:
@yshng Got it. If you want to provision a db for a review app in the future, there are some instructions on how to do that in the [README](https://github.com/Metro-Records/la-metro-councilmatic#review-apps).

yshng:
@antidipyramid Sorry - I did see that in the README, but did not know where/how to actually implement it. Luckily @xmedr was already going to walk me through Heroku this morning so we provisioned a db as part of that. It's still getting set up so the deploy preview is not fully functional at the moment, but I can ping you when it finishes scraping etc.

antidipyramid (changes requested):
Works great for me. But what do you think about adding this feature as a top level page on the toolbar, similar to Board Member Details and Event Agendas? 

I'm thinking that that would be inline with how Metro currently interacts with objects on the site. It also keeps the Settings tab reserved for Wagtail specific functionality like user/site management.

Image

yshng:
@antidipyramid Yes that sounds right to me, will change and re-request.

yshng:
@antidipyramid Wagtail now skips the Committee Display Settings item when constructing the settings menu, then I've added a item to the main menu to access it.

yshng:
Oh and upon approval, I would like to interactively rebase these commits before merging into main. Otherwise the code is broken at several points between commits. cc/@derekeder

antidipyramid (approved):

Issue #1297, opened 2026-05-05
Use MatterBodyName instead of Sponsor


neilarellano:
We noticed that filtering by meeting has some issues:

Inconsistency with Meeting Body and Tag display. Here is a screenshot of the results on the search page when I use "Filter by Meeting" for "Executive Management Committee". The meeting body is often showing the incorrect body, and the tag is showing a different meeting body, even though we don't tag meeting bodies? They're assigned in Legistar. So it's unclear why these Legistar-assigned fields would be populating incorrectly.

<img width="1466" height="720" alt="Image" src="https://github.com/user-attachments/assets/5d88a931-78c5-4c5d-842d-ba705b39ea78" />

We found out that this is because of the "Sponsor" field is not maintained very well and it is inconsistent. Instead of using the Sponsors field here:

<img width="1009" height="516" alt="Image" src="https://github.com/user-attachments/assets/e2c7b804-21f8-48f0-a373-a4f190b4e241" />

Can we instead use the MatterBodyName via API since this is more accurate.

neilarellano:
Instead of showing the sponsors field on the search results as seen here:

<img width="1363" height="740" alt="Image" src="https://github.com/user-attachments/assets/bb568508-3551-418b-abdf-e577b8d80461" />

Can we instead display the In Control Body (MatterBodyName)?

yshng:
@neilarellano
Running the new scrape with In Control body (MatterBodyName) in staging has surfaced at least one body that doesn't seem to have a corresponding Meeting Body: "2028 Olympic & Paralympic Games/Quarterly Special Board Meeting"

The closest meeting body I was able to find was "Special Board Meeting/Ad Hoc 2028 Olympic & Paralympic Games Committee"

So the immediate question is, what body should "2028 Olympic..." map onto? 

And the larger question is, is the In Control body list not sourced from the same pool as Meeting Bodies? If so, we will continue to run into this sort of problem.

neilarellano:
Hi @yshng ,

I want to clarify what you are asking here.
As for "2028 Olympic & Paralympic Games/Quarterly Special Board Meeting", there was an issue for this (1267) where we had to stop it from showing insite.

I just re-enabled this body to show up insite.

As for your question: _is the In Control body list not sourced from the same pool as Meeting Bodies?_
They are sourced from the same pool. However we have the option to not show a certain meeting body/in control body from showing on the Legistar insite. I'm not sure how that affects how things look like via API.

[...]

PR #1306, opened 2026-06-03
Use controlling body (MatterBodyName) instead of primary sponsor for search filters

yshng:
## Overview

In tandem with a [corresponding PR ](https://github.com/Metro-Records/scrapers-lametro/pull/62)in the scraper, we can now present the controlling body in the search results when filtering by Meetings.

Connects #1297

### Testing
Until the scraper you're using alongside your deployment of this PR, actually gets the above PR and starts populating `controlling_body` correctly, you will only get "Board of Directors" in the Meetings filter because the scraper is currently hard-coded to provide that for `controlling_body`

That also means of course that PR should be merged before we merge this one. I'm not sure what this means about updating existing database entries on prod.

### Notes

There was a line at the top of the search filter template like "if facet is not Controlling Body" that has been there since the [beginning of this file, 10 years ago.](https://github.com/Metro-Records/la-metro-councilmatic/blob/9f04ff1af43d992285496e686db8d212623beb65/lametro/templates/partials/search_filter.html)

So what this was guarding against, I do not know. I guess since `controlling_body` is defined in councilmatic-core Haystack indexing, we needed to override it and then skip it because we weren't using it / had hard-coded it to a placeholder value "Board of Directors"

### DIFF
--- a/lametro/search_indexes.py
+++ b/lametro/search_indexes.py
@@ -39,6 +39,9 @@ class LAMetroBillIndex(BillIndex, indexes.Indexable):
         orgs_list = [action["organization"].name for action in obj.actions_and_agendas]
         return set(orgs_list)
 
+    def prepare_controlling_body(self, obj):
+        return obj.controlling_body.name
+
     def prepare_actions(self, obj):
         return [str(action) for action in obj.actions.all()]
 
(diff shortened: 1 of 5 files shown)


yshng:
Force-pushed just to rebase against up-to-date main and split off the committee view/page PR.

Issue #1278, opened 2025-11-18
Item is showing as Withdrawn incorrectly


neilarellano:
Greetings Team!

We encountered an issue where the file ID 2025-0921 is showing as "Withdrawn" on both the Operations Committee and Executive Management Committee meetings.

<img width="1852" height="751" alt="Image" src="https://github.com/user-attachments/assets/9ba122f1-7127-470d-a6d0-f366747f2105" />

<img width="1831" height="705" alt="Image" src="https://github.com/user-attachments/assets/91a1d64a-0beb-4faf-880e-ef0af5208b67" />

This item was Withdrawn only from the Executive Management Committee meeting.

This file ID's status is set to "Agenda Ready" still. It was marked as "Withdrawn" only within the Minutes for each meeting (prior to the meeting).

We wanted to find out what the current logic is for marking items as Withdrawn. And also to see if we can adjust it so that it looks for "Withdrawn" action within the meeting minutes instead of using just the file ID status.

Please advise.

neilarellano:
Here is what we want:

FOR THE ITEM PAGE:
if the file ID is in two different meetings - do not follow the usual logic and do not show any tag in the board report page
only apply a non-"active" action tag if the file ID gets forwarded to a board meeting and the minutes for that board meeting are approved

FOR THE AGENDA PAGE:
have the file ID item tags based on the respective meeting instead of what is just on the file ID
What we want is for the action tag to be based off of the minutes action for the respective meeting. In this way if an item is withdrawn but stays in another meeting, it will show the proper status for each meeting separately.

PR #1308, opened 2026-06-08
Update bill tagging for search, bill detail, and event detail

yshng:
## Overview

Before, the inferred status of a bill depended on the last action overall. This was displayed in search, bill detail, and event detail views.

PR changes:

- Event detail shows the bill action associated with that event.
- Bill detail and search show the current status (latest action), except in particular circumstances where the bill is connected to multiple meetings. Then it follows the below logic from Issue #1278:

  1. Check if one of the meetings is Board of Directors, and if that meeting has approved minutes
  2. If so, use whatever the last action from the board is
  3. If not, only display a status if that status is "Active"

Connects #1278


lametro/models/legislative.py
```diff
@@ -166,24 +175,63 @@ def friendly_name(self):
...
-        if description and description.upper() in BILL_STATUS_DESCRIPTIONS.keys():
-            return BILL_STATUS_DESCRIPTIONS[description.upper()]["search_term"]
-        return None
+        # If on the agenda of more than one org, see if one is the Board of Directors
+        aa = self.actions_and_agendas
+        unique_orgs = {a["organization"] for a in aa if "organization" in a}
+
+        if len(unique_orgs) > 1:
+
+            try:
```
antidipyramid: I think we can assume that `Board of Directors` does exist in the db at this point.

lametro/templatetags/lametro_extras.py
```diff
@@ -355,10 +344,18 @@ def get_events_with_manual_broadcasts():
...
+def bill_status_from_event_action(description):
     if description and description.upper() in BILL_STATUS_DESCRIPTIONS.keys():
         return BILL_STATUS_DESCRIPTIONS[description.upper()]["search_term"]
-    return None
+    return ""
+
+
+@register.filter
+@stringfilter
+def inferred_status_label(status):
```
antidipyramid: What do you think about keeping the html out of the filters? In my mind, it keeps the code more readable both in this file and in the actual template.

lametro/models/legislative.py
yshng: I would never dare to say so myself. Much more comfortable with you making the call

lametro/templatetags/lametro_extras.py
yshng: That makes sense to me

yshng:
@antidipyramid Not sure what's going on with the deployment, but I have made the suggested changes!

PR #1311, opened 2026-06-11
Show translations files

xmedr:
## Overview

When needed, this pr asks the translation suite for any approved file formats or translations for agendas and board reports. It then displays those links on necessary pages.

For detail pages, we're listing out each link received and separates them based on file format. For the index page, we're using collapsable dropdowns to keep things neat, and the layout changes depending on whether there is one or multiple current/next meetings.

- Connects https://github.com/datamade/la-metro-translations/issues/37

### Demo
**An event detail page with alternate files:**
image

---

**The index with one current meeting and an open dropdown:**
image

---

**The index with multiple current meetings and closed dropdowns:**
image

### Notes

This pr is meant to be tested/brought in with a [matching pr in the translation suite](https://github.com/datamade/la-metro-translations/pull/54). Both review apps are set up to talk to each other. This pr's review app database is connected to staging's, so it should have plenty of entities to test with.

## Testing Instructions
* Head to an event's or bill's detail page
  * Example with approved translations in [boardagendas](https://la-metro-cou-feat-show--xlm9tn.herokuapp.com/event/ad-hoc-board-composition-committee-914582fb4f35/) and in the [translation suite](https://la-metro-tra-feat-doc-u-ktpdzz.herokuapp.com/document/edit/1217/)
  * If the boardagendas links don't show up, try again in just a bit. The review app seems to take a second to return results if it hasn't been interacted with in a bit.
* Confirm that links to alternate files are displayed
* Confirm that links lead to the right agenda/board report, and for the right language described in the link text
* Confirm that the layouts for the dropdowns for the index page shown in the demo above look alright (hard to test this on the review app atm since the next meeting is cancelled)


tests/test_cms.py
```diff
@@ -43,6 +44,10 @@ def test_event_admin_renders_based_on_context(
 
     dummy_service.get_agenda = get_agenda
 
+    mocker.patch(
+        "lametro.views.check_translations",
+        return_value={"pdf": [], "rtf": []},
+    )
```
xmedr, line 50: Prevents the test from trying to access the translation suite

tests/test_requests.py
```diff
@@ -1,16 +1,17 @@
...
 
 def test_timed_get(mocker):
-    ten_seconds_ago = datetime.now(tz=timezone.utc) - timedelta(seconds=20)
-    with pytest.raises(LAMetroRequestTimeoutException):
-        with freeze_time(ten_seconds_ago):
-            timed_get("https://google.com", timeout=5)
+    """
+    First run mocks time.time() so that it looks like 9 seconds have passed.
+    Second run uses real time.
+    """
```
xmedr, line 10: This was initially failing for me locally despite not making any changes to the timed_get method. Turns out that freeze_time may not have actually been freezing the time.time() method used there. So this just makes that check a little more foolproof.

lametro/views.py
```diff
@@ -109,6 +109,24 @@ def get_context_data(self, **kwargs):
         )
         context["form"] = LAMetroCouncilmaticSearchForm()
 
+        if not context["current_meeting"] and not context["upcoming_board_meetings"]:
+            return context
```
xmedr: In the index, eventdetail, and billdetail views, we're checking for translations only if entities have board reports/agendas to start with.

lametro/views.py
```diff
@@ -109,6 +109,24 @@ def get_context_data(self, **kwargs):
...
+
+        # Check for translated/converted files in the translation suite
+        meetings_to_check = (
+            context["current_meeting"] or context["upcoming_board_meetings"]
+        )
```
xmedr: This mimics the check being performed in the index's template. Start with current meeting(s), then if that doesn't exist, check for upcoming meeting(s).

lametro/templates/index/_meeting_details_next.html
```diff
@@ -1,57 +1,118 @@
...
-        <div>
-          {% if meeting.status == 'cancelled' %}
-            <h3 class="d-inline h4"><del>{{  meeting.link_html | safe }}</del></h3>
-            <small><span class="label label-stale text-light rounded">Cancelled</span></small>
-          {% else %}
-            <h3 class="h4">{{  meeting.link_html | safe }}</h3>
-          {% endif%}
-          <span class="small text-muted">{{ meeting.description }}</span>
-        </div>
+{% if upcoming_board_meetings|length > 1 %}
```
xmedr, line 3: Lots of adjustments to this template. Essentially, the view can return up to two upcoming meetings, but this template was only ever showing a single one. These changes allow it to show two if needed.

Otherwise, this acts the same as the `meeting_details_current` template, showing the translations dropdowns differently depending on whether we have one or two meetings listed.

hancush (changes requested):
Hey, @xmedr, I love the UI, but because of latency and translations being supplemental to core page content, it'd be better to retrieve them clientside so page loads aren't delayed. 

Because the request involves an API key, we should move your `check_translations` logic into its own POST route (which is CSRF protected and therefore can't be accessed without a real session token), then call it from the template after the page renders. 

Here's a rough outline of the change, c/o Claude:

```python
from django.http import JsonResponse
from django.utils.decorators import method_decorator
from django.views import View
from django.views.decorators.csrf import csrf_protect

@method_decorator(csrf_protect, name='dispatch')
class TranslationFilesView(View):
    def get(self, request, document_id):
        entity_type = request.GET.get("entity_type")
        if entity_type not in ("event", "bill"):
            return JsonResponse({"error": "invalid entity_type"}, status=400)
        
        # check_translations_logic
        if response is None:
            return JsonResponse({"pdf": [], "rtf": []})
        
        return JsonResponse(response)
```

Wired up in urls.py:

```python
path(
    "api/translations/<str:document_id>/",
    TranslationFilesView.as_view(),
    name="translation-files",
),
```

And the JS side (e.g. on DOMContentLoaded):

```javascript
fetch(`/api/translations/${documentId}/?entity_type=event`, {
    headers: { "X-CSRFToken": getCookie("csrftoken") },
})
  .then(r => r.json())
  .then(data => {
    // render translation links
  });
```

lametro/templates/common/_get_pdf_accordion.html
```diff
@@ -0,0 +1,24 @@
+{% load lametro_extras %}
```
hancush: This seems identical to get_rtf_accordion, besides the format string (PDF/RTF) and the list of items to iterate. How about one template that accepts variables for those things, which [you can pass with your `include` tag](https://docs.djangoproject.com/en/6.0/ref/templates/builtins/#:~:text=included%20template%3A).

lametro/templatetags/lametro_extras.py
```diff
@@ -377,3 +377,12 @@ def show_tooltip(label):
...
         return {"tooltip": tooltip.content} if not tooltip.disabled else None
+
+
+@register.filter
+def keyvalue(dict, key):
+    """
+    Convenience method to search for an item in a dict using a key stored in a variable.
+    Used as : dict_to_search|keyvalue:some_var_with_key
+    """
+    return dict[key]
```
hancush: Are you sure the key will exist? If not, use `.get()` so this doesn't crash on missing keys.

lametro/templates/common/_get_pdf_accordion.html
```diff
@@ -0,0 +1,24 @@
...
+
+<div class="accordion w-100" id="accordionAgendaPDFs-{{ forloop.counter }}">
```
hancush: I think this will evaluate to an empty string, unless you pass the forloop.counter value along with your include statement.

lametro/utils.py
```diff
@@ -145,3 +147,33 @@ def trace_function(frame, event, arg):
         sys.settrace(None)
 
     return resp
+
+
+def check_translations(document_id, entity_type):
+    """
+    Checks translation suite for alternate files to display on this app
+    """
+    api_url = f"http://{settings.TRANSLATION_SUITE_URL}/api/document-files/"
```
hancush: If you're sending an API key, you should use `https`.

lametro/templates/common/_get_pdf_accordion.html
xmedr: Gotcha! I'm now using the meeting_id as the unique identifier then, since that definitely gets passed through.

lametro/templates/common/_get_pdf_accordion.html
xmedr: Nice! Doing it this way turned out to be so much cleaner 🤘

xmedr (commented):
That makes sense! I've got all the feedback incorporated now

I've removed any of those requests from the index/detail views and created a specific view just for contacting the translation suite. Then, in `utils.js` there's some new methods for using our new view client side. I've also adjusted the templates to account for any changes this new pattern required

hancush (approved):

PR #1315, opened 2026-07-02
Group English download links together

yshng:
## Overview

When receiving translations, splice off the first RTF translation which is always English by the ordering in the translations app (at least it is for now, maybe I should add a check?) and put it in a special placeholder div in event detail and bill detail pages.

Connects https://github.com/datamade/la-metro-translations/issues/61

### Demo

Should look like this (note that these screenshots were taken from hardcoded versions of the template to show what it should look like when the English RTF div is populated by `renderLinks`)

[
Screenshot 2026-07-02 at 10 09 13 AM
Screenshot 2026-07-02 at 10 06 40 AM
](url)

## Testing Instructions

- Added jest to test DOM manipulation since I don't have a way to directly verify locally (can't get translations into local translations instance due to Mistral - though I suppose I could point to staging?) But it's good to have tests regardless.
- Run `npm install` then `npx jest`
- Find an meeting and board report with translations in the preview app,
  - Check that English RTF appears above with PDF, and not with the other RTF translations. 
  - Check that wording/formatting is correct

Note that `jest` testing isn't integrated into our CI workflow yet. Should it be?

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,14 @@ LA Metro Councilmatic has a basic test suite. If you need to run it, simply run:
 docker-compose -f docker-compose.yml -f tests/docker-compose.yml run --rm app
 ```
 
+There are also JavaScript unit tests, which live in `tests/js/`. Run them with:
+
+```bash
+docker-compose -f docker-compose.yml -f tests/docker-compose.yml run --rm node
+```
+
+or, on a host with Node 18+ installed, `npm ci && npm test`.
+
 ### Load testing
 
 LA Metro Councilmatic uses [Locust](https://docs.locust.io/en/stable/) for load
--- a/lametro/static/js/utils.js
+++ b/lametro/static/js/utils.js
@@ -75,6 +75,23 @@ class DetailPageTranslationUtils {
         // Display links to translations as consecutive anchor tags,
         // with a separator in between each
         if (linksArr.length > 0) {
+            // Special treatment for English RTF
+            // Languages are ordered so English will always be first
+            if (file_format === "rtf") {
+                const engFile = linksArr.shift()
+                const rtfDisplay = document.getElementById(`english-${document_type}-rtf-display`)
+                const linkEl = document.createElement("a")
+                linkEl.href = engFile.url
+                linkEl.target = "_blank"
+                const icon = document.createElement("i")
+                icon.classList.add("fa", "fa-file-text-o")
+                icon.setAttribute("aria-hidden", "true")
+                linkEl.appendChild(icon)
+                linkEl.append(" " + engFile.link_text + " [RTF]")
+                rtfDisplay.appendChild(linkEl)
+                rtfDisplay.classList.remove("d-none")
+            }
+
             const translationList = document.getElementById(`${document_type}-${file_format}s`)
             const translationDisplay = document.getElementById(`${document_type}-${file_format}-display`)
             const separator = document.createElement("span")
@@ -82,10 +99,11 @@ class DetailPageTranslationUtils {
             separator.innerHTML = "|"
 
             linksArr.map((file, index, array) => {
+
                 const linkEl = document.createElement("a")
                 linkEl.href = file.url
                 linkEl.target = "_blank"
-                linkEl.innerHTML = file.link_text
+                linkEl.append(file.link_text)
                 translationList.appendChild(linkEl)
                 if (index < array.length - 1) {
                     translationList.appendChild(separator.cloneNode(true))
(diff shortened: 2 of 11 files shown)


lametro/static/js/utils.js
```diff
@@ -75,17 +75,36 @@ class DetailPageTranslationUtils {
         // Display links to translations as consecutive anchor tags,
         // with a separator in between each
         if (linksArr.length > 0) {
+            // Special treatment for English RTF
+            // Languages are ordered so English will always be first
+            if (file_format === "rtf") {
+                const [engFile] = linksArr.splice(0,1)
```
xmedr: It looks like this line exists to remove the first element of the `linksArr` and put it into `engFile`, is that right?

If so, would [Array.shift()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/shift) serve the same purpose and make this a bit easier to read? Maybe something like:
```
const engFile = linksArr.shift()
```

lametro/static/js/utils.js
```diff
@@ -75,17 +75,36 @@ class DetailPageTranslationUtils {
...
+                const rtfDisplay = document.getElementById(`english-${document_type}-rtf-display`)
+                const linkEl = document.createElement("a")
+                linkEl.href = engFile.url
+                linkEl.target = "_blank"
+                const icon = document.createElement("i")
+                icon.classList.add("fa")
+                icon.classList.add("fa-file-text-o")
```
xmedr: We can consolidate these into one `.add()`!

```suggestion
                icon.classList.add("fa", "fa-file-text-o")
```

lametro/templates/legislation.html
xmedr: Think this might also need to have PDF at the end similar to the event template

lametro/templates/event/event.html
```diff
@@ -92,10 +92,12 @@ <h2 class="h4">Agenda</h2>
                                 <div>
                                     <strong>
                                         <a href='{{agenda.url}}' target="_blank">
-                                            <i class="fa fa-file-text-o" aria-hidden="true"></i> Download Agenda (English)
+                                            <i class="fa fa-file-text-o" aria-hidden="true"></i> Download Agenda (English) [pdf]
```
xmedr: I know Neil had this as lowercase in his screenshot, but I'm realizing we may want to standardize whether we're using `PDF/RTF` or `pdf/rtf` on event and bill detail pages. Let's maybe go with making them uppercase, since text elsewhere on the page has PDF written that way. So something like:

Image

lametro/static/js/utils.test.js
```diff
@@ -0,0 +1,144 @@
+const { IndexTranslationUtils, DetailPageTranslationUtils } = require("./utils")
```
xmedr, line 1: I'm currently getting an error after installing and running jest. It's a bit scant on details, but do you have an idea what's going on here?

```bash
>>> npx jest
Unexpected token .
```

xmedr (changes requested):
Solid stuff! I've got a few suggestions/questons below, but the app works as expected and looks great.

Thanks for adding js tests to this, especially for agendas on the index since that's harder to set up! I think adding jest to CI is a great idea! Would you be able to open an issue for that in this repo?

Also, the 61 you have linked in your Overview is a councilmatic specific issue haha. Try using just the straight up link to an issue when it's in a different repo. Here's the link for translation's issue 61: https://github.com/datamade/la-metro-translations/issues/61

lametro/static/js/utils.js
yshng: yes, I didn't know about Array.shift(), thanks!

lametro/static/js/utils.test.js
yshng: This could be a node version issue? In any case, I went ahead and just integrated jest into the Docker set up and CI workflow.

yshng: Should've mentioned test instructions are included in README update

yshng:
@xmedr note that node/jest are included in the test CI workflow that ran with the last commit on this branch

README.md
```diff
@@ -201,6 +201,14 @@ LA Metro Councilmatic has a basic test suite. If you need to run it, simply run:
 docker-compose -f docker-compose.yml -f tests/docker-compose.yml run --rm app
 ```
 
+There are also JavaScript unit tests, which live in `tests/js/`. Run them with:
+
+```bash
+docker-compose -f docker-compose.yml -f tests/docker-compose.yml run --rm node
```
xmedr, line 207: Nice, this is clutch

xmedr (approved):
This is perfect, thanks for your work on this! Bring this in and deploy it after tests pass on main

Issue #1320, opened 2026-07-09
Document SmartLogic


antidipyramid:
Connects #1314 

Our documentation doesn't explain:

1. what SmartLogic is/does within the app
2. how authorization to SmartLogic works, including key rotation
3. differences between how production & staging configure/interact with SmartLogic

This should all be clear and accessible within the Quarto docs.

PR #1324, opened 2026-07-23
Add documentation for Progress Semaphore SES fka SmartLogic

yshng:
## Overview

We didn't have easily accessible documentation for our integration with Progress SES, which we usually call "SmartLogic". I integrated some information from various sources and wrote it all down here.

Connects #1320

## Testing Instructions

You can read the file directly, or pull the branch down locally and 
- `cd docs`
- `quarto preview`

Confirm the information is correct.

### DIFF
--- a/docs/semanticsearch.qmd
+++ b/docs/semanticsearch.qmd
@@ -12,7 +12,9 @@ order: 3
 
 ## Overview
 
-Metro uses what was SmartLogic Semaphore, and is now [Progress Semaphore Semantic Enhancement Service (SES) ](https://docs.progress.com/bundle/semaphore-5-semantic-enhancement-server-administration/page/topics/ses-admin/welcome.html), to develop semantic metadata for Metro data, creating topic-based classifications for bills. This is what enables live auto-suggestions in the Board Agenda's search, as well as the ability to select from those suggestions to perform a topic search instead of a regular text search.
+Metro uses what was SmartLogic Semaphore, and is now [Progress Semaphore Semantic Enhancement Service (SES) ](https://docs.progress.com/bundle/semaphore-5-semantic-enhancement-server-administration/page/topics/ses-admin/welcome.html), to maintain an ontology of concepts and controlled vocabulary of temrs around Metro data. The ontology describes the entities within the Metro universe and their special relations to one another, for example, a "committee" has "members" that "sponsor" "bills". Controlled vocabularies function like thesauruses such that every concept has a constellation of potential terms that are established as referring to "the same thing", for example, a "bill" is a "matter" is a "board report".
+
+This system of knowledge management enables services like tagging, topic search, and live auto-suggestions, with Metro's end goal being to help users find what they are looking for more easily.
 
 Related links:
 
@@ -23,11 +25,11 @@ Related links:
 
 Two major syncs happen nightly:
 
-(1) Metro maintains a taxonomy of terms, concepts, etc with SES. SES gets new matter/bill data from Legistar, indexes the matters/bills with respect to Metro's taxonomy, and returns those indexes back to Legistar[^1], each with its own Semaphore GUID.
+(1) As Metro makes any updates to its knowledge management system, SES indexes it all and sends a mapping of indexes/topics/concepts/terms to Semaphore GUIDs to Legistar[^1].
 
 [^1]: For all indexes, see the [/MatterIndexes endpoint](https://webapi.legistar.com/Help/Api/GET-v1-Client-MatterIndexes)). And for the indexes for any particular matter, see the [the Matters/{MatterId}/Indexes endpoint](http://webapi.legistar.com/Help/Api/GET-v1-Client-Matters-MatterId-Indexes).
 
-(2) Independently, LA Metro Councilmatic does its scrape, which includes dicarding the indexes related to any changed bill and re-populating them, so stale indexes don't persist.
+(2) Independently, LA Metro Councilmatic does its scrape of Legsitar, where Metro has added associated indexes with bills. The scrape discards the indexes related to any changed bill and re-populates them, so stale indexes don't persist.
 
 Because the bill scraper only picks up the index labels and not their GUIDs, when SES is done syncing with Legistar, it pings an endpoint on LA Metro Councilmatic to run the `refresh_guid` management command to update a lookup table mapping labels to GUIDs[^2].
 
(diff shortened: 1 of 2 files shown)


docs/semanticsearch.qmd
```diff
@@ -0,0 +1,48 @@
+---
+date: 07/23/2026
+date-modified: last-modified
+title: "Semantic Search"
+keywords:
+  - Matter Index
+  - Topic Search
+  - SmartLogic
+  - SES
```
hancush, line 9: thats-hot.gif

docs/semanticsearch.qmd
```diff
@@ -0,0 +1,48 @@
...
+Related links:
+
+- [SmartLogic/Semaphore training notes (internal doc)](https://docs.google.com/document/d/173eIduyf5-obXr0Jzlo1N5eLs7Zbt7jJWvXicKPllwA/edit?tab=t.0)
+- [Progress Data Cloud migration](https://github.com/Metro-Records/la-metro-councilmatic/pull/1258)
+
+## Syncs
+
+Two major syncs happen nightly:
+
+(1) SES gets data from Legistar, creates classifications for the data, and feeds it back into the Legistar API as indexes/topics[^1], each with its own Semaphore GUID.
```
hancush: Small clarification: Metro maintains a taxonomy of terms, topics, etc., in SES. This taxonomy gets synced to Legistar. Taxonomy elements are associated with board reports ("matters") through topics ("indexes"). So, SES tells us when it has published updated taxonomy information to Legistar, which we then update in the app. The tags come through with board report scrapes. And, since the app has the topics and their taxonomy identifiers, it can look up suggestions (given a partial query) and related concepts (given a complete search term) from SES.

hancush (changes requested):
One (ish?) point of clarification. LMK if you want a suggested edit, otherwise happy for you to integrate.

docs/semanticsearch.qmd
yshng: To make sure I understand: the work SES is doing is taking the pre-established Metro taxonomy and doing the associations?

docs/semanticsearch.qmd
```diff
@@ -0,0 +1,48 @@
...
+[^2]:[`refresh_guid_trigger`](https://github.com/search?q=repo%3AMetro-Records%2Fla-metro-councilmatic%20refresh_guid_trigger&type=code)
+
+Further, because the SES-Legistar sync happens independently of the scrape, the `refresh_guid` command takes into account the possibility of lag, and does not get rid of any GUIDs that still exist in the Councilmatic database, regardless of their status in Legistar.
+
+Read more: [Metro/SmartLogic integration spec (internal doc)](https://docs.google.com/document/d/1m_nZv33YTvdiG2rLSxTXpMvjb00agOe2ZRo79gcpmHE/edit?tab=t.0)
+
+
+## Authentication and key rotation
+
+The SES API key expires periodically, so we have [a command](commands.qmd#update-ses-api-keys-on-heroku) to update it in all our Heroku environments. It runs on prod through the Heroku Scheduler add-on.
```
xmedr: This bit looks good to me! Here's some more details in case we wanted to be specific about time and usage:

- The SES key expires every _90 days_, and the scheduled command will update it if it sees that the key is _2 weeks away_ from expiring
- In order to be explicit about what we want to do with the key, we have a few flags that can be used with the management command
  - `check_ses_token` by itself only checks the expiration date
  - `check_ses_token --update_token` will check and update the key if we're two weeks away from expiration.
  - `check_ses_token --update_token --force_var_update` will check and update the key regardless of when it expires

xmedr (approved):

docs/semanticsearch.qmd
yshng: Thanks! Some of this information I think is already on the commands.qmd doc that I link to, but I can include the more specific timing information as well.

docs/semanticsearch.qmd
```diff
@@ -0,0 +1,48 @@
...
+  - Matter Index
+  - Topic Search
+  - SmartLogic
+  - SES
+order: 3
+---
+
+## Overview
+
+Metro uses what was SmartLogic Semaphore, and is now [Progress Semaphore Semantic Enhancement Service (SES) ](https://docs.progress.com/bundle/semaphore-5-semantic-enhancement-server-administration/page/topics/ses-admin/welcome.html), to develop semantic metadata for Metro data, creating topic-based classifications for bills. This is what enables live auto-suggestions in the Board Agenda's search, as well as the ability to select from those suggestions to perform a topic search instead of a regular text search.
```
hancush: We also use it to populate the related topics on the search results page!

Image

docs/semanticsearch.qmd
```diff
@@ -0,0 +1,48 @@
...
+Related links:
+
+- [SmartLogic/Semaphore training notes (internal doc)](https://docs.google.com/document/d/173eIduyf5-obXr0Jzlo1N5eLs7Zbt7jJWvXicKPllwA/edit?tab=t.0)
+- [Progress Data Cloud migration](https://github.com/Metro-Records/la-metro-councilmatic/pull/1258)
+
+## Syncs
+
+Two major syncs happen nightly:
+
+(1) Metro maintains a taxonomy of terms, concepts, etc with SES. SES gets new matter/bill data from Legistar, indexes the matters/bills with respect to Metro's taxonomy, and returns those indexes back to Legistar[^1], each with its own Semaphore GUID.
```
hancush: I think Metro just maintains the taxonomy in SES, then SES pushes the index mapping to Legistar, where Metro can assign indices to matters.

yshng: Okay, so trying to put into my own words what your comment says: 

1. Metro maintains an ontology which is like, in our world, a "bill" is something "sponsored" by a "board member" who has "membership" in a "committee".
2. SES takes in data from Legistar, returns a mapping of indices to bills 
3. ("Metro can assign indices to matters") Metro Councilmatic then updates all the bills with the new indices

This would imply that SES doesn't cause MatterIndexes to be updated on Matters in Legistar? That only happens on the frontend? Otherwise I'm not sure what "Metro can assign indices to matters" means.

Appreciate the continued clarification!

yshng: Okay during sprint planning we clarified that SES is not ingesting any data from Legistar at all, Metro is mapping indices to matters themselves, and SES is merely a tool for storing and maintaining the Metro ontology that can be queried to provide access to nodes and relations from that ontology.

hancush: yes yes yes

thank you for this concise summary of my rambling

hancush (approved):

